PR #2, opened 2019-02-12
downcase digital format src for more robust comparisons

dewey4iv:
ATLA currently displays Facet: `Mp3`. This was not in the list to be matched for `MP3` (`mp3` and `MP3` were present).

This PR changes the logic to downcase the incoming source for comparison to avoid similar bugs reminiscent of `bAd tYpING fRom tHE '90s`. 

Linked to issue: https://gitlab.com/notch8/atla_digital_library/issues/105


orangewolf:
@dewey4iv does it make sense to strip the string as well? that way spaces can't muck us up

dewey4iv:
Certainly does. Good call.

dewey4iv:
It certainly does @orangewolf. Good call.

Issue #10, opened 2019-07-01
Bulkrax Entry Interface


orangewolf:
As a depositor using the importer
I'd like to be able to see and update individual entries in an import
To help me unblock broken or incorrect imported works.

Create crud interface for Bulkrax::Entry objects (include delete / edit, maybe not create?).  There is a (bad) start to this in the importer show view, but it should be at minimum:

- Paginated
- Searchable (SQL LIKES, not solr probably?)
- Sortable
- Show (and sort/search) status messages
- Allow retry (for failed entries)
- Allow delete created work and retry (for successful entries)
- Allow update work (for successful entries)

Issue #9, opened 2019-07-01
Refactor Collection Creation


orangewolf:
Right now the OAI Collection creation doesn't use an entry format and should be changed to work more like ATLA's CDRI importer.

Also there are issues with creating Collections _while_ parsing the entries.  Because the entries get created in parallel multiple copies of the collection can be created.  To solve this could we potentially create collections while parsing the original file (as that is a single job)? And that would ensure all collections needed exist before ingest. This would also allow CSV import to create them from a defined row.

PR #13, opened 2019-07-04
CSV Importer 

geekscruff:
First pass at a CSV Importer, currently:

- supports metadata
- supports multiple file upload from a local folder, depends on a 'file' column in the csv with multiple values separated with | or ; (also small refactoring of the object factory to support multiple files)
- csv file must exist on server already; files must be in a folder called files within the same dir as the csv (working on principle we'll have an import path, eg. tmp/imports/import_no/
- supports rights statement override
- sets visibility to open (unless it's already set in a column called visibility)
- requires a column called 'identifier' in the CSV
- skips lines without an identifier
- validates required fields (static list at present)

**@todo** 

- some conflict with JS to sort out (buttons for OAI showing up in certain cases) - #16 
- collection creation - waiting on #9
- file upload #17
- field mappings (to de-couple from local metadata schema) - #10


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+        importer.parser_fields['csv_path'],
+        headers: true,
+        header_converters: :symbol,
+        encoding: 'utf-8'
+      )
+      csv_data = csv.read
+      raise StandardError, 'Identifier column is required' if csv_data.headers.include?(:identifier) == false
+
+      # skip rows without an identifier; remove any nil values with compact
+      csv_data.map { |row| row.to_h.compact! unless row[:identifier].blank? }.reject(&:blank?)
```
orangewolf: I think this code forces all the CSV rows in to memory at the same time. That could be bad if large CSVs travel through the system.  What do you think about refactorying this method and the create_works method so that a single foreach is used? It can still have the headers and skip the bad rows (you'd skip them with `next` instead of rejecting them). For the raise you could do an `if i == 0` on your iterator (which _may_ have to be a manual counter, I'm not sure).  

The last piece that's problematic is getting the count ahead of time. I'll comment about that down there

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -0,0 +1,66 @@
...
+      arr << 'files'
+      arr.join('/')
+    end
+
+    def entry_class
+      CsvEntry
+    end
+
+    def total
+      @total ||= records.count
```
orangewolf: If the performance discussion above is adopted, then I'd suggest doing a kind of naive implementation here. Let's let the out unix forbearers sort this one out for us ;-) https://stackoverflow.com/questions/2650517/count-the-number-of-lines-in-a-file-without-reading-entire-file-into-memory

If we just subtract 1 from this number (for the header) we should be in a reasonable spot and it will be a lot more performant

app/parsers/bulkrax/csv_parser.rb
geekscruff: If you see the latest commit I've switched to having #records return the CSV.foreach enumerator and #create_works loop through that with an index. Reading around CSV.foreach is the recommended way of doing this performance-wise.

app/parsers/bulkrax/csv_parser.rb
geekscruff: Done :)

app/parsers/bulkrax/csv_parser.rb
orangewolf: 👍 looks great

Issue #7, opened 2019-07-01
Work on pending specs / code coverage


orangewolf:
Goal here is to get to reasonable coverage on most basic objects

geekscruff:
Picked up in #44

PR #14, opened 2019-07-05
Increase test coverage

geekscruff:
See #7


orangewolf:
should we unWIP this and merge it as a partially completed tasks (leaving the issue open) so it doesn't get too far out of the main line?

PR #15, opened 2019-07-09
Refactor collections

### LINKED ISSUE #9: Refactor Collection Creation

geekscruff:
Fixes #9

* Moves Collection Creation to a factory / job. 
* Ensures collections get created before works.

### DIFF
--- a/app/jobs/bulkrax/import_work_job.rb
+++ b/app/jobs/bulkrax/import_work_job.rb
@@ -5,7 +5,7 @@ module Bulkrax
     def perform(*args)
       entry = Entry.find(args[0])
       begin
-        entry.build
+        reschedule(entry.id, ImporterRun.find(args[1]).id) if entry.build.blank?
         entry.save
       rescue StandardError => e
         ImporterRun.find(args[1]).increment!(:failed_records)
@@ -18,6 +18,10 @@ module Bulkrax
           ImporterRun.find(args[1]).increment!(:processed_records)
         end
       end
+    end
+
+    def reschedule(entry_id, run_id)
+      ImportWorkJob.set(wait: 1.minutes).perform_later(entry_id, run_id)
     end
   end
 end
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -37,8 +37,10 @@ module Bulkrax
       end
       add_visibility
       add_rights_statement
+      add_collections
+      self.parsed_metadata[Bulkrax.system_identifier_field] ||= [record['source_identifier']]
 
-      parsed_metadata
+      self.parsed_metadata
     end
 
     def record
@@ -49,12 +51,26 @@ module Bulkrax
       Bulkrax::CsvMatcher
     end
 
+    def collections_created?
+      return true if record['collection'].blank?
+      record['collection'].split(/\s*[:;|]\s*/).length == self.collection_ids.length
+    end
+
+    def find_or_create_collection_ids
+      return self.collection_ids if collections_created?
+      record['collection'].split(/\s*[:;|]\s*/).each do | collection |
+        c = Collection.where(Bulkrax.system_identifier_field => collection).first
+        self.collection_ids << c.id unless c.blank? || self.collection_ids.include?(c.id)
+      end unless record['collection'].blank?
+      return self.collection_ids
+    end
+
     def required_elements?(keys)
       !required_elements.map { |el| keys.include?(el) }.include?(false)
     end
 
     def required_elements
-      %w[title identifier]
+      %w[title source_identifier]
     end
   end
 end
--- a/app/models/bulkrax/entry.rb
+++ b/app/models/bulkrax/entry.rb
@@ -2,6 +2,8 @@ module Bulkrax
   class Entry < ApplicationRecord
     belongs_to :importer
     serialize :parsed_metadata, JSON
+    serialize :raw_metadata, JSON
+    serialize :collection_ids, Array
 
     attr_accessor :all_attrs, :last_exception
 
@@ -13,39 +15,56 @@ module Bulkrax
              :user,
              to: :parser
 
+    # return true or false here
     def build
       # attributes, files_dir = nil, files = [], user = nil
+      build_metadata
+      return false unless collections_created?
       begin
-        @item = Bulkrax::ApplicationFactory.for(factory_class.to_s).new(build_metadata, parser.files_path, [], user).run
-      rescue => e
+        @item = Bulkrax::ApplicationFactory.for(factory_class.to_s).new(self.parsed_metadata, parser.files_path, [], user).run
+      rescue StandardError => e
         status_info(e)
       else
         status_info
-        self.collection_id = @item.id if @item.is_a?(Collection)
       end
       return @item
     end
 
-    def collection
-      @collection ||= Collection.find(self.collection_id) if self.collection_id
+    def find_or_create_collection_ids
+      self.collection_ids
+    end
+
+    # override this to ensure any collections have been created before building the work
+    def collections_created?
+      true
     end
 
     def build_metadata
       raise 'Not Implemented'
     end
 
+    def rights_statement
+      parser.parser_fields['rights_statement']
+    end
+
+    # try and deal with a couple possible states for this input field
+    def override_rights_statement
+      %w[true 1].include?(parser.parser_fields['override_rights_statement'].to_s)
+    end
+
+    def add_rights_statement
+      self.parsed_metadata['rights_statement'] = [parser.parser_fields['rights_statement']] if override_rights_statement || self.parsed_metadata['rights_statement'].blank?
+    end
+
     def add_visibility
       self.parsed_metadata['visibility'] = 'open' if self.parsed_metadata['visibility'].blank?
     end
 
-    def add_rights_statement
-      if override_rights_statement || self.parsed_metadata['rights_statement'].blank?
-        self.parsed_metadata['rights_statement'] = [parser.parser_fields['rights_statement']]
+    def add_collections
+      if find_or_create_collection_ids.present?
+        self.parsed_metadata['collections'] ||= []
+        self.parsed_metadata['collections'] += find_or_create_collection_ids.map { |c| { id: c } }
       end
-    end
-
-    def override_rights_statement
-      ['true', '1'].include?(parser.parser_fields['override_rights_statement'].to_s)
     end
 
     def factory_class
--- a/app/models/bulkrax/oai_entry.rb
+++ b/app/models/bulkrax/oai_entry.rb
@@ -4,22 +4,16 @@ require 'ostruct'
 
 module Bulkrax
   class OaiEntry < Entry
-
     def raw_record
-      @raw_record ||= client.get_record({identifier: identifier, metadata_prefix: parser.parser_fields['metadata_prefix'] })
+      @raw_record ||= client.get_record(identifier: identifier, metadata_prefix: parser.parser_fields['metadata_prefix'])
     end
 
     def record
       raw_record.record
     end
 
-    def rights_statement
-      parser.parser_fields['rights_statement']
-    end
-
-    # try and deal with a couple possible states for this input field
-    def override_rights_statement
-      ['true', '1'].include?(parser.parser_fields['override_rights_statement'].to_s)
+    def sets
+      raw_record.record.header.set_spec
     end
 
     def contributing_institution
@@ -44,17 +38,41 @@ module Bulkrax
       end
       add_metadata('thumbnail_url', thumbnail_url)
 
-      add_visibility
-      add_rights_statement
-      self.parsed_metadata['source'] ||= [record.header.identifier]
+      self.parsed_metadata[Bulkrax.system_identifier_field] ||= [record.header.identifier]
       self.parsed_metadata['contributing_institution'] = [contributing_institution]
 
-      if collection.present?
-        self.parsed_metadata['collections'] ||= []
-        self.parsed_metadata['collections'] << {id: self.collection.id}
-      end
+      add_visibility
+      add_rights_statement
+      add_collections
 
       return self.parsed_metadata
+    end
+
+    def collections_created?
+      if sets.blank? || parser.collection_name != 'all'
+        self.collection_ids.length == 1
+      else
+        sets.length == self.collection_ids.length  
+      end
+    end
+
+    # Retrieve list of collections for the entry; add to collection_ids
+    # If OAI-PMH doesn't return setSpec in the headers for GetRecord, use parser.collection_name
+    #   in this case, if 'All' is selected, records will not be added to a collection.
+    def find_or_create_collection_ids
+      return self.collection_ids if collections_created?
+
+      if sets.blank?
+        c = Collection.where(Bulkrax.system_identifier_field => parser.collection_name).first
+        self.collection_ids << c.id unless c.blank? || self.collection_ids.include?(c.id)
+      else
+        sets.each do |set|
+          c = Collection.where(Bulkrax.system_identifier_field => set.content).first
+          self.collection_ids << c.id unless c.blank? || self.collection_ids.include?(c.id)
+        end
+      end
+      
+      return self.collection_ids
     end
 
   end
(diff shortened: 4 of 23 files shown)


orangewolf:
looks like this is coming along - one thing that presents problems currently is that collections all need to be made before any works are created, otherwise the works can not attach to them _and_ if you do the collection creation and work creation time you end up with duplicates (as works of the same collection are mostly bundled together).  

I think if we have this in a background job, there may need to be some sort of lock that prevents works from creating if there are any pending collection creations.

app/jobs/bulkrax/importer_collection_job.rb
```diff
@@ -0,0 +1,13 @@
+module Bulkrax
+  class ImporterCollectionJob < ApplicationJob
+    queue_as :import
+
+    def perform(*args)
+      entry = Entry.find(args[0])
+      importer = Importer.find(args[1])
+      collection = entry.build
```
geekscruff: @orangewolf this was my attempt at ensuring the collection gets created first, by passing back the collection object returned from the factory ... I'm not sure that would be entirely reliable. The other thing is that I rather assumed that only one collection would be being created, but I might be mistaken there.

orangewolf: I think there is likely still an issue with this creating multiples of a collection while it is running, but it does solve a lot of the other collection related problems and gives us a good place to _put_ that logic if we find we do need it. Unless there is another reason this code is WIP, let's get it merged and try it out and put some locking code in if we still need to.

geekscruff: @orangewolf  I've added some locking code - see comments above; I think it's working as it should so I think we can risk a merge and fuller test.

geekscruff: @orangewolf I have added a couple of database columns, and I've changed collection_id on entries to collection_ids (an Array), to support multiple collections.

app/models/bulkrax/oai_dc_entry.rb
```diff
@@ -6,8 +6,8 @@ class OaiDcEntry < OaiEntry
     matcher 'creator', split: true
     matcher 'date', from: ['date'], split: true
     matcher 'description'
-    matcher 'format_digital', from: ['format_digital', 'format'], parsed: true
-    matcher 'identifier', from: ['identifier'], if: ->(parser, content) { content.match(/http(s{0,1}):\/\//) }
+    matcher 'format_digital', from: %w[format_digital format], parsed: true
```
orangewolf: =-p 
%w is so trendy 
/end curmudgeon mode

app/models/bulkrax/oai_entry.rb
```diff
@@ -44,17 +38,30 @@ def build_metadata
...
-        self.parsed_metadata['collections'] ||= []
-        self.parsed_metadata['collections'] << {id: self.collection.id}
-      end
+      parsed_metadata
+    end
 
-      return self.parsed_metadata
+    # Retrieve list of collections for the entry; add to collection_ids
+    # If OAI-PMH doesn't return setSpec in the headers for GetRecord, use parser.collection_name
+    #   in this case, if 'All' is selected, records will not be added to a collection.
```
orangewolf: does that make the most sense or would it make more sense if all is selected for it to act the same way sets.blank? acts (ie add to default collection?)

app/models/bulkrax/oai_entry.rb
```diff
@@ -44,17 +38,30 @@ def build_metadata
...
+      if sets.blank?
+        c = Collection.where(Bulkrax.system_identifier_field => parser.collection_name).first
+        collection_ids << c.id unless c.blank? || collection_ids.include?(c.id)
+      else
+        sets.each do |set|
+          c = Collection.where(Bulkrax.system_identifier_field => set.content).first
+          collection_ids << c.id unless c.blank? || collection_ids.include?(c.id)
+        end
+      end
+      @collections ||= collection_ids
```
orangewolf: if you are going to ||= this, I think you should escape out of it at the top of the method. You're doing a lot of work you throw away if @collections is already set. 

Also this has the same "name is collections but is now returning collection_ids" problem mentioned above.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -10,19 +10,38 @@ def self.parser_fields
       }
     end
 
-    def run
-      create_works
-    end
-
     def records(_opts = {})
-      CSV.foreach(
+      @records ||= CSV.foreach(
```
orangewolf: does this work or does the @records become a foreach that would need to be rewound to actually be useful?

app/parsers/bulkrax/oai_dc_parser.rb
```diff
@@ -1,6 +1,8 @@
 module Bulkrax
   class OaiDcParser < ApplicationParser
-    attr_accessor :client, :headers
+    include Hyrax::Lockable
```
orangewolf: you include a lock module here, but not in csv_parser. Is there a missing include? Should this include go up to application_parser?

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -10,19 +10,38 @@ def self.parser_fields
...
         headers: true,
         header_converters: :symbol,
         encoding: 'utf-8'
       )
     end
 
+    def create_collections
+      return unless records.headers.include?(:collection)
+
+      acquire_lock_for(importer.id) do
```
orangewolf: ok, this is really cool.  It does mean Bulkrax now requires a running redis instance (which is true of all Hyrax even if you don't use sidekiq apparently, I did _not_ know that.. ¯\_(ツ)_/¯ I think that it is probably ok, but should be mentioned in the Bulkrax README as an explicit need)

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -10,19 +10,38 @@ def self.parser_fields
...
+          collection.split(/\s*[:;|]\s*/).each do |collection|
+            metadata = {
+              title: [c],
+              Bulkrax.system_identifier_field => [collection],
+              visibility: 'open',
+              collection_type_gid: Hyrax::CollectionType.find_or_create_default_collection_type.gid
+            }
+
+            new_entry = collection_entry_class.where(importer: importer, identifier: collection, raw_metadata: metadata).first_or_create!
+            ImportWorkCollectionJob.perform_later(new_entry.id, importer.current_importer_run.id)
```
orangewolf: I think for this lock to be affective these still need to be a perform_now?

orangewolf: otherwise are they spinning off in the background while you blaze through creating works?  Specifically what happens in the 10, 50 or 100 worker cases?

app/parsers/bulkrax/oai_dc_parser.rb
```diff
@@ -76,42 +74,38 @@ def list_sets
...
-          collection = Collection.where(identifier: [set.spec]).first
-          collection ||= Collection.create!(attrs)
+        next unless collection_name == 'all' || collection_name == set.spec
+
+        metadata[:title] = [set.name]
+        metadata[Bulkrax.system_identifier_field] = [set.spec]
+
+        new_entry = collection_entry_class.where(importer: importer, identifier: set.spec, raw_metadata: metadata).first_or_create!
+        acquire_lock_for(importer.current_importer_run.id) do
+          ImportWorkCollectionJob.perform_later(new_entry.id, importer.current_importer_run.id)
```
orangewolf: same issue in that you don't know when this gets finished queue wise as mentioned above

db/migrate/20190715161939_add_collections_to_importer_runs.rb
```diff
@@ -0,0 +1,6 @@
+class AddCollectionsToImporterRuns < ActiveRecord::Migration[5.1]
+  def change
+    add_column :bulkrax_importer_runs, :processed_collections, :integer, default: 0
```
orangewolf: I dig this

db/migrate/20190715162044_change_collection_ids_on_entries.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeCollectionIdsOnEntries < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :bulkrax_entries, :collection_id, :collection_ids
```
orangewolf: does the type need to change to text here?

app/models/bulkrax/entry.rb
```diff
@@ -17,45 +19,54 @@ def build
...
+        parsed_metadata['collections'] += collections.map { |c| { id: c } }
+      end
     end
 
     def factory_class
       Work
     end
 
     def status
-      if self.last_error_at.present?
```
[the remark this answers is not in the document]
geekscruff: That's rubocop changing things, although in my own code I tend not to use self / return. I'll have a look at bixby.

app/models/bulkrax/oai_dc_entry.rb
geekscruff: Sorry! (rubocop again, I can never remember the trendy syntax)

app/models/bulkrax/oai_entry.rb
geekscruff: So I think in my refactor, we don't have a default collection. This is one I wanted to chat to you about.

app/parsers/bulkrax/csv_parser.rb
geekscruff: Yes, this is where I was struggling to test effectively, because my solr kept crashing. I'll spend some more time on this.

geekscruff:
@orangewolf I have gotten into a habit of running rubocop against my code and I realise now that it makes reviewing PRs harder, and creating inconsistency across the codebase. I think we just need to decide on our code standards and at this stage it doesn't have to be rubocop/bixby.

geekscruff:
@orangewolf latest commit addresses a lot of the above and includes an alternative to the lock. I'm not convinced the existing lock mechanism is the right thing. What I've put in instead is a check that the collections exist, and a job reschedule if they don't.

This PR has turned into a bit of a monster, for what seems like a simple change and I'm worried it's turning into a time sync.

orangewolf:
> @orangewolf I have gotten into a habit of running rubocop against my code and I realise now that it makes reviewing PRs harder, and creating inconsistency across the codebase. I think we just need to decide on our code standards and at this stage it doesn't have to be rubocop/bixby.

I definitly think Rubocop is the right thing. It should really be part of our base pattern and often isn't. @smcfarlane made a big push for it for a while, but it seems like it kind of fell off again. Maybe we can do a breakfast and learn on it for the N8 team?

app/jobs/bulkrax/import_work_job.rb
```diff
@@ -4,20 +4,25 @@ class ImportWorkJob < ApplicationJob
 
     def perform(*args)
       entry = Entry.find(args[0])
+      importer_run = ImporterRun.find(args[1])
```
orangewolf: I think this causes problems - ImportRun has to be loaded just in time or multiple jobs will overwrite each other.

app/jobs/bulkrax/import_work_job.rb
```diff
@@ -4,20 +4,25 @@ class ImportWorkJob < ApplicationJob
...
           raise entry.last_exception
         else
-          ImporterRun.find(args[1]).increment!(:processed_records)
+          importer_run.increment!(:processed_records)
         end
       end
     end
+
+    def reschedule(entry_id, run_id)
+      ImportWorkJob.perform_later(entry_id, run_id)
```
orangewolf: probably should put a delay to not flood the workers with stuff they can't actually do anything with

app/models/bulkrax/entry.rb
```diff
@@ -13,39 +15,56 @@ class Entry < ApplicationRecord
              :user,
              to: :parser
 
+    # return true or false here
     def build
       # attributes, files_dir = nil, files = [], user = nil
+      build_metadata
+      return false if collections_created? == false
```
orangewolf: maybe `return false unless collections_created?`

app/models/bulkrax/entry.rb
```diff
@@ -13,39 +15,56 @@ class Entry < ApplicationRecord
...
-      rescue => e
+        Bulkrax::ApplicationFactory.for(factory_class.to_s).new(self.parsed_metadata, parser.files_path, [], user).run
+      rescue StandardError => e
         status_info(e)
       else
         status_info
-        self.collection_id = @item.id if @item.is_a?(Collection)
       end
-      return @item
+      true
```
orangewolf: I miss returning the item itself. that's quite useful for debugging.  however, @item is truthy, so I don't think it causes any other changes to go back to returning it

app/models/bulkrax/csv_entry.rb
```diff
@@ -49,12 +51,26 @@ def matcher_class
       Bulkrax::CsvMatcher
     end
 
+    def collections_created?
+      return true if record['collection'].blank?
+      record['collection'].split(/\s*[:;|]\s*/).length == self.collection_ids.length
+    end
+
+    def collections
+      return self.collection_ids if collections_created?
```
orangewolf: its still odd to me for a method called `collections` to return an array of ids and not the collections themselves.

orangewolf:
> @orangewolf latest commit addresses a lot of the above and includes an alternative to the lock. I'm not convinced the existing lock mechanism is the right thing. What I've put in instead is a check that the collections exist, and a job reschedule if they don't.
> 
> I've done more testing with this, and unearthed a few bugs, now fixed.
> 
> This PR has turned into a bit of a monster.
> 
> Questions:
> 
> * set a default collection and use it when OAI 'All' and OAI record doesn't include setSpec
> * use Hyrax::Lockable, or my collections_created? && reschedule (add a wait to that?)

I like both the Lockable and the new solution, but I'm leaning on the new one for 2 reasons. It is very specific - we don't create works until the collection exists... "oh your collection exists now? then go right ahead and build". That seems good to me. Also it means Bulkrax doesn't care about Redis, though I've been digging around more and I think that Hyrax may have a couple hidden dependencies on Redis that are not very clearly laid out.

Let's talk about the default collection thing tomorrow. I'm in and going at a humane hour and not in unavoidable meetings until high noon.

app/parsers/bulkrax/csv_parser.rb
orangewolf: I think this is still an open question?

app/models/bulkrax/oai_entry.rb
geekscruff: I've renamed to `retrieve_or_create_collection_ids` ... a bit of a mouthful, but it's clearer I think

app/models/bulkrax/csv_entry.rb
geekscruff: I've renamed to `retrieve_or_create_collection_ids`

app/parsers/bulkrax/csv_parser.rb
geekscruff: It does work. Tested with CSV Parser creating one collection and three works.

app/models/bulkrax/oai_entry.rb
orangewolf: I see, we agree that there is not a default possible in the 'all' case where no set of collections is chosen on an individual entry.

app/models/bulkrax/entry.rb
```diff
@@ -13,39 +15,56 @@ class Entry < ApplicationRecord
...
       else
         status_info
-        self.collection_id = @item.id if @item.is_a?(Collection)
       end
       return @item
     end
 
-    def collection
-      @collection ||= Collection.find(self.collection_id) if self.collection_id
+    def retrieve_or_create_collection_ids
```
orangewolf: lets call this find_or_create_collection_ids to be Railsy

PR #22, opened 2019-07-22
Allow bulkrax engine to work correctly if you install migrations into…

bkiahstroud:
… the base app

Though the recommended procedure is to leave migrations in the engine. This is because common practice in the Rails community does not match what's in the Rails guide.


orangewolf:
@bkiahstroud literally the same day as this happened Chris asked me to make a change to samvera_persona. It turns out the syntax here is only compatible with Ruby 2.5+ Can you make the same adjustment we made here: https://github.com/samvera-labs/samvera-persona/commit/65b8a5a23d8d8ad1aa8ec13c6fd9d4246f7ad58c
to this pr?

PR #30, opened 2019-08-02
WIP: Exporter

geekscruff:
This PR add support for CSV Export, and a framework for export generally. It isn't complete (see the todo list below).

Design decisions
* use Entry for both export and import, using polymorphism so that an entry can be associated with an importer or an exporter (but not both)
* add willow_sword, currently just for Zip file building, but future uses might include extending Bulkrax to offering wrapping the SWORD endpoint, and using the Bag functionality
* use parsers for both import and export, with a field called export_supported? 

Discuss
* re-use of entry, sharing methods across importer/exporter in parser
* entry sharing build method - should be more 'factory' ?
* file download and writing efficiency and approach
* perform_now ... support offline exports with email download link
* use presenter instead of work itself?

To do
* javascript to choose collection or import based on export_from field
* add seed exporter for testing
* further work on rspec tests
* comments / documentation
* authorization

### DIFF
--- a/app/controllers/bulkrax/exporters_controller.rb
+++ b/app/controllers/bulkrax/exporters_controller.rb
@@ -1,0 +1,149 @@
+require_dependency "bulkrax/application_controller"
+
+module Bulkrax
+  class ExportersController < ApplicationController
+    include Hyrax::ThemedLayoutController
+
+    before_action :set_exporter, only: [:show, :edit, :update, :destroy]
+    with_themed_layout 'dashboard'
+
+    # GET /exporters
+    def index
+      add_breadcrumb t(:'hyrax.controls.home'), main_app.root_path
+      add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
+      add_breadcrumb 'Exporters', bulkrax.exporters_path
+      @exporters = Exporter.all
+    end
+
+    # GET /exporters/1
+    def show
+    end
+
+    # GET /exporters/new
+    def new
+      add_breadcrumb t(:'hyrax.controls.home'), main_app.root_path
+      add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
+      add_breadcrumb 'Exporters', bulkrax.exporters_path
+      @exporter = Exporter.new
+    end
+
+    # GET /exporters/1/edit
+    def edit
+      add_breadcrumb t(:'hyrax.controls.home'), main_app.root_path
+      add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
+      add_breadcrumb 'Exporters', bulkrax.exporters_path
+    end
+
+    # POST /exporters
+    def create
+      @exporter = Exporter.new(exporter_params)
+      field_mapping_params
+
+      if @exporter.save
+        if params[:commit] == 'Create and Export'
+          # Use perform now for export
+          Bulkrax::ExporterJob.perform_now(@exporter.id)
+        end
+        redirect_to exporters_path, notice: 'Exporter was successfully created.'
+      else
+        render :new
+      end
+    end
+
+    # PATCH/PUT /exporters/1
+    def update
+      field_mapping_params
+      if @exporter.update(exporter_params)
+        if params[:commit] == 'Update and Re-Export All Items'
+          Bulkrax::ExporterJob.perform_now(@exporter.id)
+        end
+        redirect_to exporters_path, notice: 'Exporter was successfully updated.'
+      else
+        render :edit
+      end
+    end
+
+    # DELETE /exporters/1
+    def destroy
+      @exporter.destroy
+      redirect_to exporters_url, notice: 'Exporter was successfully destroyed.'
+    end
+
+    # GET /exporters/1/download
+    def download
+      @exporter = Exporter.find(params[:exporter_id])
+      send_content
+    end
+
+    private
+      # Use callbacks to share common setup or constraints between actions.
+      def set_exporter
+        @exporter = Exporter.find(params[:id])
+      end
+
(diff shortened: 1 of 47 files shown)


geekscruff:
@orangewolf still a bit to do, but this is exporting metadata only and metadata and files to zip

It's big, so it would be good to talk through it, see comments above about design decisions and discussion points.

app/controllers/bulkrax/exporters_controller.rb
```diff
@@ -0,0 +1,149 @@
...
+
+    before_action :set_exporter, only: [:show, :edit, :update, :destroy]
+    with_themed_layout 'dashboard'
+
+    # GET /exporters
+    def index
+      add_breadcrumb t(:'hyrax.controls.home'), main_app.root_path
+      add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
+      add_breadcrumb 'Exporters', bulkrax.exporters_path
+      @exporters = Exporter.all
```
orangewolf: can we break paginating this and the importer in to its own ticket? this would be a good one for @jpengst or @bkiahstroud to pick up.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -52,14 +54,32 @@ def create_works
...
+        ImportWorkJob.perform_later(new_entry.id, current_importer_run.id)
         increment_counters(index)
       end
     rescue StandardError => e
       errors.add(:base, e.class.to_s.to_sym, message: e.message)
     end
 
+    def create_from_importer
+      importer = Bulkrax::Importer.find(importerexporter.export_source)
+      importer.entries.each do | entry |
```
orangewolf: `find_each` so that this doesn't grow memory too big

app/jobs/bulkrax/exporter_job.rb
```diff
@@ -0,0 +1,11 @@
+module Bulkrax
```
orangewolf: this doesn't seem to be called at the moment. am I missing it?

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -52,14 +54,32 @@ def create_works
...
+        new_entry = find_or_create_entry(entry_class, record[:source_identifier], 'Bulkrax::Importer', record.to_h.compact)
+        ImportWorkJob.perform_later(new_entry.id, current_importer_run.id)
         increment_counters(index)
       end
     rescue StandardError => e
       errors.add(:base, e.class.to_s.to_sym, message: e.message)
     end
 
+    def create_from_importer
+      importer = Bulkrax::Importer.find(importerexporter.export_source)
```
orangewolf: I think that this and its sister method below are what should be done by the ExportJob?

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -52,14 +54,32 @@ def create_works
...
+        query = "#{ActiveFedora.index_field_mapper.solr_name(Bulkrax.system_identifier_field)}:\"#{entry.identifier}\""
+        work_id = ActiveFedora::SolrService.query(query, fl: 'id', rows: 1).first['id']
+        new_entry = find_or_create_entry(entry_class, work_id, 'Bulkrax::Exporter')
+        Bulkrax::ExportWorkJob.perform_now(new_entry.id,  current_exporter_run.id)
+      end
+    end
+
+    def create_from_collection
+      work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}").map(&:id)
+      work_ids.each do | wid |
```
orangewolf: does find_each work here? if so we should probably do it.

db/migrate/20190729124607_create_bulkrax_exporters.rb
```diff
@@ -0,0 +1,19 @@
+class CreateBulkraxExporters < ActiveRecord::Migration[5.1]
+  def change
+    create_table :bulkrax_exporters do |t|
+      t.string :name
+#      t.string :admin_set_id
```
orangewolf: ?

db/migrate/20190731114016_change_importer_and_exporter_to_polymorphic.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeImporterAndExporterToPolymorphic < ActiveRecord::Migration[5.1]
+  def change
+    remove_reference :bulkrax_entries, :importer
+    add_reference :bulkrax_entries, :importerexporter, polymorphic: true, index: {:name => "index_bulkrax_entries_on_importerexporter_type_and_id"}
```
orangewolf: this is disassociates all existing entries. can we deconstruct the helpers to make it less so? I didn't look up the actual syntax, but I think it might be something like:

```
remove_index :bulkrax_entries_importer_id
rename_column :bulkrax_entries, :importer_id, :importerexporter_id
add_column :bulkrax_entries, :importerexporter_type, :string, default: 'Bulkrax::Importer'
add_index: :bulkrax_entries, [:importerexporter_id, :importerexporter_type], name: "index_bulkrax_entries_on_importerexporter_type_and_id"
```

lib/bulkrax.rb
```diff
@@ -2,15 +2,18 @@
...
       { name: "OAI - Dublin Core", class_name: "Bulkrax::OaiDcParser", partial: "oai_fields" },
       { name: "OAI - Qualified Dublin Core", class_name: "Bulkrax::OaiQualifiedDcParser", partial: "oai_fields" },
       { name: "CSV - Comma Separated Values", class_name: "Bulkrax::CsvParser", partial: "csv_fields" }
     ]
 
-    self.system_identifier_field = "source"
+    self.import_path = 'tmp/imports'
+    self.export_path = 'tmp/exports'
+
+    self.system_identifier_field = 'source'
```
orangewolf: use the source luke!

PR #31, opened 2019-08-02
remove atla specific code

geekscruff:
This PR removes ATLA specific code. An accompanying PR in ATLA will add the necessary code there.

It also:
* removes self.parser_fields from the parser - this doesn't appear to be used any more
* moves the identifier setting in the entry build to the top, I introduced a bug where the identifier was not being set (because of unnecessary cacheing).

### DIFF
--- a/app/models/bulkrax/concerns/has_local_processing.rb
+++ b/app/models/bulkrax/concerns/has_local_processing.rb
@@ -1,0 +1,3 @@
+module Bulkrax::Concerns::HasLocalProcessing
+	def add_local; end
+end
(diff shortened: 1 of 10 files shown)


app/models/bulkrax/concerns/has_local_processing.rb
```diff
@@ -0,0 +1,3 @@
+module Bulkrax::Concerns::HasLocalProcessing
+	def add_local; end
```
bkiahstroud: Extra tab space

orangewolf:
@geekscruff merge at will!

PR #37, opened 2019-08-09
CSV create collections - avoid duplicate calls

geekscruff:
In the CSV parser, the collection name will be repeated on each line, leading to multiple create collection jobs.

When running multiple concurrent jobs, these will run concurrently leading to errors in trying to create the collection at the same time.

This PR grabs a list of collections from the CSV and only creates unique ones.

### DIFF
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -20,21 +20,20 @@ module Bulkrax
       # does the CSV contain a collection column?
       return unless import_fields.include?(:collection)
 
-      records.each do |record|
-        next if record[:collection].blank?
+      # retrieve a list of unique collections
+      collections = records.map { |r| r[:collection] }.compact.uniq
 
-
+      collections.each do |collection_record|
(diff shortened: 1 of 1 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -20,21 +20,20 @@ def create_collections
...
 
         # split by ; |
-        record[:collection].split(/\s*[;|]\s*/).each do |collection|
+        collection_record.split(/\s*[;|]\s*/).each do |collection|
           metadata = {
             title: [collection],
             Bulkrax.system_identifier_field => [collection],
             visibility: 'open',
-            collection_type_gid: Hyrax::CollectionType.find_or_create_default_collection_type.gid,
-            contributing_institution: [record[:contributing_institution]].compact
```
bkiahstroud: @geekscruff I assume we're removing this because it's specific to ATLA?

geekscruff: yes, and it should be added by the add_local method in the entry now.

PR #42, opened 2019-08-29
Fixes a bug in collections_created?

geekscruff:
There was a bug in the logic of `Bulkrax::OaiEntry#collections_created?`. We were looking for sets.length to be the same as the length of collection_ids but if we are harvesting from only one set, but the record belongs to many in OAI, the logic breaks down. Instead, we should check for sets.length only when harvesting 'All', and otherwise look for one collection_id.

Whilst works can be in multiple collections, OAI entries should only have multiple collection_ids when running 'all'.


orangewolf:
Great catch @geekscruf

PR #43, opened 2019-09-05
Colletions created bugfix(2)

geekscruff:
There was still an issue with the logic, which should now be fixed:

* when a set is selected in the importer, ONE and only ONE collection is created, for that set
* entry.collection_ids will receive the identifier

* when 'all' is selected, a collection is created for ALL collections, if we can retrieve them from OAI for the record, in this case the entry.collection_ids will correspond to the number of sets

* when 'all' is selected, and we cannot retrieve sets from OAI for the record, no collections are created and entry.collection_ids will be 0

### DIFF
--- a/app/models/bulkrax/oai_entry.rb
+++ b/app/models/bulkrax/oai_entry.rb
@@ -47,8 +47,9 @@ module Bulkrax
     end
 
     def collections_created?
+      
       if parser.collection_name == 'all'
-        sets.blank? || (sets.length == self.collection_ids.length)
+        sets.blank? || (sets.exist? && sets.length == self.collection_ids.length)
       else
         self.collection_ids.length == 1
       end
--- a/app/models/bulkrax/oai_entry.rb
+++ b/app/models/bulkrax/oai_entry.rb
@@ -47,11 +47,10 @@ module Bulkrax
     end
 
     def collections_created?
-      
       if parser.collection_name == 'all'
-        sets.blank? || (sets.exist? && sets.length == self.collection_ids.length)
+        sets.blank? || (sets.present? && sets.size == self.collection_ids.size)
       else
-        self.collection_ids.length == 1
+        self.collection_ids.size == 1
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/models/bulkrax/oai_entry.rb
```diff
@@ -47,10 +47,9 @@ def build_metadata
     end
 
     def collections_created?
-      if sets.blank? || parser.collection_name != 'all'
-        self.collection_ids.length == 1
-      elsif parser.collection_name == 'all'
-        sets.length == self.collection_ids.length
+      
+      if parser.collection_name == 'all'
+        sets.blank? || (sets.exist? && sets.length == self.collection_ids.length)
```
orangewolf: I know length is an alias, but size is the more common ruby array count method. I also know you didn’t change it from before, but if you are ok with it, I might swap them

app/models/bulkrax/oai_entry.rb
```diff
@@ -62,16 +61,13 @@ def collections_created?
     def find_or_create_collection_ids
       return self.collection_ids if collections_created?
 
-      if sets.blank?
-        c = Collection.where(Bulkrax.system_identifier_field => parser.collection_name).first
+      if sets.blank? || parser.collection_name != 'all'
+        c = Collection.where(Bulkrax.system_identifier_field => importer.unique_collection_identifier(parser.collection_name)).first
+        self.collection_ids << c.id unless c.blank? || self.collection_ids.include?(c.id)
```
orangewolf: Let’s break this up in to more lines. I _think_ you are saying
```
if c.present? && !self.collection_ids.include?(c.id)
  self.collection_ids << c.id
end
```

But I got it wrong the first time and that makes me think I’ll get it wrong in the future ;-)

app/models/bulkrax/oai_entry.rb
```diff
@@ -62,16 +61,13 @@ def collections_created?
...
+      else # All - collections should exist for all sets
+        c = Collection.where(Bulkrax.system_identifier_field => importer.unique_collection_identifier(set.content)).first
```
orangewolf: This seems like a big change. Before we were adding every collection id in the all case. Now are are only adding one?

geekscruff: that's wrong, fixed now.

PR #46, opened 2019-09-27
make sure remote_urls are actually urls before trying to import them

orangewolf:


app/factories/bulkrax/object_factory.rb
```diff
@@ -135,12 +135,17 @@ def file_attributes
 
     def new_remote_files
       @new_remote_files ||= if attributes[:remote_files].present? && object.present? && object.file_sets.present?
-                              attributes[:remote_files].reject do |file|
-                                existing = object.file_sets.detect { |f| f.import_url && f.import_url == file[:url] }
-                                existing
+                              attributes[:remote_files].select do |file|
```
orangewolf: note that I flipped the logic to positive instead of rejection.

PR #50, opened 2019-10-28
Database and association changes for exporter

geekscruff:
This PR adds migrations and changes to models / parser to make way for the exporter work.

When the new migrations are run, they should update existing `Bulkrax::Entry objects`, moving the `importer_id` to `importerexporter_id` before removing the `importer_id` column.

In addition:

* fix `spec/lib/bulkrax.rb`
* fix `spec/models/bulkrax/csv_entry_spec.rb`
* remove simple_form dependency; allow rails 5.2.*

### DIFF
--- a/app/models/bulkrax/entry.rb
+++ b/app/models/bulkrax/entry.rb
@@ -3,7 +3,12 @@ module Bulkrax
     include Bulkrax::Concerns::HasMatchers
     include Bulkrax::Concerns::HasLocalProcessing
 
-    belongs_to :importer
+    # @deprecated Please use importerexporter instead
+    #custom_deprecator = ActiveSupport::Deprecation.new('next-release', 'bulkrax')
+    #ActiveSupport::Deprecation.deprecate_methods(Bulkrax::Entry, importer: :importerexporter, deprecator: custom_deprecator)
+    belongs_to :importer, required: false
+    
+    belongs_to :importerexporter, polymorphic: true
     serialize :parsed_metadata, JSON
     # do not serialize raw_metadata as so we can support xml or other formats
     # keep it raw.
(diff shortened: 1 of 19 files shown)


app/models/bulkrax/entry.rb
```diff
@@ -3,16 +3,20 @@ class Entry < ApplicationRecord
...
     serialize :parsed_metadata, JSON
     # do not serialize raw_metadata as so we can support xml or other formats
     # keep it raw.
     serialize :collection_ids, Array
 
     attr_accessor :all_attrs, :last_exception
 
-    delegate :parser, :mapping,
-             to: :importer
+    delegate :parser, :mapping, to: :importerexporter
```
orangewolf: if you are making this change, you are not really deprecating the above, you are just removing it.  I think that's ok, just worth pointing out.  What you may want to do is remove belongs_to :importer above but put it back in the migration by opening up the class in the migration file and adding it.

db/migrate/20190731114016_change_importer_and_exporter_to_polymorphic.rb
```diff
@@ -0,0 +1,13 @@
+class ChangeImporterAndExporterToPolymorphic < ActiveRecord::Migration[5.1]
```
orangewolf:
```ruby
module Bulkrax
  class Entry < ApplicationRecord
     belongs_to :importer
  end
end
```

Above first line

app/models/bulkrax/entry.rb
geekscruff: ahh, I didn't think of doing it like that, a much better way, although it was fun figuring out how to do the deprecation messages.

geekscruff: I'll merge and do a small additional PR

PR #53, opened 2019-10-29
CSV Exporter for Bulkrax

geekscruff:
This PR adds the framework for exporters and a CSV Exporter.

Design Choices:
* Make the Bulkrax::Entry polymorphic, so that it can be used for either Importer / Exporter (but not both at the same time).
* Extract out generic code into concerns

Also:
* Fix tests for controllers


app/assets/javascripts/bulkrax/exporters.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
```
orangewolf: lets remove this one

PR #61, opened 2019-10-31
Make work type configurable; add defaults

### LINKED ISSUE #56: Make Work Type Configurable
It would be nice to remove the existing object factory code and allow work types to be more configurable. At a minimum we'd like to have work types be settable by field in the parser and overall for a given importer.

geekscruff:
Fixes #56 

Adds a configuration for `default_work_type`, set in `lib/bulkrax.rb` to Work and in the template application initializer (`config/initializers/bulkrax.rb`) to the first registered curation concern.

Refactors the `factory_class` method in the Entry to look for the following:

1. 'work_type' in the parsed_metadata
2. 'work_type' in the fields_mapping
3. Bulkrax.default_work_type

Also:

Adds a couple of missing tests for the configuration, and fixes some tests in entry.rb that were not acutally testing anything.


orangewolf:
Two very tiny code clarity things. On the whole this is extremely well thought out!

One question I have - you still need a `ClassName`Factory to exist to use this. I wonder if that is really necessary or if we can work around it?

geekscruff:
@orangewolf thanks, those changes are done. Re the factories, I agree but I'd like to do that in a different PR if that's OK with you.

PR #65, opened 2019-11-04
Fixes for CSV Import

geekscruff:
Newer versions of rails need `require 'csv'` in the CSV parser.

There is an issue where an incomplete mapping will omit unmapped (but valid) fields. The only time fields should be omitted is if they are explicitly declared as excluded or not supported by hyrax.

### DIFF
--- a/app/models/concerns/bulkrax/has_matchers.rb
+++ b/app/models/concerns/bulkrax/has_matchers.rb
@@ -57,14 +57,10 @@ module Bulkrax
     # @param field [String] the importer field name
     # @return [Array] hyrax fields
     def field_to(field)
-      return [field] if mapping.blank?
-      # retrieve the mapping
-      fields = mapping.map {
-        |key,value| 
-        key if (value['from'] && value['from'].include?(field)) || key == field 
-      }.compact
-      return fields unless fields.blank?
-      return [field]
+      fields = mapping&.map { |key, value|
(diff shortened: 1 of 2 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -6,6 +6,7 @@ def self.export_supported?
     end
 
     def records(_opts = {})
+      require 'csv'
```
orangewolf: this should probably be at the top of the file style wise.

app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -58,11 +58,13 @@ def field_supported?(field)
...
-      
-      mapping.map {
+      # retrieve the mapping
+      fields = mapping.map {
         |key,value| 
         key if (value['from'] && value['from'].include?(field)) || key == field 
       }.compact
+      return fields unless fields.blank?
+      return [field]
     end
```
orangewolf: Could this be cleaned up as something like

```ruby
def field_to(field)
  fields = mapping&.map { |key, value|
    key if (value['from'] && value['from'].include?(field)) || key == field
  }&.compact
  return fields || [field]
end
```

orangewolf:
neither of these things are functional changes, just styles. if we need to get the bug fix out, I can live with them to fix later.

geekscruff:
@orangewolf fixed and merged :)

PR #75, opened 2019-11-10
Fix field_to 

geekscruff:
field_to was still returning an empty array


app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -60,6 +60,7 @@ def field_to(field)
       fields = mapping&.map { |key, value|
         key if (value['from'] && value['from'].include?(field)) || key == field
       }&.compact
+      fields = nil if fields.blank?
```
geekscruff: added this as otherwise it would still return an empty array

PR #80, opened 2019-11-18
Improvements to CSV Importer

geekscruff:
This PR:

* makes it explicit when a CSV won't import because of missing required fields - previously, this would silently fail with no error message. Now an error is raised and will be seen in sidekiq.
* adds a `#validate_import` method to the parser, which can be expanded on for the validation work. Here it is used to check for required fields in CSV.
* returns a sensible error message if Collection doesn't implement/index the sytem identifier field - previously a Solr error would be thrown which didn't easily identify the issue
* dries up the find_collection methods used by the entries
* sets the image application factory system identifier to be the Bulkrax system identifier (if different identifiers are supplied, we'll get duplicate records) - this will be removed in #77 
* refactors the ImportWorkJob and Entry#build method to make the code clearer and more logical; added tests for the Job

### DIFF
--- a/app/jobs/bulkrax/import_work_job.rb
+++ b/app/jobs/bulkrax/import_work_job.rb
@@ -25,7 +25,6 @@ module Bulkrax
 
     def reschedule(entry_id, run_id)
       ImportWorkJob.set(wait: 1.minutes).perform_later(entry_id, run_id)
-      true
     end
   end
 end
--- a/app/jobs/bulkrax/import_work_job.rb
+++ b/app/jobs/bulkrax/import_work_job.rb
@@ -4,22 +4,28 @@ module Bulkrax
 
     def perform(*args)
       entry = Entry.find(args[0])
-      build = entry.build
-      entry.save
-      if build == true
+      build_result = entry.build
+      if build_result.present?
+        entry.save!
         ImporterRun.find(args[1]).increment!(:processed_records)
-      elsif build == false && entry.last_exception.blank?
-        reschedule(entry.id, ImporterRun.find(args[1]).id)
-      elsif entry.last_exception.present?
-        raise entry.last_exception
-      end
-      rescue StandardError => e
+      else
+        # do not retry here because whatever parse error kept you from creating a work will likely
+        # keep preventing you from doing so.
+        entry.save!
         ImporterRun.find(args[1]).increment!(:failed_records)
+      end
+      rescue CollectionsCreatedError => e
+        reschedule(args[0], args[1])
+      # Exceptions here are not an issue with building the work.
+      # Those are caught seperately, these are more likely network, db or other unexpected issues.
+      # Note that these temporary type issues do not raise the failure count
+      rescue StandardError => e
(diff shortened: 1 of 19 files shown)


app/models/concerns/bulkrax/import_behavior.rb
```diff
@@ -7,13 +7,12 @@ def build_for_importer
       build_metadata
       return false unless collections_created?
       begin
-        @item = factory.run
+        factory.run
+        return true
       rescue StandardError => e
         status_info(e)
-      else
-        status_info
```
orangewolf: you are no longer clearing the errors once an entry succeeds

app/jobs/bulkrax/import_work_job.rb
```diff
@@ -4,23 +4,18 @@ class ImportWorkJob < ApplicationJob
 
     def perform(*args)
       entry = Entry.find(args[0])
-      begin
-        if entry.build.present?
-          entry.save
-        elsif entry.last_exception.blank?
-          reschedule(entry.id, ImporterRun.find(args[1]).id)
-        end
+      build = entry.build
```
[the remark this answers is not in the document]
orangewolf: what about putting build_for_import back how it was and doing something like this here instead?
```ruby
def perform(*args)
  entry = Entry.find(args[0])

  # reschedule in to the future if we are waiting for the collection to finish
  if !entry.collections_created?
    reschedule(entry.id, ImporterRun.find(args[1]).id)
    return true
  end

  if entry.build
    entry.save!
    ImporterRun.find(args[1]).increment!(:processed_records)
  else
    # do not retry here because whatever parse error kept you from creating a work will likely
    # keep preventing you from doing so.
    ImporterRun.find(args[1]).increment!(:failed_records)
  end

# Exceptions here are not an issue with building the work.
# Those are caught seperately, these are more likely network, db or other unexpected issues.
# Note that these temporary type issues do not raise the failure count
rescue StandardError => e
  raise e
end
```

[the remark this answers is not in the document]
geekscruff: I've refactored along the lines of things, could you have a look? It's better - I agree it was too complex before.

app/models/concerns/bulkrax/import_behavior.rb
geekscruff: reinstated

app/jobs/bulkrax/import_work_job.rb
```diff
@@ -4,27 +4,28 @@ class ImportWorkJob < ApplicationJob
...
+      # Exceptions here are not an issue with building the work.
+      # Those are caught seperately, these are more likely network, db or other unexpected issues.
+      # Note that these temporary type issues do not raise the failure count
+      rescue StandardError => e
+        raise e
     end
 
     def reschedule(entry_id, run_id)
       ImportWorkJob.set(wait: 1.minutes).perform_later(entry_id, run_id)
+      true
```
orangewolf: dont need this

PR #82, opened 2019-11-25
WIP: Generic Factory

orangewolf:
This is a proposal turbo charging ObjectFactory and for removing WorkFactory, CollectionFactory, ImageFactory and the need for per work type factories of any sort. This denormilzation allows the generic factory to happily make whatever concern type objects it needs to. 

This is as much more psudo code to show than finished work. Tests will need to be adjusted, readme needs to be updated and some careful checking needs to be done, but I think this is the right direction.


orangewolf:
this code was incorporated in to #77

PR #85, opened 2019-11-28
Bugfixes

geekscruff:
- Move build_metadata into the begin otherwise errors in build_metadata won't be caught
- Change CollectionsCreatedError to an Exception so it won't be caught as StandardError
- Add a default `validate_import` method to ApplicationParser
- Reinstate collections creation in object factory

### DIFF
--- a/app/factories/bulkrax/with_associated_collection.rb
+++ b/app/factories/bulkrax/with_associated_collection.rb
@@ -6,15 +6,15 @@ module Bulkrax
     # which is used by Hyrax::Actors::AddAsMemberOfCollectionsActor
     def create_attributes
       if attributes[:collection].present?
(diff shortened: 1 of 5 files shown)


app/factories/bulkrax/with_associated_collection.rb
```diff
@@ -6,31 +6,31 @@ module WithAssociatedCollection
     # which is used by Hyrax::Actors::AddAsMemberOfCollectionsActor
     def create_attributes
       if attributes[:collection].present?
-        super.except(:collections).merge(member_of_collections_attributes: {0 => {id: collection.id}})
+        transform_attributes.except(:collections).merge(member_of_collections_attributes: {0 => {id: collection.id}})
```
orangewolf: yep, good find

PR #88, opened 2019-12-04
i34 changed records to entries and added total collections and works

jpengst:
WIP. Still need to increment total_collection_records in increment_counters(index)


orangewolf:
@jpengst probably worth renaming `total_records` `total_work_entries`  (this will need a global search and replace and a migration) and `total_collection_records` to `total_collection_entries` (you can update your existing migration since it isn't pushed out yet.

PR #90, opened 2019-12-05
i36 added showpage breadcrumbs and minor styling

jpengst:
Image 2019-12-04 at 4 50 28 PM

Note: I'm not sure what `<p id="notice"> <%= notice %>` does and if it should be inside or outside the main body content.


geekscruff:
@jpengst notice is used to display notices that get passed back from the controller, like 'You have successfully logged in', so I doubt we really need it in the show page but for consistency let's leave it as is.

PR #91, opened 2019-12-05
I15 bagit importer - extend to parent-child relationships and collections for rdf

geekscruff:
Adds
* a parser in the application matcher for extracting the work type from a field called 'model'
* a config for specifying the field in the source data to used to identify children 
* a config for specifying the field in the source data to used to identify collections
* new method call on ImporterJob to create relationships
* new methods in the importer and parser
* new job ChildRelationshipsJob

This needs a lot more manual testing, but it has been tested reasonably thoroughly with Bagit, and has rspec tests for CSV.


app/factories/bulkrax/with_associated_collection.rb
```diff
@@ -5,8 +5,9 @@ module WithAssociatedCollection
     # Strip out the :collection key, and add the member_of_collection_ids,
```
orangewolf: is this file doing anything any more? It seems like this functionality is moved in to the object_factory and isn't included anywhere else now

app/jobs/bulkrax/import_work_collection_job.rb
```diff
@@ -10,7 +10,6 @@ def perform(*args)
         ImporterRun.find(args[1]).increment!(:processed_collections)
       rescue => e
         ImporterRun.find(args[1]).increment!(:failed_collections)
-        raise e
```
orangewolf: by not re-raising the job will complete and be deleted and never be re-run

app/matchers/bulkrax/application_matcher.rb
```diff
@@ -63,6 +63,28 @@ def parse_types(src)
       src.to_s.strip.titleize
     end
 
+    # Allow for mapping a model field to the work type or collection
+    def parse_model(src)
```
orangewolf: 👍

app/models/bulkrax/importer.rb
```diff
@@ -74,6 +74,10 @@ def import_collections
       parser.create_collections
     end
 
+    def add_parent_child_relationships
+      parser.create_parent_child_relationships
```
orangewolf: Is it better to do `importer.create_parent_child_relationships` (note the create) and then delegate it to the parser?

app/parsers/bulkrax/application_parser.rb
```diff
@@ -37,15 +38,94 @@ def records(opts = {})
...
+          return if children.length == 0
+          Rails.logger.warn("Adding #{children.length} children to parent entry #{parent.id} (expected #{value.length})")
+        end
+        parent_id = parent.id
+        child_entry_ids = children.map { |c| c.id }
+        ChildRelationshipsJob.perform_later(parent_id, child_entry_ids, current_importer_run.id)
+      end
+    end
+
+    def setup_parents
```
orangewolf: OO wise, I think it might make sense to do `def parents` and `@parents` instead of setup_parents and not persist them.

app/parsers/bulkrax/application_parser.rb
```diff
@@ -37,15 +38,94 @@ def records(opts = {})
...
+
+    def create_works
+      raise 'must be defined' if importer?
+    end
+
+    # Optional, only used by certain parsers
+    # Other parsers should override with a custom or empty method
+    # Will be skipped unless the record is a Hash
+    def create_parent_child_relationships
+      parents = setup_parents
```
orangewolf: see comment below on why I think these 2 lines can be deleted (since [].each does nothing and doesn't need to be skipped)

app/parsers/bulkrax/application_parser.rb
```diff
@@ -37,15 +38,94 @@ def records(opts = {})
...
+
+        # not finding the entries here indicates that the given identifiers are incorrect
+        # in that case we should log that
+        children = []
+        children = value.map do | child |
+          entry_class.where(
+            identifier: child,
+            importerexporter_id: importerexporter.id,
+            importerexporter_type: 'Bulkrax::Importer',
+          ).first
```
orangewolf: children will potentially contain `[Child, nil, Child]`

orangewolf: and if only 1 is nil, the count will be the same below

orangewolf: I think you need `child = entry_class.where...` and then deal with incrementing the error here where you know what it is.  Probably coupled with replacing map with each to build the children array as you go.

app/jobs/bulkrax/child_relationships_job.rb
```diff
@@ -0,0 +1,97 @@
...
+          end
+          work_parent_work_child(members_works) unless members_works.blank?
+        end
+      # Not all of the Works/Collections exist yet; reschedule
+      rescue Bulkrax::ChildWorksError
+        reschedule(args[0], args[1], args[2])
+      # Exceptions here are not an issue with adding the relationships.
+      # Those are caught seperately, these are more likely network, db or other unexpected issues.
+      # Note that these temporary type issues do not raise the failure count
+      rescue StandardError => e
```
orangewolf: this code is redudenant. if you remove `rescue StandardError => e; raise e` it will do exactly the same thing

app/parsers/bulkrax/application_parser.rb
geekscruff: does the additon of `.compact.uniq` solve this? (as per latest)

geekscruff:
@orangewolf sorry, I realise I hadn't done a master rebase (done that now). I've resolved all bar one review comment (just a small question on that). Do you want to have one more pass through this?

app/parsers/bulkrax/application_parser.rb
orangewolf: oh, I see, the compact does remove the nils, so yes, that seems ok.

PR #101, opened 2019-12-10
I17 browse everything

geekscruff:
Adds Browse Everything for file system (mount) and cloud file upload.

Features:
* For CSV, used only for the files, not the CSV itself, and so separate to to the file upload toggle
* For BagIt, as option alongside the others
* Selected files are listed, and the button turns green

Issues:
* perform_now in download job will not work for large uploads - needs a mechanism from stopping the importer from running until all downloads are done
* it is not clear how to get the directory structure of an upload, therefore we need to accept only zipped files for BagIt

Screenshot 2019-12-11 at 14 13 55

For Bagit - one upload / file path only

Screenshot 2019-12-11 at 14 14 18

### DIFF
--- a/app/assets/javascripts/bulkrax/importers.js.erb
+++ b/app/assets/javascripts/bulkrax/importers.js.erb
@@ -5,8 +5,8 @@ function prepBulkrax(event) {
   var refresh_button = $('.refresh-set-source')
   var base_url = $('#importer_parser_fields_base_url')
   var initial_base_url = base_url.val()
-  var file_path = $('#importer_parser_fields_import_file_path').val()
-  handleFileToggle(file_path)
+  var file_path_value = $('#importer_parser_fields_import_file_path').val()
+  handleFileToggle(file_path_value)
 
   // handle refreshing/loading of external sets via button click
   $('body').on('click', '.refresh-set-source', function(e) {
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/bulkrax/importers.js.erb
```diff
@@ -97,6 +140,10 @@ function handleSourceLoad(refresh_button, base_url, external_set_select) {
   })
 }
 
+function handleSelectedFiles() {
+  console.log('YO!')
```
orangewolf: yo indeed. is this just a no-op on purpose or something that is todo?

app/controllers/bulkrax/importers_controller.rb
```diff
@@ -42,15 +42,12 @@ def edit
 
     # POST /importers
     def create
-      file = params[:importer][:parser_fields].delete(:file)
+      file = params[:importer][:parser_fields].delete(:file) if params[:importer][:parser_fields].present?
```
orangewolf: Hash#delete of a missing key is nil, so a check is not needed here.

```
2.5.1 :001 > a = {one: 1, two: 2}
 => {:one=>1, :two=>2}
2.5.1 :002 > a.delete(:three)
 => nil
2.5.1 :003 > a.delete(:two)
 => 2
```

app/controllers/bulkrax/importers_controller.rb
```diff
@@ -106,6 +100,22 @@ def external_sets
...
+      def files_for_import(file, cloud_files)
+        return if file.blank? && cloud_files.blank?
+        if file.present?
+          @importer[:parser_fields]['import_file_path'] = write_import_file(file)
+        end
+        if cloud_files.present?
+          # For BagIt, there will only be one bag, so we get the file_path back
+          # For Csv, we expect only file uploads, so we won't get the file_path back
+          target = retrieve_files(cloud_files)
+          @importer[:parser_fields]['import_file_path'] = target unless target.blank?
```
orangewolf: what happens with @import[:parser_fields]['import_file_path'] is blank on retrieve_files?

app/controllers/bulkrax/importers_controller.rb
```diff
@@ -128,7 +138,41 @@ def set_importer
...
       # Only allow a trusted parameter "white list" through.
       def importer_params
-        params.require(:importer).permit(:name, :admin_set_id, :user_id, :frequency, :parser_klass, :limit, field_mapping: {}, parser_fields: {})
+        params.require(:importer).permit(:name, :admin_set_id, :user_id, :frequency, :parser_klass, :limit, :selected_files, field_mapping: {}, parser_fields: {})
+      end
+
+      # @todo - investigate getting directory structure
+      # @todo - investigate using perform_later, and haveing the importer check for
+      #   DownloadCloudFileJob before it starts
+      def retrieve_files(files)
```
orangewolf: it feels like retrieve_files should move to a method on the parser, which would give you a simpler way to handle the different importer needs.

having said that, the files.each_pair code would work just as well for the case of 1 as it does for the case of many and could be simplified to be the same in both cases, leaving you with only the file path logic as different.

app/controllers/bulkrax/importers_controller.rb
geekscruff: For CSV we expect import_file_path to already be set - I've improved the comment.

app/controllers/bulkrax/importers_controller.rb
geekscruff: I've refactored and tried to make the code a little clearer.

geekscruff:
In the course of testing I ended up refactoring the validate_importer method, changing it to valid_import? then logging any messages on the importer, rather than having them raised to sidekiq for a pointless retry. This needs more work, as these errors aren't persisted or surfaced anywhere. But that could be done during the validation work.

It came up during this work in testing.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -39,10 +39,11 @@ def required_elements
       %w[title source_identifier]
     end
 
-    def validate_import
-      raise "Missing required elements, required elements are: #{required_elements.join(', ')}" unless required_elements?(import_fields)
-      # check file_paths; this will raise an error if any files are missing
-      true if file_paths.present?
+    def valid_import?
```
geekscruff: see comment, changing this to always return true or false and handling the error otherwise we end up with pointless retries.

app/models/bulkrax/csv_entry.rb
```diff
@@ -57,7 +57,7 @@ def build_metadata
       end
 
       # construct full file path
-      self.parsed_metadata['file'] = self.parsed_metadata['file'].split(/\s*[:;|]\s*/).map {|f| file_path(f)} if self.parsed_metadata['file'].present?
+      self.parsed_metadata['file'] = record['file'].split(/\s*[:;|]\s*/).map {|f| file_path(f)} if self.record['file'].present?
```
orangewolf: inconsistant self - `record['file']` or `self.record['file']`

PR #113, opened 2019-12-19
JSON API and CLI

geekscruff:
This PR:

* adds a JSON api to the importers for create / update / delete
* adds a CLI to create / update an importer via the JSON API
* uses Token authentication for JSON API
* adds documentation to the [wiki](https://github.com/samvera-labs/bulkrax/wiki)

TODO - add more tests


orangewolf:
the more I think about this, especially in light of terraform and kubernetes I really think "this is the way"

PR #116, opened 2019-12-23
Small fixes to object factory

geekscruff:
* destroy_existing_files nowbeing called on replace_files == true
* use object, rather than @object apart from when we're setting @object
* re-raise the CollectionsCreatedError

### DIFF
--- a/app/models/concerns/bulkrax/import_behavior.rb
+++ b/app/models/concerns/bulkrax/import_behavior.rb
@@ -9,7 +9,7 @@ module Bulkrax
         build_metadata
         raise CollectionsCreatedError unless collections_created?
         @item = factory.run
-      rescue RSolr::Error::Http => e
+      rescue RSolr::Error::Http, CollectionsCreatedError => e
         raise e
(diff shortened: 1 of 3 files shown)


app/models/concerns/bulkrax/import_behavior.rb
```diff
@@ -9,7 +9,7 @@ def build_for_importer
         build_metadata
         raise CollectionsCreatedError unless collections_created?
         @item = factory.run
-      rescue RSolr::Error::Http => e
+      rescue RSolr::Error::Http, CollectionsCreatedError => e
```
bkiahstroud: @geekscruff Should this be `Bulkrax::CollectionsCreatedError`, similar to https://github.com/samvera-labs/bulkrax/pull/116/files#diff-4d6669f1402d7a1df9781a44b4f9d438R18?

geekscruff: I would have thought so, I'll do some testing in case I did that for a reason.

PR #117, opened 2019-12-23
Collection and Work entries showpage

jpengst:
Added showpage for individual entries.  
TODO: Still need to add/connect Exporter showpage to entries showpage in a future ticket.


app/controllers/bulkrax/entries_controller.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require_dependency "bulkrax/application_controller"
+require_dependency "oai"
+require 'fileutils'
+
+module Bulkrax
+  class EntriesController < ApplicationController
```
bkiahstroud: Is line 3 necessary if you're  inheriting from ApplicationController on line 8?

app/controllers/bulkrax/entries_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+    # GET /importers/1/entries/1
+    def show
+      @importer = Importer.find(params[:importer_id])
+      @entry = Entry.find(params[:id])
+      add_breadcrumb t(:'hyrax.controls.home'), main_app.root_path
+      add_breadcrumb t(:'hyrax.dashboard.breadcrumbs.admin'), hyrax.dashboard_path
+      add_breadcrumb 'Importers', bulkrax.importers_path
+      add_breadcrumb @importer.name, bulkrax.importer_path(@importer.id)
+      add_breadcrumb @entry.id
+      # render 'importers/entries/show.html.erb'
```
bkiahstroud: Is this comment necessary?

app/controllers/bulkrax/entries_controller.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
```
bkiahstroud: Rubocop is pleased 😉

app/controllers/bulkrax/entries_controller.rb
jpengst: Good catch, ty

app/controllers/bulkrax/entries_controller.rb
jpengst: @geekscruff Is line 3 necessary?

geekscruff: Probably not, I'll have a test.

app/views/bulkrax/entries/show.html.erb
```diff
@@ -0,0 +1,56 @@
...
+      <strong>Entry ID:</strong>
+      <%= @entry.id %>
+    </p>
+
+    <p>
+      <strong>Type:</strong>
+      <%= @entry.factory_class %>
+    </p>
+
+    <p>
```
geekscruff: for raw_metadata and parsed_metadata, could you format the hash like the importer ones ... I'd suggest checking whether the data is a hash first (as for some entries it could be a blob of text or something else).

So, 

```
if x.present? && x.is_a?(Hash)
  format it
else
  just print it
end
```

app/views/bulkrax/importers/show.html.erb
```diff
@@ -82,7 +82,7 @@
               <td><span class="glyphicon glyphicon-remove" style="color: red;"></span> <%= e.status %></td>
             <% end %>
             <td><%= e.status_at %></td>
-            <td>TODO: Actions</td>
+            <td><%= link_to raw('<span class="glyphicon glyphicon-info-sign"></span>'), bulkrax.importer_entry_path(@importer.id, e.id) %></td>
```
geekscruff: could you also link the identifier to the entry (both collection and work)?

app/views/bulkrax/entries/show.html.erb
```diff
@@ -0,0 +1,56 @@
...
+      <strong>Raw Metadata:</strong>
+      <%= @entry.raw_metadata %>
+    </p>
+
+    <p>
+      <strong>Parsed Metadata:</strong>
+      <%= @entry.parsed_metadata %>
```
geekscruff: for parsed_metadata, could you format the hash like the importer ones?

PR #127, opened 2020-01-06
Importer index page counters improvements

geekscruff:
This PR does the following ensures that both collections and works are counted in the enqueued, processed and failed columns. It also adds the total collections to the `current_importer_run.create!` call and streamlines `increment_counters`.

### DIFF
--- a/app/models/concerns/bulkrax/importer_exporter_behavior.rb
+++ b/app/models/concerns/bulkrax/importer_exporter_behavior.rb
@@ -17,13 +17,7 @@ module Bulkrax
     end
 
     def increment_counters(index)
-      current_importer_run.total_work_entries = if limit.to_i.positive?
-                                                  limit
-                                                elsif parser.total.positive?
-                                                  parser.total
-                                                else
-                                                  index + 1
-                                                end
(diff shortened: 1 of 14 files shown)


app/models/concerns/bulkrax/importer_exporter_behavior.rb
```diff
@@ -17,13 +21,7 @@ def next_import_at
...
 
     def increment_counters(index)
-      current_importer_run.total_work_entries = if limit.to_i.positive?
-                                                  limit
-                                                elsif parser.total.positive?
-                                                  parser.total
-                                                else
-                                                  index + 1
-                                                end
+      current_importer_run.total_work_entries = index + 1 unless limit.to_i.positive? || parser.total.positive?
```
bkiahstroud: @geekscruff If this line hits the `unless`, then wouldn't `current_importer_run.total_work_entries` (which sounds like an int) be assigned a boolean? Is that what we want?

geekscruff: Good spot!

bkiahstroud: Ahh I see, that makes sense

PR #132, opened 2020-01-13
Create and Validate

jpengst:
Add a 'Create and Validate' feature that validates the importer before importing the data. 

This now:
* creates the importer and entries but not the works themselves
* offers discard and continue, discard deletes the importer and entries, discard re-runs it with work creation re-enabled

It is light on specs. I've added basic specs to the controller.

### DIFF
--- a/app/controllers/bulkrax/importers_controller.rb
+++ b/app/controllers/bulkrax/importers_controller.rb
@@ -118,7 +118,7 @@ module Bulkrax
         files_for_import(file, cloud_files) unless file.nil? && cloud_files.nil?
         # do not perform the import
         if params[:commit] == 'Update Importer'
-          # do nothing
+        # do nothing
         # OAI-only - selective re-harvest
         elsif params[:commit] == 'Update and Harvest Updated Items'
           Bulkrax::ImporterJob.perform_later(@importer.id, true)
@@ -131,7 +131,11 @@ module Bulkrax
         else
           Bulkrax::ImporterJob.perform_later(@importer.id)
         end
+<<<<<<< HEAD
         
+=======
+
+>>>>>>> 4eacfc4... rudimentary controller spec
         if api_request?
           json_response('updated', :ok, 'Importer was successfully updated.')
         else
--- a/app/controllers/bulkrax/importers_controller.rb
+++ b/app/controllers/bulkrax/importers_controller.rb
@@ -82,14 +82,13 @@ module Bulkrax
         files_for_import(file, cloud_files)
         if params[:commit] == 'Create and Import'
           Bulkrax::ImporterJob.perform_later(@importer.id)
-          if api_request?
-            json_response('create', :created, 'Importer was successfully created.')
-          else
-            redirect_to importers_path, notice: 'Importer was successfully created.'
-          end
         elsif params[:commit] == 'Create and Validate'
           Bulkrax::ImporterJob.perform_now(@importer.id)
-          redirect_to importer_path(@importer), notice: 'Importer was successfully validated.'
+        end
+        if api_request?
+          json_response('create', :created, 'Importer was successfully created.')
+        else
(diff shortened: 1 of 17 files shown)


app/controllers/bulkrax/importers_controller.rb
```diff
@@ -52,8 +52,16 @@ def create
       field_mapping_params
       if @importer.save
         files_for_import(file, cloud_files)
-        Bulkrax::ImporterJob.perform_later(@importer.id) if params[:commit] == 'Create and Import'
-        redirect_to importers_path, notice: 'Importer was successfully created.'
+        if params[:commit] == 'Create and Import'
+          Bulkrax::ImporterJob.perform_later(@importer.id)
+          redirect_to importers_path, notice: 'Importer was successfully created.'
+        elsif params[:commit] == 'Create and Validate'
+          validate_only = true
```
geekscruff: we need to set `@importer.validate_only = true` before the save action, but only if `params[:commit] == 'Create and Validate'`

so move line 59 to before 53, and make it

`@importer.validate_only = true if params[:commit] == 'Create and Validate'`

I would expect ` redirect_to importers_path` to work

app/jobs/bulkrax/importer_job.rb
```diff
@@ -4,7 +4,7 @@ module Bulkrax
   class ImporterJob < ApplicationJob
     queue_as :import
 
-    def perform(importer_id, only_updates_since_last_import = false)
+    def perform(importer_id, only_updates_since_last_import = false, validate_only = false)
```
geekscruff: actually, don't do this, we'll use the validate_only field on the importer itself

app/controllers/bulkrax/importers_controller.rb
```diff
@@ -123,6 +131,11 @@ def update
         else
           Bulkrax::ImporterJob.perform_later(@importer.id)
         end
+<<<<<<< HEAD
+        
+=======
+
+>>>>>>> 4eacfc4... rudimentary controller spec
```
bkiahstroud: @geekscruff

spec/controllers/bulkrax/importers_controller_spec.rb
```diff
@@ -152,6 +161,43 @@ def authenticate_user!
...
+      it 'calls update' do
+        expect(controller).to receive(:update)
+<<<<<<< HEAD
+        put :continue, params: { importer_id: importer.to_param, }, session: valid_session
+      end
+
+=======
+        put :continue, params: { importer_id: importer.to_param }, session: valid_session
+      end
+>>>>>>> 4eacfc4... rudimentary controller spec
```
bkiahstroud: @geekscruff More merge conflict stuff

app/controllers/bulkrax/importers_controller.rb
```diff
@@ -77,9 +77,14 @@ def create
...
+      @importer.validate_only = true if params[:commit] == 'Create and Validate'
       if @importer.save
         files_for_import(file, cloud_files)
-        Bulkrax::ImporterJob.perform_later(@importer.id) if params[:commit] == 'Create and Import'
+        if params[:commit] == 'Create and Import'
+          Bulkrax::ImporterJob.perform_later(@importer.id)
+        elsif params[:commit] == 'Create and Validate'
+          Bulkrax::ImporterJob.perform_now(@importer.id)
+        end
         if api_request?
```
bkiahstroud: @geekscruff Would refactoring this to use `@importer.parser.perform_method` work? If so, it would nix the need for the conditional entirely. 
E.g. 
```ruby
      if @importer.save
        files_for_import(file, cloud_files)
        Bulkrax::ImporterJob.send(@importer.parser.perform_method, @importer.id)
        if api_request?
```

geekscruff: yes!

Issue #70, opened 2019-11-08
Rubocop / Bixby


geekscruff:
Rubocop and bixby are part of the bulkrax codebase, but have not been run over the code as yet. The task is to:

(To be completed by #107) 
- [x] set-up rubocop and bixby
- [x] run `rubocop -a` to autocorrect
- [x] run `rubocop --auto-gen-config` to analyse errors

(To be completed by #114 and others)
- [x] fix error or add enable / disable to skip

Note - as of 8/11/2019 rubocop would not run against bulkrax with bixby due to lots of errors; needs analysis

Part of #44

bkiahstroud:
I can't self-assign this to me, but I'd like to tackle it @geekscruff

orangewolf:
this was completed

Issue #135, opened 2020-01-23
Fix failing ChildRelationshipsJob spec


bkiahstroud:
### Steps to Reproduce: 

1. Run specs 

1. Trouble spec should fail 

### Trace: 
```
Failures:

  1) Bulkrax::ChildRelationshipsJob error handling reschedule does not increment failed children
     Failure/Error: @child_entries = args[1].map { |e| Entry.find(e) }

     ActiveRecord::RecordNotFound:
       Couldn't find Bulkrax::Entry with 'id'=2
     # ./app/jobs/bulkrax/child_relationships_job.rb:10:in `block in perform'
     # ./app/jobs/bulkrax/child_relationships_job.rb:10:in `map'
     # ./app/jobs/bulkrax/child_relationships_job.rb:10:in `perform'
     # ./spec/jobs/bulkrax/child_relationships_job_spec.rb:181:in `block (4 levels) in <module:Bulkrax>'
```

PR #136, opened 2020-01-23
i70d - rubocop fixes for app, lib, and some spec directories

bkiahstroud:
#70


.rubocop_todo.yml
```diff
@@ -158,21 +32,21 @@ Metrics/ClassLength:
...
 Metrics/MethodLength:
-  Max: 130
+  Max: 22
 
-# Offense count: 4
+# Offense count: 1
 # Configuration parameters: CountComments.
 Metrics/ModuleLength:
-  Max: 158
+  Max: 104
```
bkiahstroud: Why did all these `Metrics` maximums change?

bkiahstroud: Follow up: I've played around with this a little bit and can't find a definitive reason why these changed. A theory is that, as offense counts have gone down, max Metric values have as well as their worse offenders have been resolved/ignored. If these Metric cops become an issue in the future, we can just alter them in the `.rubocop.yml` file. Thus, this PR is ready to be moved forward.

bkiahstroud:
@geekscruff No new specs are failing (the known failing spec still does, captured here: https://github.com/samvera-labs/bulkrax/issues/135)

PR #150, opened 2020-02-23
Bagit with CSV - add relative files

geekscruff:
This PR extends the BagIt parser to gather the files relative to the metadata file (ie. in the same folder and any subfolders) and add those to the raw_metadata['file']. This is skipped if the CSV itself has a file column.

This brings the CSVEntry into line with the RDFEntry.

TODO:
* this code has been only *lightly* tested, and tests have not been added


orangewolf:
Let’s block this for the moment. I think a bagit based implementation is a needed here

PR #152, opened 2020-02-25
BagIt files and CSV traversal

orangewolf:
~~1) couple of small fixes for CSV and importer show page~~ (merged already)

2) I got in to a state where an array was being passed in. This code protects from that

3) Bagit CSV updates
  a) I really think that if any subdirectories are valid bags that the import should commence. Requiring all bags to be valid feels like a super high bar and doesn't really allow for incremental fixing which is a big part of the Oregon Digital work. Plus, who is to say that there are no valid reasons to have a subdirectory that isn't a bag?

  b) Get only bag files, not all files from the bag. Includes files for CSV as well as RDF.

### DIFF
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -122,13 +122,14 @@ module Bulkrax
 
     # Gather the paths to all bags; skip any stray files
     def bag_paths
-      if bag?(import_file_path)
-        [import_file_path]
-      elsif bags?(import_file_path)
-        Dir.glob("#{import_file_path}/*").reject { |d| File.file?(d) }
-      else
-        raise 'No valid bags found'
-      end
+      result = if bag?(import_file_path)
+                 [import_file_path]
+               else
+                 Dir.glob("#{import_file_path}/*").select { |d| bag?(d) }
+               end
+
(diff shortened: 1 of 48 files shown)


app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -33,10 +33,14 @@ def import_fields
...
     def records(_opts = {})
-      raise 'No metadata files were found' if metadata_paths.blank?
-      @records ||= metadata_paths.map do |path|
+      raise 'No BagIt records were found' if bags.blank?
+      @records ||= bags.map do |bag|
+        path = metadata_path(bag)
+        raise 'No metadata files were found' if path.blank?
         data = entry_class.read_data(path)
-        entry_class.data_for_entry(data, path)
+        data = entry_class.data_for_entry(data, path)
```
bkiahstroud: `data` assigned twice

geekscruff: It's being re-assigned once it's being used so it's fine, but maybe a different variable name would be sensible here?

orangewolf: I kind of want the old value to be garbage collected as it can be significant

PR #156, opened 2020-02-28
Fixes bug in cloud params

geekscruff:


app/controllers/bulkrax/importers_controller.rb
```diff
@@ -74,7 +74,7 @@ def create
         return return_json_response unless valid_create_params?
       end
       file = params[:importer][:parser_fields].delete(:file)
-      cloud_files = params.delete(:selected_files)
+      cloud_files = params[:importer].delete(:selected_files)
```
orangewolf: does this needs a nil check? `params[:importer]&.delete(:selected_files)`

geekscruff: Importer will never be nil so I think we're fine. The method that receives the cloud_files checks for nil.

PR #172, opened 2020-03-13
add rubocop binstub

bkiahstroud:
Ensures `rubocop` always runs the correct version. The main motivation for this was so that linting tools work correctly inside text editors.


Gemfile.lock
```diff
@@ -1,5 +1,5 @@
 GIT
-  remote: https://github.com/notch8/willow_sword.git
+  remote: git://github.com/notch8/willow_sword.git
```
orangewolf: moves from secure to insecure

orangewolf: not a big deal, just something to note

PR #177, opened 2020-03-19
Exporter show page

bkiahstroud:
Update the exporter show page in line with the importer show page (but without the Collections section). 

Screen Shot 2020-03-26 at 6 13 36 PM

- [x]  The Exporter show page looks like Importer show page; at the top of the page, all metadata from the Exporter should be shown. At the bottom, there will not be any collections.

- [x]  The Exporter index links through to the Exporter show page from the Exporter name and an `i` icon

- [x] There is a route for `exporter/x/entries/x` 

- [x]  The Entries on the Exporter show page link through to the Entry's show page via `identifier` and `i`con 

- [x]  The Entry show page has a link back to the Exporter (in the same way as the importer entries do)

### DIFF
--- a/app/views/bulkrax/entries/show.html.erb
+++ b/app/views/bulkrax/entries/show.html.erb
@@ -47,30 +47,38 @@
       <% end %>
     </p>
 
+    <%# TODO: i18n %>
     <p>
-      <% if @entry.last_succeeded_at != nil %>
-        <% if @importer.class == Bulkrax::Importer %>
+      <% if @importer.present? %>
+        <% if @entry.last_succeeded_at != nil %>
           <strong>Importer:</strong>
           <%= link_to @importer.name, importer_path(@importer) %>
         <% else %>
-        <%# TODO: Create Exporter showpage table for individual items in order to link to the item %>
-          <strong>Exporter Link: Showpage has not been setup yet.</strong>
+          <strong>Importer Link: Item has not been imported successfully</strong>
         <% end %>
-      <% else %>
-        <strong>Importer Link: Item has not been imported successfully</strong>
+      <% elsif @exporter.present? %>
+        <% if @entry.last_succeeded_at != nil %>
+          <strong>Exporter:</strong>
+          <%= link_to @exporter.name, exporter_path(@exporter) %>
+        <% else %>
+          <strong>Exporter Link: Item has not been exported successfully</strong>
+        <% end %>
       <% end %>
     </p>
 
-    <p>
-      <% factory_record = @entry.factory.find %>
-      <% if @entry.last_succeeded_at != nil && factory_record.present? %>
-        <strong><%= @entry.factory_class.to_s %> Link:</strong>
-        <% if @entry.factory_class.to_s == "Collection" %>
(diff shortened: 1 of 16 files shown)


orangewolf:
@bkiahstroud is this still a draft?

bkiahstroud:
@orangewolf Yes, I'll be picking it back up later this week.

app/views/bulkrax/entries/show.html.erb
```diff
@@ -36,41 +37,46 @@
...
+    <% if @importer.present? %>
+      <p>
+        <% factory_record = @entry.factory.find %>
+        <% if @entry.last_succeeded_at != nil && factory_record.present? %>
+          <strong><%= @entry.factory_class.to_s %> Link:</strong>
+          <% if @entry.factory_class.to_s == "Collection" %>
+            <%= link_to @entry.factory_class.to_s, hyrax.polymorphic_path(@entry.factory.find) %>
+          <% else %>
+            <%= link_to @entry.factory_class.to_s, main_app.polymorphic_path(@entry.factory.find) %>
+          <% end %>
```
bkiahstroud: **Question** 

Is this link (see here: https://share.getcloudapp.com/RBudbgNP) necessary on the Exporter show page? 

The current `Bulkrax::ObjectFactory` class only supports `Importers`, refactoring to that class would be necessary to get this to work for `Exporters`. Also, there is a custom `Bulkrax::OregonDigitalObjectFactory` override class in OD that would also need to be refactored.

orangewolf: I think linking to the actual item is pretty cool for closing the loop, I don't see anything that really prevents that code from being moved back in to the exporter.  The factory itself doesn't change, just the code at the bottom of Importerbehavior.  @geekscruff what do you think?

bkiahstroud: The problem I was seeing was that [this line at the bottom of ImporterBehavior](https://github.com/samvera-labs/bulkrax/blob/master/app/models/concerns/bulkrax/import_behavior.rb#L62) passes `#replace_files` as a param, which is undefined on Exporter. I tried a couple fixes (if it's an exporter, pass false instead of `#replace_files`, etc) when I realized there's an override `#factory` method that would also need to be changed [here](https://gitlab.com/notch8/oregon-digital/-/blob/bulkrax/app/models/concerns/bulkrax/has_local_processing.rb#L6). 

All that being said, what would the best way to fix this be?

geekscruff: Given the exporter is exporting a work, doesn't each entry know what that work is without needing to refer to the factory (which it shouldn't have to as it isn't creating/updating works)?

geekscruff: there's a method in exporter_bahaviour.rb called work - wouldn't that do it?

bkiahstroud: Thanks @geekscruff! https://github.com/samvera-labs/bulkrax/pull/177/commits/3e1c33d35cb7a63ced426beabc3825c09e91ae0f

PR #179, opened 2020-03-21
Functional jQuery DataTables on Importer and Exporter index views

bkiahstroud:
Adds the following features to the Importer and Exporter index tables: 

- Limit 

- Search 

- Pagination 

- Sorting (by column) 

As well as customizes the `language` used in the table to avoid confusion (by default, `DataTable` refers to rows as 'entries', which is its own concept in Bulkrax). 

Image 2020-03-20 at 6 04 22 PM


geekscruff:
This is a huge improvement. It conveys the sense thay bulkrax is capable of coping with lots of imports, which it is.

PR #180, opened 2020-03-23
Move file processing

orangewolf:
We had file adds for BagIt based stuff in 2 places - which is one too many


app/models/bulkrax/csv_entry.rb
```diff
@@ -18,15 +18,14 @@ def self.read_data(path)
...
-    def self.data_for_entry(data, path = nil)
+    def self.data_for_entry(data)
       # If a multi-line CSV data is passed, grab the first row
       data = data.first if data.is_a?(CSV::Table)
       raw_data = data.to_h
       # If the collection field mapping is not 'collection', add 'collection' - the parser needs it
       raw_data[:collection] = raw_data[collection_field.to_sym] if raw_data.keys.include?(collection_field.to_sym) && collection_field != 'collection'
       # If the children field mapping is not 'children', add 'children' - the parser needs it
       raw_data[:children] = raw_data[collection_field.to_sym] if raw_data.keys.include?(children_field.to_sym) && children_field != 'children'
-      raw_data[:file] = record_file_paths(path) if raw_data[:file].blank?
```
orangewolf: when bagit was used, this was causing files to be added twice (due to new bagit based adding in bagit parser).  when bagit was not used, this is where the weird "[]" were coming from.

app/models/bulkrax/xml_entry.rb
```diff
@@ -27,7 +27,6 @@ def self.data_for_entry(data, path = nil)
               Nokogiri::XML::Node::SaveOptions::NO_DECLARATION | Nokogiri::XML::Node::SaveOptions::NO_EMPTY_TAGS
           ).delete("\n").delete("\t").squeeze(' '), # Remove newlines, tabs, and extra whitespace
         collection: collections,
-        file: record_file_paths(path),
```
orangewolf: this is the one edit I'm least sure about - I think attach every file that's at this path is what this code was doing, which I don't think is right for the XML parser, but I'd love clarification

spec/models/bulkrax/csv_entry_spec.rb
```diff
@@ -38,14 +38,13 @@ module Bulkrax
...
 
-          allow(subject).to receive(:record).and_return('source_identifier' => '3', 'title' => 'some title', 'file' => './spec/fixtures/csv/test file.csv')
+          allow(subject).to receive(:record).and_return('source_identifier' => '3', 'title' => 'some title')
           allow(File).to receive(:exist?).with('./spec/fixtures/csv/test_file.csv').and_return(true)
         end
 
         it 'sets up the file_path and removes spaces from filenames' do
           subject.build
           expect(subject.status).to eq('succeeded')
-          expect(subject.parsed_metadata['file']).to eq(['./spec/fixtures/csv/test_file.csv'])
```
bkiahstroud: @orangewolf I believe the point of this spec was to make sure that if a file was uploaded that had spaces in the filename, that those spaces would get replaced with underscores. Is that no longer needed? If so, should the whole spec be removed?

orangewolf: Hmm, I wonder why it was broken then. I’ll look at it

orangewolf: you were right - this was a bug... now it is not

PR #183, opened 2020-03-26
show view should just display the field mapping - not try and change it

orangewolf:


Gemfile
```diff
@@ -20,7 +20,7 @@ gem "oai"
 gem 'rsolr', '>= 1.0'
 gem "rspec-rails"
 gem 'twitter-typeahead-rails', '0.11.1.pre.corejavascript'
-gem 'willow_sword', github: 'notch8/willow_sword'
+gem 'willow_sword', git: 'https://github.com/notch8/willow_sword.git'
```
orangewolf: this kept switching between git: and https: in the lock file. now it wont do that

PR #184, opened 2020-03-28
Rubocop GitHub Action Workflow

bkiahstroud:
**Sources:**

- https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows 

- https://github.com/actions/cache/issues/144#issuecomment-579323937

- https://github.com/actions/cache#Skipping-steps-based-on-cache-hit


bulkrax.gemspec
```diff
@@ -31,5 +31,4 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency 'sqlite3', '~> 1.3.13'
   s.add_development_dependency 'simplecov'
-  s.add_development_dependency 'bixby'
```
bkiahstroud, line 34: Unsure if this is an okay change to make, but moving `bixby` to its own `group` in the Gemfile really sped up the `bundle install`.

orangewolf: I'm good with it. Julie?

geekscruff: me too

PR #185, opened 2020-03-30
Various bug fixes and improvements

### LINKED ISSUE #182: CSV can't be re-imported
To do 'proper' round-tripping export requires source_identifier (the configured one).

Also if we don't have a system_identifier to start with, we also can't round trip.

geekscruff:
This PR:

* removes outdated information from the README and adds links to the wiki
* fixes a bug where the column headers in the export CSV are out of sync with the parsed_metadata being written to the line - fixes #182
* adds source_identifier to the export_metadata to better support round tripping; adds the work id to the system_identifier field of the work being exported, so that it can be round-tripped back in - fixes #182  
* removes the directory for the work when exporting files - this won't work when reimporting with cloud files
* adds support for zip upload of CSV and files
* adds toggle to raw_metadata and parsed_metadata on entry
* refactors mapping so that it is properly memoized


lib/bulkrax.rb
```diff
@@ -104,9 +108,8 @@ class << self
         "resource_type" => { from: ["type"], parsed: true },
         "remote_files" => { from: ["thumbnail_url"], parsed: true }
       },
-      "Bulkrax::CsvParser" => {
-        "remote_files" => { from: ["remote_files"], parsed: true }
-      },
+      # When empty, a default_field_mapping will be generated
+      "Bulkrax::CsvParser" => {},
```
geekscruff: I'm unsure why you added remote_files here @orangewolf ... but I don't see why it's needed? If the CSV contains a remote_files column, it will be imported. I suspect this could be specific to Atla?

app/controllers/bulkrax/importers_controller.rb
```diff
@@ -73,7 +73,7 @@ def create
       if api_request?
         return return_json_response unless valid_create_params?
       end
-      file = params.to_unsafe_h[:importer][:parser_fields].delete(:file)
+      file = params[:importer][:parser_fields].delete(:file)
```
geekscruff: @orangewolf using to_unsafe_h here means we end up with the whole tmpfile in the db, which we don't want. I know you needed this for cloud files, but is it really needed for file?

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -197,17 +198,20 @@ def file_paths
...
         r[:file].split(/\s*[:;|]\s*/).map do |f|
-          file = File.join(files_path, f.tr(' ', '_'))
-          return file if File.exist?(file)
-          raise "File #{file} does not exist"
+          file = File.join(path_to_files, f.tr(' ', '_'))
+          if File.exist?(file)
+            file
+          else
+            raise "File #{file} does not exist"
+          end
```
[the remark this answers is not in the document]
geekscruff: the return statement in there was causing problems, also wrapping it in if/else like this is also a little more readable - it's clear that the raise only happens if the file doesn't exist. I know rubocop is unhappy with me :(

app/controllers/bulkrax/importers_controller.rb
orangewolf: when I tested this file was also not working - did you try uploading with a file after this change? Another approach might be to make a method that removes the item from the params, then permits that now extracted item instead of calling `to_unsafe_h`. That's probably a better way to go for both items now that I think about it.

lib/bulkrax.rb
orangewolf: I added that because this code needs to always be called for remote files columns: `https://github.com/samvera-labs/bulkrax/blob/master/app/matchers/bulkrax/application_matcher.rb#L50` 

Leaving it in the matcher means you can still override it, but if you don't wrap the remote files in `{ url: #{value}}` then the files won't actually import. 

We could fix the spec (maybe set them to blank in the tests) or force remote_file to be parsed all the time (which I think is less explicit and harder to find).  You likely want it on by default for bagit and xml parser too.

geekscruff: Actually I was thinking about checking whether a parse method exists and setting parse to true within the lambda that generates the default mapping ... I think that would fix this in a cleaner way?

orangewolf: yep, that would work

orangewolf:
Seeing this failure - not sure if its a spec problem or a code problem. Thoughts?

```
 1) Bulkrax::Importer field_mapping oai_parser retrieves the default field mapping for oai_dc
     Failure/Error:
       expect(importer.mapping).to eq(
         "contributor" => { "excluded" => false, "from" => ["contributor"], "if" => nil, "parsed" => false, "split" => false },
         "coverage" => { "excluded" => false, "from" => ["coverage"], "if" => nil, "parsed" => false, "split" => false },
         "creator" => { "excluded" => false, "from" => ["creator"], "if" => nil, "parsed" => false, "split" => false },
         "date" => { "excluded" => false, "from" => ["date"], "if" => nil, "parsed" => false, "split" => false },
         "description" => { "excluded" => false, "from" => ["description"], "if" => nil, "parsed" => false, "split" => false },
         "format" => { "excluded" => false, "from" => ["format"], "if" => nil, "parsed" => false, "split" => false },
         "identifier" => { "excluded" => false, "from" => ["identifier"], "if" => nil, "parsed" => false, "split" => false },
         "language" => { "excluded" => false, "from" => ["language"], "if" => nil, "parsed" => false, "split" => false },
         "publisher" => { "excluded" => false, "from" => ["publisher"], "if" => nil, "parsed" => false, "split" => false },

       expected: {"contributor"=>{"excluded"=>false, "from"=>["contributor"], "if"=>nil, "parsed"=>false, "split"=>fal...=>false}, "type"=>{"excluded"=>false, "from"=>["type"], "if"=>nil, "parsed"=>false, "split"=>false}}
            got: {"contributor"=>{"excluded"=>false, "from"=>["contributor"], "if"=>nil, "parsed"=>false, "split"=>fal...=>false}, "type"=>{"excluded"=>false, "from"=>["type"], "if"=>nil, "parsed"=>false, "split"=>false}}

       (compared using ==)

       Diff:

       @@ -5,12 +5,12 @@
        "description" => {"excluded"=>false, "from"=>["description"], "if"=>nil, "parsed"=>false, "split"=>false},
        "format" => {"excluded"=>false, "from"=>["format"], "if"=>nil, "parsed"=>false, "split"=>false},
        "identifier" => {"excluded"=>false, "from"=>["identifier"], "if"=>nil, "parsed"=>false, "split"=>false},
       -"language" => {"excluded"=>false, "from"=>["language"], "if"=>nil, "parsed"=>false, "split"=>false},
       +"language" => {"excluded"=>false, "from"=>["language"], "if"=>nil, "parsed"=>true, "split"=>false},
        "publisher" => {"excluded"=>false, "from"=>["publisher"], "if"=>nil, "parsed"=>false, "split"=>false},
        "relation" => {"excluded"=>false, "from"=>["relation"], "if"=>nil, "parsed"=>false, "split"=>false},
        "rights" => {"excluded"=>false, "from"=>["rights"], "if"=>nil, "parsed"=>false, "split"=>false},
        "source" => {"excluded"=>false, "from"=>["source"], "if"=>nil, "parsed"=>false, "split"=>false},
       -"subject" => {"excluded"=>false, "from"=>["subject"], "if"=>nil, "parsed"=>false, "split"=>false},
       +"subject" => {"excluded"=>false, "from"=>["subject"], "if"=>nil, "parsed"=>true, "split"=>false},
        "title" => {"excluded"=>false, "from"=>["title"], "if"=>nil, "parsed"=>false, "split"=>false},
        "type" => {"excluded"=>false, "from"=>["type"], "if"=>nil, "parsed"=>false, "split"=>false},

     # ./spec/models/bulkrax/importer_spec.rb:51:in `block (4 levels) in <module:Bulkrax>'

Finished in 1 minute 39.18 seconds (files took 13.88 seconds to load)
207 examples, 1 failure, 7 pending
```

PR #187, opened 2020-04-06
we now detect parsed fields, so these need to be swapped

orangewolf:
I'm pretty sure this is the right direction on these. Doctor Julie can I get a consult?


geekscruff:
👍  ... but made another commit to make rubocop happy, so it's needs your review now @orangewolf

PR #194, opened 2020-05-04
was having issues with relating works to each other

orangewolf:
ended up doing some refactoring to find the bug

### DIFF
--- a/app/jobs/bulkrax/child_relationships_job.rb
+++ b/app/jobs/bulkrax/child_relationships_job.rb
@@ -13,7 +13,7 @@ module Bulkrax
       else
         work_membership
       end
-      # Not all of the Works/Collections exist yet; reschedule
+    # Not all of the Works/Collections exist yet; reschedule
     rescue Bulkrax::ChildWorksError
       reschedule(args[0], args[1], args[2])
     end
--- a/app/jobs/bulkrax/child_relationships_job.rb
+++ b/app/jobs/bulkrax/child_relationships_job.rb
@@ -6,87 +6,106 @@ module Bulkrax
     queue_as :import
 
     def perform(*args)
-      @entry = Entry.find(args[0])
-      @child_entries = args[1].map { |e| Entry.find(e) }
-      @child_works_hash = build_child_works_hash
-      @importer_run_id = args[2]
-      @user = @entry.importerexporter.user
+      @args = args
 
-      if @entry.factory_class == Collection
-        # add collection to works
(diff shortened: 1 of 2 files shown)


app/jobs/bulkrax/child_relationships_job.rb
```diff
@@ -6,87 +6,106 @@ class ChildRelationshipsJob < ApplicationJob
...
-        if members_works.length < @child_entries.length # rubocop:disable Style/IfUnlessModifier
-          Rails.logger.warn("Cannot add collections as children of works: #{(@child_entries.length - members_works.length)} collections were discarded for parent entry #{@entry.id} (of #{@child_entries.length})")
-        end
-        work_parent_work_child(members_works) unless members_works.blank?
+        work_membership
       end
-    # Not all of the Works/Collections exist yet; reschedule
+      # Not all of the Works/Collections exist yet; reschedule
     rescue Bulkrax::ChildWorksError
       reschedule(args[0], args[1], args[2])
```
bkiahstroud, line 18: Should instances of `args` here also be replaced with `@args`?

app/jobs/bulkrax/child_relationships_job.rb
```diff
@@ -6,87 +6,106 @@ class ChildRelationshipsJob < ApplicationJob
...
-        @child_works_hash.each_pair { |k, v| members_works << k if v[:class_name] != 'Collection' }
-        if members_works.length < @child_entries.length # rubocop:disable Style/IfUnlessModifier
-          Rails.logger.warn("Cannot add collections as children of works: #{(@child_entries.length - members_works.length)} collections were discarded for parent entry #{@entry.id} (of #{@child_entries.length})")
-        end
-        work_parent_work_child(members_works) unless members_works.blank?
+        work_membership
       end
-    # Not all of the Works/Collections exist yet; reschedule
+      # Not all of the Works/Collections exist yet; reschedule
     rescue Bulkrax::ChildWorksError
```
bkiahstroud, line 17: This comment appears to be referring to the `rescue`, so I think the original spacing was more clear.

app/jobs/bulkrax/child_relationships_job.rb
orangewolf: not really, just because its in the same method

app/jobs/bulkrax/child_relationships_job.rb
orangewolf: agreed

PR #199, opened 2020-06-04
add delete job, track records without source ids without creating ent…

orangewolf:
…ries, properly show invalid records


app/views/bulkrax/importers/show.html.erb
```diff
@@ -47,6 +47,9 @@
         <strong>Error:</strong> <%= @importer.last_error['error_class'] %> - <%= @importer.last_error['error_message'] %><br />
         <strong>Error Trace:</strong> <button id='err_toggle'>Toggle</button><br>
         <span id='error_trace' style='display: none'><%= @importer.last_error['error_trace'] %></span>
+      <% elsif @importer.importer_runs.last.invalid_records.present? %>
```
bkiahstroud, line 50: Would this throw a `undefined method for nil:NilClass` error if a user created a new importer without running it and then navigated to this view?

orangewolf: in which Kiah is 100% right =-)

PR #200, opened 2020-06-05
Add Spec Run To CI

orangewolf:
Super Green


orangewolf:
I thought automated specs would be a good PR #200 

@bkiahstroud I renamed the Rubocop to be lint / rubocop since "Kind of thing / Thing" made more sense to me. Let me know if you want that switched back.

.github/workflows/lint.yml
```diff
@@ -1,4 +1,4 @@
-name: Rubocop
+name: Lint
```
bkiahstroud, line 1: Extra whitespace.

.github/workflows/test.yml
```diff
@@ -0,0 +1,39 @@
...
+        id: ruby
+        with:
+          ruby-version: 2.6.5
+          cache-available: ${{ steps.cache.outputs.cache-hit == 'true' }}
+
+      - name: Setup sqlite3
+        run: sudo apt-get install libsqlite3-dev
+
+      - name: Install dependencies
+        run: bundle install
```
bkiahstroud, line 33: Do we need to bundle all the gems to run the specs? If not, should we flesh out the `:test` group in the Gemfile to allow us to target those specifically?

bkiahstroud: Bundling all the gems took 4 and a half minutes on the last run.

orangewolf: yes, you need them all, the stuff in dev/test that is slow is the stuff for running the specs.

.github/workflows/lint.yml
orangewolf: fixed

orangewolf:
yes you need the .gitkeep - it forces the directory to exist as git does not keep track of directories, only stuff in directories.

PR #237, opened 2020-09-14
Status overhaul

orangewolf:
Completely obsolete the old status structure and move to a more clear, explicit status setting.  Statuses are now tied to a given run, which means that it should be easier to keep status separate from previous runs and to show the status of old runs if we need to.


app/parsers/bulkrax/application_parser.rb
```diff
@@ -3,7 +3,7 @@
 module Bulkrax
   class ApplicationParser
     attr_accessor :importerexporter
-    delegate :only_updates, :limit, :current_exporter_run, :current_importer_run, :errors,
+    delegate :only_updates, :limit, :current_run, :current_run, :errors,
```
bkiahstroud, line 6: Duplicate `:current_run` delegations

spec/factories/bulkrax_importers.rb
```diff
@@ -37,6 +37,7 @@
     limit { 10 }
     parser_fields { { 'import_file_path' => 'spec/fixtures/csv/good.csv' } }
     field_mapping { {} }
+    after :create, &:current_run
```
bkiahstroud, line 40: @orangewolf Would it be better to move this to the model? Some kind of `after` hook that automatically creates a Run object after an Importer or Exporter is created instead of manually invoking `current_run` throughout the code?

app/jobs/bulkrax/child_relationships_job.rb
```diff
@@ -78,7 +78,6 @@ def work_child_collection_parent(work_id)
         ImporterRun.find(importer_run_id).increment!(:processed_children)
       rescue StandardError => e
         entry.status_info(e)
-        entry.save!
```
bkiahstroud, line 81: Do we still need to save `entry` here, even though we no longer need to in order to capture the status? Was the only purpose of these `save!`s to capture the error?

orangewolf: it was the only reason for the save, nothing else has changed in the entry at that point.

app/parsers/bulkrax/application_parser.rb
orangewolf: good catch

spec/factories/bulkrax_importers.rb
orangewolf: We have to do it on purpose. We create one during the ImportJob, and then pass it through the various import step.  Current_run makes a new record each session of runs.

bkiahstroud (approved):

PR #238, opened 2020-09-14
Remove extra required attribute from oai form input field

smithjp:
I was seeing a warning in the logs that there were 2 "required" keys here


orangewolf:
@smithjp does this change leave the fields are required or no required? The tag in the code you are changing was making it explicitly not required.

smithjp:
There is a second required key for that input field which specifies that the field is not required.  I was just removing one of them.  The field is still not required.
https://github.com/samvera-labs/bulkrax/pull/238/files#diff-6462b738e77193848f201030f3cfa24fR25

orangewolf:
Wow if I had expanded the code in the little window I would have seen that. Thank you very much for this =-)

PR #239, opened 2020-09-14
Remove extra create button from import form

smithjp:
I noticed that there were 3 "create" buttons on the importer form but only "Create and Validate" had a specified action in the controller, so the other two triggered the default action of creating and importing.  I think it would make sense to remove either "Create" or "Create and Import" or to add a third action in the controller.

### DIFF
--- a/app/views/bulkrax/importers/new.html.erb
+++ b/app/views/bulkrax/importers/new.html.erb
@@ -13,8 +13,6 @@
             |
             <%= form.button :submit, value: 'Create and Import', class: 'btn btn-primary' %>
             |
-            <%= form.button :submit, value: 'Create', class: 'btn btn-primary' %>
-            |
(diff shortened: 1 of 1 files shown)


orangewolf:
@smithjp Thank you! I think the bug is that "Create" with out Import is still importing records. Can you take a look at https://github.com/samvera-labs/bulkrax/pull/241 and see if it solves the issue so that we still have 3 viable options?

The intended behavior is:
Create - Just save, like making a draft almost
Create and Import - Save and then start the importer
Create and Validate - Save and run pre-flight validations

If you are good with that then I propose we close this PR and merge #241

smithjp:
I'm sorry, I didn't see that one.  I tested #241 locally, and it does solve the issue.  I agree that closing this PR and merging #241 is the way to go.

PR #240, opened 2020-09-15
Improve Exporter Show

orangewolf:
Bring it more in line with Importer Show.


orangewolf:
this code has all sense been replaced.

Issue #234, opened 2020-08-29
If original CSV causes a failure, the importer will continue to fail when csv is corrected


crisr15:
# Summary

When an importer is run with an improperly configured csv, later attempts to edit the importer and upload a corrected csv will fail, producing the same errors as the original file, despite corrections. Changing the name of the file sidesteps this.

alishaevn:
from present day testing this is no longer an issue. can I close this @crisr15 @orangewolf ?

orangewolf:
yep!

PR #242, opened 2020-09-15
Fix corrected entries functionality

smithjp:
Possibly related to #234 ?
When we were testing the csv importer locally, we noticed that the corrected entries function did not seem to be working.  After downloading the failed records, updating the records, and reuploading the csv, the records still failed with the original error messages.  After looking a little more closely, it appears that the records are being parsed during the validation step in the importer job (https://github.com/samvera-labs/bulkrax/blob/master/app/jobs/bulkrax/importer_job.rb#L15) without first setting the only_updates importer attribute.  Since the records are only parsed once (https://github.com/samvera-labs/bulkrax/blob/master/app/parsers/bulkrax/csv_parser.rb#L27), the corrected entries are never used.  I added a line in the importer job to set the only_updates attribute before the validation step and removed the only_updates option from the import_works method.  This seems to have fixed the problem for us.

### DIFF
--- a/app/jobs/bulkrax/importer_job.rb
+++ b/app/jobs/bulkrax/importer_job.rb
@@ -12,9 +12,10 @@ module Bulkrax
     end
 
     def import(importer, only_updates_since_last_import)
+      importer.only_updates = only_updates_since_last_import || false
       return unless importer.valid_import?
       importer.import_collections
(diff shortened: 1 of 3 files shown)


bkiahstroud (approved):
This looks good to me, thanks!

Issue #216, opened 2020-07-15
Add ability to set adminset


ShanaLMoore:
Bulkrax defaults to using the default admin set, despite selecting a different option from the form.

PR #243, opened 2020-09-18
Add admin_set_id to parsed_metadata

smithjp:
Currently, it seems that bulkrax defaults to the default admin set for imported works, ignoring the admin set selected in the form.  I added a method to add the admin set id to the parsed metadata for importers and added tests to verify that the admin set id is captured.  I believe this is related to #216


orangewolf (approved):
Looks great!

PR #253, opened 2020-10-20
Add row param to solr queries for exports

smithjp:
We have noticed that collection and work type exports seemed to be limited to 10 even with a form limit of 0 or blank. The ActiveFedora::SolrService.query method uses the solr default limit of 10 unless the rows parameter is used, so I added a rows parameter with a value close to the maximum allowed integer value.

### DIFF
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -116,7 +116,7 @@ module Bulkrax
     end
 
     def create_from_collection
-      work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}").map(&:id)
+      work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}", rows: 2_000_000_000).map(&:id)
       work_ids.each_with_index do |wid, index|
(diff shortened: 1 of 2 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -116,7 +116,7 @@ def create_from_importer
     end
 
     def create_from_collection
-      work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}").map(&:id)
+      work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}", rows: 2_000_000_000).map(&:id)
```
orangewolf, line 119: I think this is the right play, but whenever I see it all I can think is:

orangewolf:
@smithjp codewise this looks great - Rubocop and Rspec are both failing. If you have a moment to fix those up, I'd appreciate it, if not, just let me know and I'd be happy to pull the branch down and do it. Either is good for me, I just don't want to step on your code ;-)

smithjp:
Sorry about that!  I'll fix the errors.

bkiahstroud (approved):
Thanks for fixing those!

PR #254, opened 2020-11-09
Export from date range

sephirothkod:
allows for exporting within a date range, and visibility (status)

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -76,11 +76,20 @@ module Bulkrax
     end
 
     def build_export_metadata
+      hyrax_record
       make_round_trippable
       self.parsed_metadata = {}
       self.parsed_metadata['id'] = hyrax_record.id
       self.parsed_metadata[self.class.source_identifier_field] = hyrax_record.id
       self.parsed_metadata['model'] = hyrax_record.has_model.first
+      build_mapping_metadata
+      unless hyrax_record.is_a?(Collection)
+        self.parsed_metadata['file'] = hyrax_record.file_sets.map { |fs| filename(fs).to_s unless filename(fs).blank? }.compact.join('; ')
+      end
+      self.parsed_metadata
+    end
+
+    def build_mapping_metadata
       mapping.each do |key, value|
         next if Bulkrax.reserved_properties.include?(key) && !field_supported?(key)
         next unless hyrax_record.respond_to?(value['from']&.first.to_s)
@@ -91,10 +100,6 @@ module Bulkrax
           self.parsed_metadata[key] = prepare_export_data(data)
         end
       end
-      unless hyrax_record.is_a?(Collection)
-        self.parsed_metadata['file'] = hyrax_record.file_sets.map { |fs| filename(fs).to_s unless filename(fs).blank? }.compact.join('; ')
-      end
-      self.parsed_metadata
     end
 
     def prepare_export_data(datum)
@@ -137,7 +142,7 @@ module Bulkrax
           self.collection_ids << c.id unless c.blank? || self.collection_ids.include?(c.id)
         end
       end
-      return self.collection_ids
+      self.collection_ids
     end
 
     def required_elements?(keys)
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -83,8 +83,8 @@ module Bulkrax
       self.parsed_metadata['model'] = hyrax_record.has_model.first
       mapping.each do |key, value|
         next if Bulkrax.reserved_properties.include?(key) && !field_supported?(key)
-        next unless hyrax_record.respond_to?(key)
-        data = hyrax_record.send(key)
+        next unless hyrax_record.respond_to?(value['from']&.first.to_s)
+        data = hyrax_record.send(value['from'].first)
         if data.is_a?(ActiveTriples::Relation)
           self.parsed_metadata[key] = data.map { |d| prepare_export_data(d) }.join('; ').to_s unless value[:excluded]
         else
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -76,7 +76,6 @@ module Bulkrax
     end
 
     def build_export_metadata
-      hyrax_record
       make_round_trippable
       self.parsed_metadata = {}
       self.parsed_metadata['id'] = hyrax_record.id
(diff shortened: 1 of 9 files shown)


bkiahstroud (commented):
@sephirothkod This is looking great! I have a few questions / requests:

app/views/bulkrax/exporters/_form.html.erb
```diff
@@ -85,5 +102,16 @@
...
+    // get the date filter option and show the corresponding date selectors
+    $('.exporter_date_filter').change(function() {
+        if($('.exporter_date_filter').find(".boolean").is(":checked"))
+            $('#date_filter_picker').removeClass('hidden');
+        else
+            $('#date_filter_picker').addClass('hidden');
+    });
+
+    if($('.exporter_date_filter').find(".boolean").is(":checked"))
+        $('#date_filter_picker').removeClass('hidden');
```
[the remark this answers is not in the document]
sephirothkod: This runs on page load. This deals with the checkbox being checked on update but date filter defaulting to hidden.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -103,35 +103,52 @@ def create_parent_child_relationships
...
+
+    def current_entry_ids
+      case importerexporter.export_from
+      when 'collection'
+        ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}"+extra_filters, rows: 2_000_000_000).map(&:id)
+      when 'worktype'
+        ActiveFedora::SolrService.query("has_model_ssim:#{importerexporter.export_source}"+extra_filters, rows: 2_000_000_000).map(&:id)
+      when 'importer'
+        importer = Bulkrax::Importer.find(importerexporter.export_source)
+        importer_ids = importer.entries.where(type: importer.parser.entry_class.to_s, last_error: [nil, {}, '']).map(&:identifier)
```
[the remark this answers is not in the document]
sephirothkod: wow, I should have done that from the get go! Rookie mistake, thanks for catching it.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -103,35 +103,52 @@ def create_parent_child_relationships
...
-      non_errored_entries.each_with_index do |entry, index|
+      current_entry_ids.each_with_index do |entry, index|
         break if limit_reached?(limit, index)
         query = "#{ActiveFedora.index_field_mapper.solr_name(Bulkrax.system_identifier_field)}:\"#{entry.identifier}\""
         work_id = ActiveFedora::SolrService.query(query, fl: 'id', rows: 1).first['id']
         new_entry = find_or_create_entry(entry_class, work_id, 'Bulkrax::Exporter')
         Bulkrax::ExportWorkJob.perform_now(new_entry.id, current_run.id)
       end
     end
+=end
```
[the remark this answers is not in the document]
sephirothkod: yes it can be deleted, I like to keep a commented version while I am working so that I don't forget the original goal of the method.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -103,35 +103,52 @@ def create_parent_child_relationships
...
+    def current_entry_ids
+      case importerexporter.export_from
+      when 'collection'
+        ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source}"+extra_filters, rows: 2_000_000_000).map(&:id)
+      when 'worktype'
+        ActiveFedora::SolrService.query("has_model_ssim:#{importerexporter.export_source}"+extra_filters, rows: 2_000_000_000).map(&:id)
+      when 'importer'
+        importer = Bulkrax::Importer.find(importerexporter.export_source)
+        importer_ids = importer.entries.where(type: importer.parser.entry_class.to_s, last_error: [nil, {}, '']).map(&:identifier)
+        ActiveFedora::SolrService.query("doc_id_tesim:(#{importer_ids.join(" OR ")})"+extra_filters, rows:2_000_000_000).map(&:id)
```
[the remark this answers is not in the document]
sephirothkod: working on resolving this already. I think because I changed the logic it is expecting a hash with just a single id from the solrservice, but now its getting all of them at once and returning an array of ids. updating the spec.

app/views/bulkrax/exporters/_form.html.erb
```diff
@@ -60,6 +60,23 @@
     hint: 'leave blank or 0 for all records',
     label: t('bulkrax.exporter.labels.limit') %>
 
+  <%= form.input :date_filter,
+                 as: :boolean,
+                 label: t('bulkrax.exporter.labels.filter_by__date') %>
```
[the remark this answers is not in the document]
sephirothkod: yup. lol. rookie mistake.

app/models/bulkrax/csv_entry.rb
```diff
@@ -76,25 +76,30 @@ def add_file
     end
 
     def build_export_metadata
+      hyrax_record
```
bkiahstroud, line 79: @sephirothkod Do we need to call `#hyrax_record` here? If so, why?

app/views/bulkrax/exporters/_form.html.erb
```diff
@@ -60,6 +60,23 @@
...
+                   as: :date,
+                   label: t('bulkrax.exporter.labels.start_date') %>
+
+    <%= form.input :finish_date,
+                   as: :date,
+                   label: t('bulkrax.exporter.labels.finish_date') %>
+  </div>
+  <%= form.input :work_visibility,
+                 collection: form.object.work_status_list,
+                 label: t('bulkrax.exporter.labels.status') %>
```
bkiahstroud, line 77: I think these labels should have i18n values in the [bulkrax.en.yml](https://github.com/samvera-labs/bulkrax/blob/export-from-date-range/config/locales/bulkrax.en.yml) file

bkiahstroud (approved):
One tiny request and one question. Otherwise, approved!

app/models/bulkrax/csv_entry.rb
sephirothkod: no we don't. I'll remove it.

app/views/bulkrax/exporters/_form.html.erb
sephirothkod: Do we need them? I'll add them, but I have never added them on projects in the past, unless I needed a label that said something different. I18n will auto-generate a human readable label based on the text after the last '.' which is all we usually need for english.

bkiahstroud: I've noticed that sometimes, if a translation is missing, a hover will be added to the element that says "missing translation:...". I think this depends on the element because it doesn't always happen. 

Either way, I feel like it would be a future-proof change as we do eventually want to [support all the same languages that Hyrax does](https://github.com/samvera-labs/bulkrax/issues/255).

bkiahstroud: Especially if we want to move out of samvera-labs into core samvera.

sephirothkod: fair enough, changes have been pushed.

PR #261, opened 2020-12-02
Fix file paths for Zip CSV imports 

bkiahstroud:
Only extract zip files once, instead of attempting to extract them for every Entry created

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -79,11 +79,15 @@ module Bulkrax
       self.parsed_metadata['model'] = hyrax_record.has_model.first
       build_mapping_metadata
 
+      # TODO: fix the "send" parameter in the conditional below
+      # currently it returns: "NoMethodError - undefined method 'bulkrax_identifier' for #<Collection:0x00007fbe6a3b4248>"
       if mapping['collection']&.[]('join')
-        self.parsed_metadata['collection'] = hyrax_record.member_of_collections.map { |c| c.send(work_identifier)&.first }.compact.uniq.join(';')
+        self.parsed_metadata['collection'] = hyrax_record.member_of_collection_ids.join('; ')
+        #   self.parsed_metadata['collection'] = hyrax_record.member_of_collections.map { |c| c.send(work_identifier)&.first }.compact.uniq.join(';')
       else
         hyrax_record.member_of_collections.each_with_index do |collection, i|
-          self.parsed_metadata["collection_#{i + 1}"] = collection.send(work_identifier)&.first
+          self.parsed_metadata["collection_#{i + 1}"] = collection.id
+          #     self.parsed_metadata["collection_#{i + 1}"] = collection.send(work_identifier)&.first
         end
       end
 
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -97,7 +97,11 @@ module Bulkrax
     end
 
     def current_run
-      @current_run ||= self.importer_runs.create!
+      @current_run ||= if file? && zip?
+                         self.importer_runs.create!
+                       else
+                         self.importer_runs.create!(total_work_entries: self.limit || parser.total, total_collection_entries: parser.collections_total)
+                       end
     end
 
     def last_run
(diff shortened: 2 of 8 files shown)


orangewolf:
@bkiahstroud can you look to see if this code is still relevant?

alishaevn:
## TLDR; importing a zip file only creates the files once so we're able to round trip csv's now. 🎉 

in testing round tripping for BL I discovered that importing a csv that had been exported with a type of "work type" or "collection" failed with `Zip::DestinationFileExistsError`. (see image below) I spoke with @orangewolf and @sephirothkod about it and we noticed that this pr was still sitting in draft state.

I reached out to @bkiahstroud to see if he remembered the direction he was headed in when he started this pr 9 months ago.

> I think the problem was when/where we were unzipping the file, like my commit message mentions. I think it tries to unzip it every time it creates an entry, and as soon as it hits the second entry it throws the error saying the file already exists because it was unzipped when making the first entry
I think this is also why the counters for zip imports are always in the thousands even if there’s only 100 rows
I think the idea I had was to unzip the zip file once, when the importer first starts, then just go look at it when creating the entries 

I pulled his branch and merged in the current version of `main`. I tested a combination of changes he had done and changes that were currently in `main` until I found the right combo. I'm now able to round trip the "work type" and "collection" export types! 🎉 

image

app/models/bulkrax/importer.rb
```diff
@@ -97,7 +97,7 @@ def schedulable?
     end
 
     def current_run
-      @current_run ||= self.importer_runs.create!(total_work_entries: self.limit || parser.total, total_collection_entries: parser.collections_total)
```
orangewolf, line 100: it seems like this will break total reporting for anything that is not a zip file.

spec/models/bulkrax/importer_spec.rb
```diff
@@ -21,14 +21,20 @@ module Bulkrax
     end
 
     describe 'importer run' do
-      before do
-        allow_any_instance_of(Bulkrax::OaiDcParser).to receive(:collections_total).and_return(1)
-      end
-
-      it 'creates an ImporterRun with total_work_entries set to the value of limit' do
+      it 'finds or creates a new ImporterRun' do
```
orangewolf, line 24: I think this test is likely wrong now (see comment above)

orangewolf (changes requested):

app/models/bulkrax/importer.rb
alishaevn: good point. I made this a conditional instead and tested each of the following successfully:
- import a csv of a work with no files
- import of a csv of a work with remote files
- import a zip of a csv with a "files" folder (local file)
- export by work type
- export by collection
- export by importer
- round trip of work type export
- round trip of collection export
- round trip of an importer export

spec/models/bulkrax/importer_spec.rb
alishaevn: fixed.

app/models/bulkrax/csv_entry.rb
```diff
@@ -79,13 +79,17 @@ def build_export_metadata
...
+        self.parsed_metadata['collection'] = hyrax_record.member_of_collection_ids.join('; ')
+        #   self.parsed_metadata['collection'] = hyrax_record.member_of_collections.map { |c| c.send(work_identifier)&.first }.compact.uniq.join(';')
       else
         hyrax_record.member_of_collections.each_with_index do |collection, i|
-          self.parsed_metadata["collection_#{i + 1}"] = collection.send(work_identifier)&.first
+          self.parsed_metadata["collection_#{i + 1}"] = collection.id
+          #     self.parsed_metadata["collection_#{i + 1}"] = collection.send(work_identifier)&.first
         end
-      end        
+      end
```
alishaevn, line 92: @sephirothkod and @orangewolf bringing your attention here. the code we merged into `main` earlier doesn't work. I'm using the singular "collection" version of the code that keeps with what braydon was working on, but I'm not using the "send" method because it gave me the error below.
this works for BL and since braydon is deploying OD from his branch, figured this is ok to go into `main` for the time being.
Screen Shot 2021-09-09 at 11 05 36 PM

orangewolf (approved):

PR #266, opened 2021-01-22
Add guards to new migrations

bkiahstroud:
Running migrations are throwing errors: 

```sh
PG::DuplicateColumn: ERROR:  column "start_date" of relation "bulkrax_exporters" already exists
: ALTER TABLE "bulkrax_exporters" ADD "start_date" date excluded from capture: DSN not set
```

Guard to prevent these errors.


jpengst:
👍

PR #273, opened 2021-04-22
Correctly calculate totals for zipped csv imports

bkeese:
In the case where a zip file is being imported, total was counting the lines in the zip file, not the contained .csv file.


orangewolf:
looks great! thank you

PR #276, opened 2021-04-28
make sure different versions of an importer with the same id do not cross-contaminate

orangewolf:
Especially important in multi-tenant, but also relevant if db is reset


labradford (approved):
Looks good

PR #285, opened 2021-07-08
File update

sephirothkod:
allows files in filesets to be updated rather than destroyed and reimported


app/controllers/bulkrax/importers_controller.rb
```diff
@@ -96,8 +97,8 @@ def create
     end
 
     # PATCH/PUT /importers/1
+    # rubocop:disable all
     def update
-      # rubocop:disable Style/IfInsideElse
```
orangewolf, line 100: giving up on rubocop here I see ;-) can we add a todo for refactoring later?

app/factories/bulkrax/object_factory.rb
```diff
@@ -218,7 +221,7 @@ def update_attributes
       # a way that is compatible with how the factory needs them.
       def transform_attributes
         attributes.slice(*permitted_attributes)
-                  .merge(file_attributes)
+                  .merge(file_attributes(@update_files))
```
orangewolf, line 224: i'd suggest using an accessor here. even if it is just `attr_accessor :update_files`, by avoiding using `@update_files` you gain a lot of flexibility by making everything a message (local variables and methods) and not using instance variables directly.

app/models/bulkrax/csv_entry.rb
```diff
@@ -22,7 +22,9 @@ def self.read_data(path)
     def self.data_for_entry(data)
       # If a multi-line CSV data is passed, grab the first row
       data = data.first if data.is_a?(CSV::Table)
+      raw_model = data[:model]
       raw_data = data.to_h
+      raw_data[:model] = raw_model
```
orangewolf, line 27: `raw_data_model_data[:model] = raw_data[:model].data` ;-) 

looks like `raw_model` never gets used anywhere else and this code feels confusing to me. `data` still exists unchanged after the to_h, so maybe 

```
raw_data[:model] = data[:model]
```

and a comment as to why?

app/models/bulkrax/csv_entry.rb
```diff
@@ -61,7 +62,7 @@ def build_metadata
       add_admin_set_id
       add_collections
       add_local
-
+      self.save
```
orangewolf, line 65: why is this saving here? save on the entry is called after the full build is done

orangewolf (changes requested):

app/models/bulkrax/csv_entry.rb
sephirothkod: This was a really weird case because it was a CSV element. when it uses to_h it sometimes doesn't translate the model correctly. So I pulled it out first then after it does to_h I overwrite the possibly incorrect model with the correct one. I'll add a comment explaining this.

app/factories/bulkrax/object_factory.rb
sephirothkod: I was already setting it as a attr_reader but then didn't use it! thanks for catching this rookie mistake!

app/models/bulkrax/csv_entry.rb
sephirothkod: I can't remember why I did this but I feel like it was something to do with testing for metadata updating correctly. I'll remove it and if we see any negative effects we can put it back

app/factories/bulkrax/object_factory.rb
orangewolf: not rookie, just the kinda thing that gets left

orangewolf (approved):
thanks for the updates

orangewolf:
@sephirothkod this looks good to me - can you take a quick peak at the resolved merge then hit the button if it looks good.

PR #287, opened 2021-07-13
importers and exporters links in hyrax 3.x

sephirothkod:
fixes issue where bulkrax link do not show in the sidebar in hyrax 3.x


orangewolf:
@sephirothkod this is great - I'm curious is it still hitting the partial incorrectly? we had added a comment to `app/views/hyrax/dashboard/sidebar/_repository_content.html.erb` at one point, and removed it at another, but I see its still there.

sephirothkod:
@orangewolf That wasn't happening but when checking that I found another issue, the code has been safely moved to the initializer and I am getting no more NameErrors.

PR #288, opened 2021-07-16
objects-in-csv-revision

alishaevn:
# Summary
currently the object metadata is coming back as only an object. when I compared that against the data that's returned from the british library ui when multiple creators are added, it's not a simple object though.

#### current implementation per the spec
```
{
  "source"=>["2"],
  "title"=>["some title"],
  "creator"=>{
    "creator_first_name"=>"Fake",
    "creator_last_name"=>"Fakerson",
    "creator_position"=>"Leader, Jester, Queen",
    "creator_language"=>"English"
  },
  "file"=>[],
  "visibility"=>"open",
  "rights_statement"=>[nil],
  "admin_set_id"=>"MyString"
}
```

#### intended implementation per the UI
"creator" is an object (or series of objects), wrapped in an array, wrapped in a string, wrapped in another array
```
{
  "source"=>["2"],
  "title"=>["some title"],
  "creator"=> ["[
    {
      "creator_given_name":"Creator given name 1",
      "creator_family_name":"Creator family name 1",
      "creator_name_type":"Personal",
      "creator_orcid":"reator1",
      "creator_isni":"reator1",
      "creator_position":"1",
      "creator_institutional_relationship":["Staff member"]
    }
  ]"],
  "file"=>[],
  "visibility"=>"open",
  "rights_statement"=>[nil],
  "admin_set_id"=>"MyString"
}
```

### DIFF
--- a/app/models/concerns/bulkrax/has_matchers.rb
+++ b/app/models/concerns/bulkrax/has_matchers.rb
@@ -46,6 +46,13 @@ module Bulkrax
           multiple_metadata?(name, node_name, node_content, object)
         else
           Rails.logger.info("Bulkrax Column automatically matched #{node_name}, #{node_content}")
+          # if object
+          #   puts "node_content >> #{node_content}"
+          #   puts "object >> #{object}"
+          #   puts "parsed_metadata >> #{parsed_metadata}"
+          #   puts "name >> #{name}"
+            # debugger
+          # end
 
           node_content = node_content.content if node_content.is_a?(Nokogiri::XML::NodeSet)
           next parsed_metadata[object][name] = Array.wrap(node_content.to_s.strip).join('; ') if object && node_content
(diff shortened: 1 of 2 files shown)


orangewolf:
@alishaevn I would have expected these to be 
```
 "creator"=> [
    {
      "creator_given_name":"Creator given name 1",
      "creator_family_name":"Creator family name 1",
      "creator_name_type":"Personal",
      "creator_orcid":"reator1",
      "creator_isni":"reator1",
      "creator_position":"1",
      "creator_institutional_relationship":["Staff member"]
    }
]
```

(aka the array, but not the string / array). the reason fo this is that creator is multiple as covered in https://github.com/samvera-labs/bulkrax/pull/282 

we may have to do extra to handle the difference between the array of hashes and the extra wrapping we're seeing from the UI

alishaevn:
closing this ticket. having a single object returned for a single object in the csv and an array of objects returned for multiple objects in the csv is the expected bulkrax behavior we're creating.

if a project needs some other implementation, it can be handled on a custom basis.

PR #289, opened 2021-07-16
multiple-objects-in-csv

alishaevn:
co-authors: @alishaevn @orangewolf 

# Ticket
https://github.com/samvera-labs/bulkrax/issues/296

# Summary
the parser needs to be able to handle situations where a field is an array objects, e.g.:
```
"creator"=> ["[
  {
    "creator_given_name":"Creator given name 1",
    "creator_family_name":"Creator family name 1",
    "creator_name_type":"Personal",
    "creator_orcid":"reator1",
    "creator_isni":"reator1",
    "creator_position":"0",
    "creator_institutional_relationship":["Staff member"]
  }, {
    "creator_given_name":"Creator given name 2",
    "creator_family_name":"Creator family name 2",
    "creator_name_type":"Personal",
    "creator_orcid":"reator2",
    "creator_isni":"reator2",
    "creator_position":"1",
    "creator_institutional_relationship":["Research associate"]
  }, {
    "creator_organization_name":"Creator organisation name 3",
    "creator_name_type":"Organisational",
    "creator_isni":"reator3",
    "creator_ror":"www.Creator-organisation-ROR-3.com",
    "creator_grid":"www.Creator-organisation-GRID-3.com",
    "creator_wikidata":"www.Creator-organisation-Wikidata-3.com",
    "creator_position":"2"
  }
]"]
```

# Expected behavior
- bulkrax can handle an array of objects with a prefixed key
- bulkrax can handle an array of objects without a prefixed key
- [configuration steps](https://github.com/samvera-labs/bulkrax/wiki/Configuration#mapping-from-multiple-objects) have been added to the wiki

# Demo
image

# Resources
- https://github.com/samvera/hyrax/blob/b034218b89dde7df534e32d1e5ade9161e129a1d/app/models/concerns/hyrax/basic_metadata.rb
- https://dublincore.org/specifications/dublin-core/dcmi-terms/


spec/models/bulkrax/csv_entry_spec.rb
```diff
@@ -90,29 +90,63 @@ module Bulkrax
...
+            'single_object_language' => 'english'
+          )
+        end
+
+        it 'succeeds' do
+          metadata = subject.build_metadata
+          expect(metadata['single_object']['first_name']).to eq('Fake')
+          expect(metadata['single_object']['last_name']).to eq('Fakerson')
+          expect(metadata['single_object']['position']).to include('Leader', 'Jester', 'Queen')
+          expect(metadata['single_object']['language']).to eq(['English'])
```
alishaevn, line 136: @orangewolf although we set `single_object` to be `multiple:false`, this check now equates to an array of strings because the `language` property [here](https://github.com/samvera/hyrax/blob/b034218b89dde7df534e32d1e5ade9161e129a1d/app/models/concerns/hyrax/basic_metadata.rb) is defaulted to multiple.

`expect(metadata['single_object']['single_object_language']).to eq('English')` in the spec above still comes back as a string, since `single_object_language` is not a predefined property.


do we want to keep it this way? I [made an adjustment](https://github.com/samvera-labs/bulkrax/pull/289/commits/6aace7ef4f067c9d39820730042acb7d03f9bf49) that would alter `multiple?` to return false if the field OR the object were set to false. but I backtracked it because I started to doubt that direction.

alishaevn: or would you prefer `expect(metadata['single_object']['single_object_language'])` be adjusted to also equal `['English']` since that goes along with the spirit of how `language` is supposed to react?

app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -77,6 +84,10 @@ def object_name(field)
     # @return [Array] hyrax fields
     def field_to(field)
       fields = mapping&.map do |key, value|
+        if value['from'].instance_of?(Array)
+          value['from'] = value['from'].join(', ')
+        end
```
alishaevn, line 89: without this change, `object` and `object_multiple` would both come back false. this happened because the field names had numbers at the end so the `include?` below would return false.

spec/models/bulkrax/csv_entry_spec.rb
alishaevn: this was resolved with the removal of `include ::Hyrax::BasicMetadata` from the "Work" class.

spec/models/bulkrax/csv_entry_spec.rb
```diff
@@ -121,21 +121,97 @@ module Bulkrax
...
       end
+
+      context 'with multiple objects and fields prefixed' do
+        let(:importer) do
+          FactoryBot.create(:bulkrax_importer_csv, field_mapping: {
+                              'multiple_objects_first_name' => { from: ['multiple_objects_first_name_1', 'multiple_objects_first_name_2'], object: 'multiple_objects' },
+                              'multiple_objects_last_name' => { from: ['multiple_objects_last_name_1', 'multiple_objects_last_name_2'], object: 'multiple_objects' },
+                              'multiple_objects_position' => { from: ['multiple_objects_position_1', 'multiple_objects_position_2'], object: 'multiple_objects' },
+                              'multiple_objects_language' => { from: ['multiple_objects_language_1'], object: 'multiple_objects', parsed: true }
+                            })
```
bkiahstroud, line 147: Because we use this same "multiple objects" importer in our spec more than once, can we make it its own sub-factory or factory trait?

spec/test_app/app/models/work.rb
```diff
@@ -3,5 +3,6 @@
 # Generated by hyrax:models:install
 class Work < ActiveFedora::Base
   include ::Hyrax::WorkBehavior
-  property :creator, predicate: ::RDF::Vocab::DC.creator
+  property :single_object, predicate: ::RDF::Vocab::DC.creator, multiple: false
+  property :multiple_objects, predicate: ::RDF::Vocab::DC.creator
```
bkiahstroud, line 7: Is `::RDF::Vocab::DC.creator` still correct with this change? I'm not sure how much it matters since this is the test work

app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 
+# TODO(alishaevn): see if these rules can be adhered to, instead of disabled
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/ParameterLists
+# rubocop:disable Metrics/CyclomaticComplexity
```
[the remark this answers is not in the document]
alishaevn: discussed this with kiah. the todo will be resolved later.

app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -103,12 +113,21 @@ def multiple_metadata?(name, node_name, node_content, object = false)
...
-        if multiple
+        if object_multiple
+          index = parsed_metadata[object].find_index { |obj| obj[name].nil? }
+
+          if index.nil?
+            parsed_metadata[object] << {}
+            parsed_metadata[object][parsed_metadata[object].length - 1][name] = Array.wrap(result).join('; ')
+          else
+            parsed_metadata[object][index][name] = Array.wrap(result).join('; ')
+          end
```
bkiahstroud, line 129: Please add a comment / comments explaining what these lines are doing and why we're doing it the way we are. Personally, I would not know what this was doing just by looking at the code

alishaevn:
```suggestion
        if object_multiple
          # find the index of the first object in the `object` array where the `name` key doesn't already exist
          index = parsed_metadata[object].find_index { |obj| obj[name].nil? }

          if index.nil?
            # if all existing objects have our `name` key in it already
            # push a new object to the end of the array and add the `result` to it
            parsed_metadata[object] << {}
            parsed_metadata[object][parsed_metadata[object].length - 1][name] = Array.wrap(result).join('; ')
          else
            # if an object already exists that doesn't have our `name` key in it, add the `result` to that object
            parsed_metadata[object][index][name] = Array.wrap(result).join('; ')
          end
```

bkiahstroud (approved):

Issue #297, opened 2021-07-26
Simplify source identifying


alishaevn:
# Summary
the `system_identifier_field` and `source_identifier_field_mapping` are two different config settings that are causing confusion

# Acceptance Criteria
- [x] add `source_identifier` as an additional field mapping property to determine the value per entry type
- [x] have bulkrax set the source identifier per the chosen field mapping

orangewolf:
#302 and #304 help close this out

PR #298, opened 2021-07-26
simplify-source-identifying

alishaevn:
# Ticket
https://github.com/samvera-labs/bulkrax/issues/297

# Summary
the `system_identifier_field` and `source_identifier_field_mapping` are two different config settings that are causing confusion. 

# Expected behavior
- the `source_identifier_key` method extracts the source_identifier per entry type
- the `field_to` method properly handles array and string types


app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -86,9 +86,13 @@ def get_object_name(field)
     # @return [Array] hyrax fields
     def field_to(field)
       fields = mapping&.map do |key, value|
-        value['from'] = value['from'].join(', ') if value['from'].instance_of?(Array)
+        return unless value
 
-        key if (value.present? && value['from']&.include?(field)) || key == field
+        if value['from'].instance_of?(Array)
```
alishaevn, line 91: @orangewolf using a guard clause above instead of `if value&.[]('from').instance_of?(Array)` on line 91 because rubocop didn't like that. offense: "Do not chain ordinary method call after safe navigation operator."

 plus it reads a bit cleaner this way.

orangewolf (approved):

Issue #296, opened 2021-07-26
Object, multiple objects and nested object field mapping


alishaevn:
# Related to
https://gitlab.com/notch8/britishlibrary/-/issues/58

# Summary
bulkrax field mapping needs to be updated to handle more complexity

# Acceptance criteria
Bulkrax can:
- [x] handle an object
- [x] handle multiple objects
- [x] handle an object with a property that has multiple values
- [x] handle csv headings that are prepended or appended with numerators

orangewolf:
this was completed

PR #300, opened 2021-07-28
296 object with nested values

alishaevn:
# Ticket
https://github.com/samvera-labs/bulkrax/issues/296

# Expected behavior
- getting the index of an item based on the column header so mappings are always correct
- object keys that have an array as the value succeed

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -49,7 +49,7 @@ module Bulkrax
     end
 
     def key_without_numbers(key)
-      key.sub(/_\d+/, '').sub(/^\d+_/, '')
+      key.sub(/_\d+/, '').sub(/_\d+/, '').sub(/^\d+_/, '')
     end
 
     def build_metadata
@@ -61,7 +61,8 @@ module Bulkrax
       self.parsed_metadata[Bulkrax.system_identifier_field] = [record['source_identifier']]
       record.each do |key, value|
         next if key == 'collection'
-        add_metadata(key_without_numbers(key), value)
+        index = key[/\d+/].to_i - 1 if key[/\d+/].to_i != 0
+        add_metadata(index, key_without_numbers(key), value)
       end
 
       add_file
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -61,8 +61,9 @@ module Bulkrax
       self.parsed_metadata[Bulkrax.system_identifier_field] = [record['source_identifier']]
       record.each do |key, value|
         next if key == 'collection'
+
         index = key[/\d+/].to_i - 1 if key[/\d+/].to_i != 0
-        add_metadata(index, key_without_numbers(key), value)
+        add_metadata(key_without_numbers(key), value, index)
       end
 
       add_file
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -49,7 +49,7 @@ module Bulkrax
     end
 
     def key_without_numbers(key)
-      key.sub(/_\d+/, '').sub(/_\d+/, '').sub(/^\d+_/, '')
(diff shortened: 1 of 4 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -49,7 +49,7 @@ def keys_without_numbers(keys)
     end
 
     def key_without_numbers(key)
-      key.sub(/_\d+/, '').sub(/^\d+_/, '')
+      key.sub(/_\d+/, '').sub(/_\d+/, '').sub(/^\d+_/, '')
```
alishaevn, line 52: the double check is needed to handle `alpha_1_2`.
doing `key.sub(/_\d+_\d+/, '').sub(/^\d+_/, '')` also worked for the above example, but failed for `alpha_1`.

spec/models/bulkrax/csv_entry_spec.rb
```diff
@@ -203,15 +203,49 @@ module Bulkrax
...
         end
-        # rubocop:enable RSpec/ExampleLength
       end
+
+      context 'with object fields prefixed and properties with multiple values' do
+        let(:importer) do
+          FactoryBot.create(:bulkrax_importer_csv, field_mapping: {
+                              'multiple_objects_first_name' => { from: ['multiple_objects_first_name'], object: 'multiple_objects' },
+                              'multiple_objects_last_name' => { from: ['multiple_objects_last_name'], object: 'multiple_objects' },
+                              'multiple_objects_position' => { from: ['multiple_objects_position'], object: 'multiple_objects', nested_type: 'Array' }
```
alishaevn, line 220: setting `nested_type: 'Array'` instead of something like `array: true` so it allows for handling other nested types (e.g. object) in the future. if that ever becomes a need.

app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -100,42 +101,38 @@ def excluded?(field)
...
-    def matched_metadata?(matcher, multiple, name, node_content, object, object_multiple)
+    def matched_metadata?(matcher, multiple, name, index, node_content, object_name, object_multiple)
       result = matcher.result(self, node_content)
       return unless result
 
-      if object
+      if object_name
         if object_multiple
-          # find the index of the first object in the `object` array where the `name` key doesn't already exist
-          index = parsed_metadata[object].find_index { |obj| obj[name].nil? }
```
alishaevn, line 123: setting the index this way no longer worked with the nested value because it would just create new objects instead of adding the value to existing key

app/models/bulkrax/csv_entry.rb
```diff
@@ -61,7 +61,9 @@ def build_metadata
       self.parsed_metadata[Bulkrax.system_identifier_field] = [record['source_identifier']]
       record.each do |key, value|
         next if key == 'collection'
-        add_metadata(key_without_numbers(key), value)
+
+        index = key[/\d+/].to_i - 1 if key[/\d+/].to_i != 0
```
alishaevn, line 65: using the csv column header to determine the index instead. the conditional is needed so that keys without numbers don't return a -1.
```
key = 'alpha_1'
index = key[/\d+/].to_i - 1 if key[/\d+/].to_i != 0        => 1

key = 'alpha'
index = key[/\d+/].to_i - 1 if key[/\d+/].to_i != 0        => nil
```

app/models/bulkrax/csv_entry.rb
sephirothkod: can you use gsub instead?

alishaevn: ahhhhh. yeah, that works! thanks!

sephirothkod (approved):

orangewolf (commented):
@alishaevn this is really good. thank you!

alishaevn:
details about this code has been added to https://github.com/samvera-labs/bulkrax/wiki/Configuration#mapping-from-objects

PR #302, opened 2021-07-30
source id refactor and proper error if missing

sephirothkod:
allows setting source_identifier field in mapping hash. provides detailed error if import is not valid (missing source_id). refactors source_identifier to used mapped value throughout.


sephirothkod:
@orangewolf if you have time to go over this with me and make sure everything is the way you anticipated, that would be greatly appreciated.

alishaevn:
@sephirothkod everything looks like it's good to me. although without updated specs, I'm not completely certain.

- can you merge in master again? there are a few changes from https://github.com/samvera-labs/bulkrax/pull/298 that I believe will conflict with what you have here
- why the commented out code? is that just until you can talk with rob?

sephirothkod:
@alishaevn I didn't realize that you had done some source_id refactoring recently as well. I'll have a look. I set this as a draft just so Rob could verify that this is what he wanted, and then I'll push updated specs and remove comments, etc.

app/factories/bulkrax/object_factory.rb
```diff
@@ -7,16 +7,16 @@ class ObjectFactory
     include Bulkrax::FileFactory
     define_model_callbacks :save, :create
     class_attribute :system_identifier_field
-    attr_reader :attributes, :object, :unique_identifier, :klass, :replace_files, :update_files
+    attr_reader :attributes, :object, :source_identifier_value, :klass, :replace_files, :update_files
     self.system_identifier_field = Bulkrax.system_identifier_field
```
orangewolf, line 11: i expected to see a change to this. the idea is that you specify a field in the field mapping and the from on that is the source_identifier and to two on it is the system_identifier

app/models/bulkrax/csv_entry.rb
```diff
@@ -32,8 +32,8 @@ def self.data_for_entry(data)
       return raw_data
     end
 
-    def self.source_identifier_field
-      Bulkrax.source_identifier_field_mapping[self.to_s] || 'source_identifier'
+    def source_identifier
+      importerexporter&.parser&.source_identifier.to_s || 'source_identifier'
```
orangewolf, line 36: entry has parser directly, I think you may be taking the long way home

app/parsers/bulkrax/application_parser.rb
```diff
@@ -40,6 +40,12 @@ def records(_opts = {})
       raise StandardError, 'must be defined'
     end
 
+    def source_identifier
```
orangewolf, line 43: seems good and yes, it is what I meant

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -33,16 +33,29 @@ def import_fields
...
-      !required_elements.map { |el| keys.map(&:to_s).include?(el) }.include?(false)
+      return unless keys.present?
+      !missing_elements(keys).present?
+    end
+
+    def missing_elements(keys)
+      required_elements - keys.map(&:to_s)
+    end
+
+    def source_identifier
```
orangewolf, line 44: if this is in application parser does it need to be here too?

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -33,16 +33,29 @@ def import_fields
...
+      @source_identifier ||=  importerexporter.mapping.select do |_, h|
+        h.key?("source_identifier")
+      end.values.first&.[]("from")&.first&.to_sym || :source_identifier
     end
 
     def required_elements
-      %w[title source_identifier]
+      %w[title] + [source_identifier.to_s]
```
orangewolf, line 51: this will conflict with Alisha's changes. I think there is come confusion about who was doing what. I thought you were going to focus just on the source_idenfier vs system_identifier bit and Alisha would tackle fixing csv parser.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -33,16 +33,29 @@ def import_fields
...
     end
 
     def valid_import?
-      required_elements?(import_fields) && file_paths.is_a?(Array)
+      error_alert = "Missing at least one required element, missing element(s) are: #{missing_elements(import_fields).join(', ')}"
```
orangewolf, line 55: ah! that's why that stopped happening =-)

orangewolf (changes requested):
seems like a good start. left some comments. I'd also expect this work to contain:
- [ ] removal/replacement of Bulkrax.system_identifier with the "to" on the source_id specified field. 
- [x] validation on there being only 1 source_id in the field map
- [x] removal of any code that uses "id" (this may already be done?)
- [ ] update generator initialize showing how to set the source_id on a field
- [ ] readme section on why bulkrax requires a uniq id field on input
- [ ] spec fixes

edit(sephirothkod): changed to checkbox list

app/factories/bulkrax/object_factory.rb
sephirothkod: I totally understand the 'from' being the source_id, and the 'to' being the system_id. we don't have access to the parser/importer/entry here so it must be passed in to the factory, there isn't really anything I can do to change that...

app/models/bulkrax/csv_entry.rb
sephirothkod: yup! fixed and moved to base entry class

app/parsers/bulkrax/csv_parser.rb
sephirothkod: nope, removed

app/factories/bulkrax/object_factory.rb
orangewolf: @sephirothkod passing system_identifier into the constructor from the entry seems like a solution. Then we aren’t declaring it at the class level here any more.  maybe rename it work identifier? Or something else not so similar to source_identifier?

app/parsers/bulkrax/csv_parser.rb
sephirothkod: using Alisha's version of required elements

sephirothkod:
All that is needed is testing and specs. working on that this afternoon.

app/factories/bulkrax/object_factory.rb
sephirothkod: named it work_identifier throughout

orangewolf (approved):

PR #306, opened 2021-08-06
Remove last_error, up database text field size

### LINKED ISSUE #303: Database text fields need larger limit
The text type database fields are not specified with the limit option and so have the default limit of 65k. This results in errors, particularly when error backtraces are written to the `last_error` fields bulkrax uses, or when an import uses a large metadata file.

Suggestion is to use `limit: 16777215` for text fields in migrations to indicated a MEDIUMTEXT field is needed in the db schema.

orangewolf:
- We stopped writing last error in order to use the statuses table. There were still some places in the code that used it (and would never pass). @sephirothkod one of these was the "export the errors" for csv parsers.
- We use serialized text fields for storing error backtraces and field sets. These values were potentially too small for some use cases.

Closes #303

### DIFF
--- a/app/models/bulkrax/entry.rb
+++ b/app/models/bulkrax/entry.rb
@@ -14,12 +14,10 @@ module Bulkrax
     belongs_to :importerexporter, polymorphic: true
     alias importer importerexporter
     alias exporter importerexporter
-    has_many :statuses, as: :statusable, dependent: :destroy
 
     serialize :parsed_metadata, JSON
     # Do not serialize raw_metadata as so we can support xml or other formats
     serialize :collection_ids, Array
-    serialize :last_error, JSON
 
     paginates_per 5
 
--- a/app/models/concerns/bulkrax/errored_entries.rb
+++ b/app/models/concerns/bulkrax/errored_entries.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module Bulkrax
+  module ErroredEntries
+    extend ActiveSupport::Concern
+
+    def write_errored_entries_file
+      if @errored_entries.blank?
+        entry_ids = importerexporter.entries.pluck(:id)
+        error_statuses = Bulkrax::Status.latest_by_statusable
+                                        .includes(:statusable)
+                                        .where('bulkrax_statuses.statusable_id IN (?) AND bulkrax_statuses.statusable_type = ? AND status_message = ?', entry_ids, 'Bulkrax::Entry', 'Failed')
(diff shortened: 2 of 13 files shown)


app/models/bulkrax/entry.rb
```diff
@@ -14,12 +14,10 @@ class Entry < ApplicationRecord
     belongs_to :importerexporter, polymorphic: true
     alias importer importerexporter
     alias exporter importerexporter
-    has_many :statuses, as: :statusable, dependent: :destroy
```
orangewolf, line 17: this just moved to the include so we didn't have to write it 3 times.

app/models/concerns/bulkrax/errored_entries.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
```
orangewolf, line 2: moved this stuff to a module so Rubocop would leave me alone about the CsvParser class length

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -128,10 +129,13 @@ def current_work_ids
       when 'worktype'
         ActiveFedora::SolrService.query("has_model_ssim:#{importerexporter.export_source + extra_filters}", rows: 2_000_000_000).map(&:id)
       when 'importer'
-        importer = Bulkrax::Importer.find(importerexporter.export_source)
-        # TODO: this seems likely to be broken
-        entry_ids = importer.entries.where(type: importer.parser.entry_class.to_s, last_error: [nil, {}, '']).map(&:identifier)
-        ActiveFedora::SolrService.query("#{work_identifier}_tesim:(#{entry_ids.join(' OR ')})#{extra_filters}", rows: 2_000_000_000).map(&:id)
+        entry_ids = importer.entries.pluck(:id)
```
orangewolf, line 132: this is 3 queries (one for initial entries, one for statuses and a last one of the entries of those statuses) how ever it is only every 3 queries and not an n+1. I'd rather do it in 1 (entries where that last status is "Complete" and the importer is this importer), but I couldn't figure out how to do that.

spec/test_app/db/schema.rb
```diff
@@ -29,11 +29,10 @@
     t.string "collection_ids"
     t.string "type"
     t.integer "importerexporter_id"
-    t.text "raw_metadata"
-    t.text "parsed_metadata"
+    t.text "raw_metadata", limit: 16777215
```
orangewolf, line 32: @dlpierce is this what you had in mind?

alishaevn (approved):

PR #330, opened 2021-08-28
fixes files updating when only metadata should be updated.

sephirothkod:


app/factories/bulkrax/object_factory.rb
```diff
@@ -186,40 +186,49 @@ def collection_type(attrs)
...
       attributes.slice(*permitted_attributes)
                 .merge(file_attributes(update_files))
+      else
+        attributes.slice(*permitted_attributes)
+      end
+    end
+
+    def transform_create_attributes
+        attributes.slice(*permitted_attributes)
+            .merge(file_attributes(update_files))
```
bkiahstroud, line 231: @sephirothkod there's some odd spacing in these two methods, but otherwise looks good to me 👍

bkiahstroud (approved):

app/factories/bulkrax/object_factory.rb
sephirothkod: ya lint failed so I'm fixing that now.

orangewolf (changes requested):
I think this can and should be done more simply.  give me a moment

PR #331, opened 2021-08-28
do not pass files if the update type is metadata only

orangewolf:


sephirothkod (approved):

sephirothkod:
@orangewolf this is tested and confirmed working

Issue #321, opened 2021-08-23
Update exporting to handle a single object, multiple objects and nested objects


alishaevn:
# Related to
- https://gitlab.com/notch8/britishlibrary/-/issues/148
- https://github.com/samvera-labs/bulkrax/issues/296

# Summary
bulkrax exporting needs to be updated to handle more complexity

# Acceptance criteria
Bulkrax export can:
- [x] handle an object
- [x] handle multiple objects
- [x] handle an object with a property that has multiple values
- [x] handle csv headings that are prepended or appended with numerators

PR #334, opened 2021-09-02
321 export all headers

alishaevn:
Ref https://github.com/samvera-labs/bulkrax/issues/321

# Expected behavior
- numerated headers are present on an exported csv
- object headers are grouped by the property name (e.g. creator_name_1, creator_name_2, creator_age_1) and not the object itself (e.g. creator_name_1, creator_age_1, creator_name_2). **this may be contrary to how headers are organized on the import but the ordering does not appear to affect csv's being imported or exported successfully.**

# Demo
example of exported csv: [export.csv](https://github.com/samvera-labs/bulkrax/files/7108868/export.csv)

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -47,12 +47,12 @@ module Bulkrax
       self.parsed_metadata = {}
       self.parsed_metadata[work_identifier] = [record[source_identifier]]
       record.each do |key, value|
-        next if key == 'collection'
+        next if %w[collection collections].include?(key_without_numbers(key))
 
         index = key[/\d+/].to_i - 1 if key[/\d+/].to_i != 0
         add_metadata(key_without_numbers(key), value, index)
       end
-      add_file
+      add_file unless importerexporter.metadata_only?
       add_visibility
       add_rights_statement
       add_admin_set_id
@@ -78,13 +78,15 @@ module Bulkrax
       self.parsed_metadata[source_identifier] = hyrax_record.send(work_identifier)
       self.parsed_metadata['model'] = hyrax_record.has_model.first
       build_mapping_metadata
-      if mapping['collections']&.[]('join')
-        self.parsed_metadata['collections'] = hyrax_record.member_of_collection_ids.join('; ')
+
+      if mapping['collection']&.[]('join')
+        self.parsed_metadata['collection'] = hyrax_record.member_of_collections.map { |c| c.send(work_identifier)&.first }.compact.uniq.join(';')
       else
-        hyrax_record.member_of_collection_ids.each_with_index do |collection_id, i|
-          self.parsed_metadata["collections_#{i + 1}"] = collection_id
+        hyrax_record.member_of_collections.each_with_index do |collection, i|
+          self.parsed_metadata["collection_#{i + 1}"] = collection.send(work_identifier)&.first
         end
-      end
+      end        
+
       build_files unless hyrax_record.is_a?(Collection)
       self.parsed_metadata
     end
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -103,7 +103,7 @@ module Bulkrax
 
         data = object_key.present? ? hyrax_record.send(value['object']) : hyrax_record.send(key.to_s)
         if object_key.present?
-          next self.parsed_metadata[key] = '' if data.empty?
+          next if data.empty?
           data = data.first if data.is_a?(ActiveTriples::Relation)
 
           object_metadata(data, object_key)
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -14,9 +14,9 @@ module Bulkrax
     def self.read_data(path)
       raise StandardError, 'CSV path empty' if path.blank?
       CSV.read(path,
-               headers: true,
-               header_converters: :symbol,
-               encoding: 'utf-8')
+        headers: true,
+        header_converters: :symbol,
+        encoding: 'utf-8')
     end
 
     def self.data_for_entry(data, _source_id)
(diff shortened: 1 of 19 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -209,11 +209,31 @@ def key_allowed(key)
 
     # All possible column names
     def export_headers
+      # need to find the numerated headers, which may be different per entry
+      # and then remove the duplicates
+      parsed_metadata_keys = []
+      importerexporter.entries.where(identifier: current_work_ids)[0..limit || total].each do |entry|
```
alishaevn, line 215: couldn't think of another way to do this. unsure if this will be super slow when exporting 1000 works or something though.

app/models/bulkrax/csv_entry.rb
```diff
@@ -103,7 +103,7 @@ def build_mapping_metadata
 
         data = object_key.present? ? hyrax_record.send(value['object']) : hyrax_record.send(key.to_s)
         if object_key.present?
-          next self.parsed_metadata[key] = '' if data.empty?
+          next if data.empty?
```
alishaevn, line 106: the code previously was adding numerated headers if the array had more than 1 object, but non numerated headers if the array had 1 object.
e.g. we would get `contributor_family_name`, `contributor_family_name_1`, `contributor_family_name_2`

now we get `contributor_family_name_1`, `contributor_family_name_2`

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -136,7 +132,12 @@ def create_new_entries
       current_work_ids.each_with_index do |wid, index|
         break if limit_reached?(limit, index)
         new_entry = find_or_create_entry(entry_class, wid, 'Bulkrax::Exporter')
-        Bulkrax::ExportWorkJob.perform_now(new_entry.id, current_run.id)
+        entry = Bulkrax::ExportWorkJob.perform_now(new_entry.id, current_run.id)
+
+        if entry
+          self.headers ||= []
```
alishaevn, line 138: without setting `headers` as an array first, line 139 would only return `true` instead of the actual headers.

orangewolf: that appears to be because you didn't port the initialization line over from the other code =-)

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -136,7 +132,12 @@ def create_new_entries
...
+          self.headers |= entry.parsed_metadata.keys
```
orangewolf, line 139: I think this code is wrong - the headers should be the "from"s not the "to"s

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -209,11 +210,30 @@ def key_allowed(key)
...
-      headers << source_identifier.to_s
-      headers << 'model'
-      headers << 'collections'
-      importerexporter.mapping.each_key { |key| headers << key if key_allowed(key) }
+      headers = []
+      # the object keys are not part of a work types default metadata but the object itself is,
+      # so we check according to it and add the acceptable headers
+      importerexporter.mapping.each do |key, value|
+        if value.key?('object') && key_allowed(value['object'])
+          object_keys = self.headers.each { |sh| sh.to_s.include?(value['object']) }
```
orangewolf, line 218: this code will (unless it throws an exception) always do the same thing as `object_keys = self.headers`.  The loop here appears to be just for fun.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -209,11 +210,30 @@ def key_allowed(key)
...
+          next object_keys.each { |item| headers << item }
```
orangewolf, line 219: why do you need `next` here? doesn't it do object.keys.each {|item| headers << item} all the time?

app/parsers/bulkrax/csv_parser.rb
orangewolf: turns out this is right, its the headers that csv entry send out that were wrong.

app/parsers/bulkrax/csv_parser.rb
orangewolf: `each` returns every item in the iterator so `[1, 2, 3, 4].each {|i| i if i.odd?}` returns `[1,2,3,4]`.  `[1,2,3,4].detect {|i| i.odd?}` would return `[1,3]` `[1,2,3,4].map {|i| i if i.odd?}` would return `[1,nil,3,nil]`.  I think its the detect that you wanted, but then that code got changed a lot in #339

app/parsers/bulkrax/csv_parser.rb
alishaevn: just double checked https://github.com/samvera-labs/bulkrax/pull/337/files again. what part did I miss?

app/parsers/bulkrax/csv_parser.rb
alishaevn: gotcha. I started off with `each` and then changed the method to `select` on friday for the same reason you stated. but something started messing up with the headers. I don't remember what. so I changed it back.

app/parsers/bulkrax/csv_parser.rb
alishaevn: my intention was so that the code on line 222 didn't get run. `key_allowed(key)` would return false anyway for an object field mapping so the header wouldn't be added, but I figured no point having the code run.

unless the `next` has no bearing on line 222 since it's out of the loop?

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -44,8 +40,9 @@ def missing_elements(keys)
     end
 
     def valid_import?
-      error_alert = "Missing at least one required element, missing element(s) are: #{missing_elements(import_fields).join(', ')}"
-      raise StandardError, error_alert unless required_elements?(import_fields)
+      import_strings = keys_without_numbers(import_fields.map(&:to_s))
```
alishaevn, line 43: 👍🏾

PR #337, opened 2021-09-03
accumulate headers when creating entries

orangewolf:


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -136,7 +132,8 @@ def create_new_entries
       current_work_ids.each_with_index do |wid, index|
         break if limit_reached?(limit, index)
         new_entry = find_or_create_entry(entry_class, wid, 'Bulkrax::Exporter')
-        Bulkrax::ExportWorkJob.perform_now(new_entry.id, current_run.id)
+        entry = Bulkrax::ExportWorkJob.perform_now(new_entry.id, current_run.id)
+        headers |= entry.parsed_metadata.keys
```
orangewolf, line 136: this should mean that `self.headers` has the full, uniq set of headers once all the entries are made.

alishaevn:
used this code in https://github.com/samvera-labs/bulkrax/pull/334

PR #339, opened 2021-09-06
Refactor csv export to use from values instead of to values.

orangewolf:
CSV export was using the wrong headers. If you specify `"title" => {from: ['display_title']}` in the mapper,
then you would expect the export to say `display_title` and not `title`. This was not the case.

This commit also adds a new option for the field mapping - join. The default behavior on export has been to
join fields together, but field_1,field_2 easier for users to edit. This will be the new default and join allows
you to specify the old behavior.

Also cleaned up object code on exporter and reduced cyclomatic complexity on some methods in csv entry.

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -108,7 +108,7 @@ module Bulkrax
 
     def build_object(object_key, value)
       data = hyrax_record.send(value['object'])
-      next if data.empty?
+      return if data.empty?
 
       data = data.to_a if data.is_a?(ActiveTriples::Relation)
       object_metadata(Array.wrap(data), object_key)
@@ -148,7 +148,11 @@ module Bulkrax
 
     def object_metadata(data, object_key)
       data = data.map { |d| eval(d) }.flatten # rubocop:disable Security/Eval
+
       data.each_with_index do |obj, index|
+        # allow the object_key to be valid whether it's a string or symbol
+        obj = obj.with_indifferent_access
+
         next unless obj[object_key]
         if obj[object_key].is_a?(Array)
           obj[object_key].each_with_index do |_nested_item, nested_index|
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -14,9 +14,9 @@ module Bulkrax
     def self.read_data(path)
       raise StandardError, 'CSV path empty' if path.blank?
       CSV.read(path,
-               headers: true,
-               header_converters: :symbol,
-               encoding: 'utf-8')
+        headers: true,
+        header_converters: :symbol,
+        encoding: 'utf-8')
     end
 
     def self.data_for_entry(data, _source_id)
@@ -38,14 +38,6 @@ module Bulkrax
 
     def self.children_field
       Bulkrax.parent_child_field_mapping[self.to_s] || 'children'
-    end
-
-    def keys_without_numbers(keys)
-      keys.map { |key| key_without_numbers(key) }
-    end
-
-    def key_without_numbers(key)
-      key.gsub(/_\d+/, '').sub(/^\d+_/, '')
     end
 
     def build_metadata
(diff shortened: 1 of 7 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -40,14 +40,6 @@ def self.children_field
       Bulkrax.parent_child_field_mapping[self.to_s] || 'children'
     end
 
-    def keys_without_numbers(keys)
```
orangewolf, line 43: keys_without_numbers moved to somewhere that both the parser and entry can access it

app/models/bulkrax/csv_entry.rb
```diff
@@ -124,29 +147,27 @@ def prepare_export_data(datum)
     end
 
     def object_metadata(data, object_key)
-      data = convert_to_hash(data)
-
+      data = data.map { |d| eval(d) }.flatten # rubocop:disable Security/Eval
```
orangewolf, line 150: I would love a better solution here. I think that converting to json and back is just as dangerous but also more complex.

alishaevn: why is it just as dangerous? I'm not familiar with `eval` outside of when @sephirothkod and I were pairing on the code in `convert_to_hash` and he mentioned that `eval` is unsafe.

app/parsers/bulkrax/application_parser.rb
```diff
@@ -25,6 +25,7 @@ def self.import_supported?
 
     def initialize(importerexporter)
       @importerexporter = importerexporter
+      @headers = []
```
alishaevn, line 28: 👍🏾

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -202,25 +199,14 @@ def write_files
       end
     end
 
-    def key_allowed(key)
-      !Bulkrax.reserved_properties.include?(key) &&
-        new_entry(entry_class, 'Bulkrax::Exporter').field_supported?(key) &&
+    def export_key_allowed(key)
+      new_entry(entry_class, 'Bulkrax::Exporter').field_supported?(key) &&
         key != source_identifier.to_s
     end
```
alishaevn, line 205:
```suggestion
    def export_key_allowed(key)
      new_entry(entry_class, 'Bulkrax::Exporter').field_supported?(key) &&
      key != source_identifier.to_s
    end
```

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -202,25 +199,14 @@ def write_files
       end
     end
 
-    def key_allowed(key)
-      !Bulkrax.reserved_properties.include?(key) &&
-        new_entry(entry_class, 'Bulkrax::Exporter').field_supported?(key) &&
+    def export_key_allowed(key)
+      new_entry(entry_class, 'Bulkrax::Exporter').field_supported?(key) &&
         key != source_identifier.to_s
```
alishaevn, line 204: is it a rails thing to have subsequent lines in a conditional be indented beyond the first line?

spec/models/bulkrax/csv_entry_spec.rb
```diff
@@ -364,10 +364,10 @@ module Bulkrax
 
         it 'succeeds' do
           metadata = subject.build_export_metadata
-          expect(metadata['first_name_1']).to eq('Fake')
-          expect(metadata['last_name_1']).to eq('Fakerson')
-          expect(metadata['position_1']).to include('Leader', 'Jester', 'Queen')
-          expect(metadata['language_1']).to eq('english')
+          expect(metadata['single_object_first_name_1']).to eq('Fake')
```
alishaevn, line 367: since the import and export features are now both using the "from" property, is there still a need to have 2 different tests based on the "to" field changing?

app/models/bulkrax/csv_entry.rb
sephirothkod: I would also like an answer here, I have always been told that `eval` exists to never be used. ie. it is such a security risk that you should never use it other than for debugging.

app/parsers/bulkrax/csv_parser.rb
sephirothkod: yes this is proper spacing in rails

PR #340, opened 2021-09-07
include object headers again

alishaevn:
# Related to
https://github.com/samvera-labs/bulkrax/pull/334
https://github.com/samvera-labs/bulkrax/pull/339

# Summary
- the object headers were not being exported in https://github.com/samvera-labs/bulkrax/pull/339

# Expected behavior
- object headers are exported again

# Example export
[export.csv](https://github.com/samvera-labs/bulkrax/files/7123162/export.csv)

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -108,6 +108,7 @@ module Bulkrax
 
     def build_object(object_key, value)
       data = hyrax_record.send(value['object'])
+      # byebug
       return if data.empty?
 
       data = data.to_a if data.is_a?(ActiveTriples::Relation)
@@ -147,17 +148,19 @@ module Bulkrax
     end
 
     def object_metadata(data, object_key)
-      # eval turns all of the keys into symbols
       data = data.map { |d| eval(d) }.flatten # rubocop:disable Security/Eval
+
       data.each_with_index do |obj, index|
-        object_symbol = object_key.to_sym
-        next unless obj[object_symbol]
-        if obj[object_symbol].is_a?(Array)
-          obj[object_symbol].each_with_index do |_nested_item, nested_index|
-            self.parsed_metadata["#{key_for_export(object_key)}_#{index + 1}_#{nested_index + 1}"] = prepare_export_data(obj[object_symbol][nested_index])
+        # allow the object_key to be valid whether it's a string or symbol
+        obj = obj.with_indifferent_access
+
+        next unless obj[object_key]
+        if obj[object_key].is_a?(Array)
+          obj[object_key].each_with_index do |_nested_item, nested_index|
+            self.parsed_metadata["#{key_for_export(object_key)}_#{index + 1}_#{nested_index + 1}"] = prepare_export_data(obj[object_key][nested_index])
           end
         else
-          self.parsed_metadata["#{key_for_export(object_key)}_#{index + 1}"] = prepare_export_data(obj[object_symbol])
+          self.parsed_metadata["#{key_for_export(object_key)}_#{index + 1}"] = prepare_export_data(obj[object_key])
         end
       end
     end
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -108,10 +108,11 @@ module Bulkrax
 
     def build_object(object_key, value)
       data = hyrax_record.send(value['object'])
-      next if data.empty?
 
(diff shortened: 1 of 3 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -108,10 +108,11 @@ def build_mapping_metadata
 
     def build_object(object_key, value)
       data = hyrax_record.send(value['object'])
-      next if data.empty?
```
alishaevn, line 111: this returns "syntaxerror: invalid next" because you can't call `next` outside a loop

app/models/bulkrax/csv_entry.rb
```diff
@@ -147,15 +148,17 @@ def prepare_export_data(datum)
...
 
     def object_metadata(data, object_key)
+      # eval turns all of the keys into symbols
       data = data.map { |d| eval(d) }.flatten # rubocop:disable Security/Eval
       data.each_with_index do |obj, index|
-        next unless obj[object_key]
-        if obj[object_key].is_a?(Array)
-          obj[object_key].each_with_index do |_nested_item, nested_index|
-            self.parsed_metadata["#{key_for_export(object_key)}_#{index + 1}_#{nested_index + 1}"] = prepare_export_data(obj[object_key][nested_index])
+        object_symbol = object_key.to_sym
```
alishaevn, line 154: initially I reassigned `object_key` to be a symbol. but that caused an issue when being passed into the `key_for_export` method. so instead of reconverting `object_key` into a string inside of `key_for_export`, I created the new variable `object_symbol`.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -211,16 +211,13 @@ def export_headers
       # we don't want access_control_id exported and we want file at the end
       # also sort the headers so they're grouped and easier to find
       headers.delete('access_control_id') if headers.include?('access_control_id')
-      headers.delete('file') if headers.include?('file')
```
alishaevn, line 214: no longer adding `file` and `collections` here because they are now numerated and in `self.headers`. I imported a csv I exported last week (in a zip along with the files) and it imported correctly too.

app/models/bulkrax/csv_entry.rb
orangewolf: yes, I refactored at the very end to make rubocop happy. this should just be `return if data.empty?`

alishaevn: yup! that's what I changed it to. 🙂

app/models/bulkrax/csv_entry.rb
orangewolf: I would have thought these needed to be strings all the time. are you sure they need to be symbols?

app/parsers/bulkrax/csv_parser.rb
orangewolf: that makes sense to me.

app/models/bulkrax/csv_entry.rb
alishaevn: in my testing I discovered that the eval version of `data` returned
```
[{:creator_family_name=>"Kornicki", :creator_given_name=>"Peter Francis", :creator_isni=>"110780723", :creator_name_type=>"Personal"}]
```

therefore, `obj[object_key]` was always blank because it was looking for a key with a data type of string, and not symbol.

alishaevn: although....the specs _are_ failing for the opposite reason. they are expecting strings and not symbols. 🤔

app/parsers/bulkrax/csv_parser.rb
orangewolf: this makes sense to me

app/models/bulkrax/csv_entry.rb
alishaevn: this was resolved by calling `with_in_different_access` on the obj hash inside of the `each`.

spec/parsers/bulkrax/csv_parser_spec.rb
```diff
@@ -259,14 +259,12 @@ module Bulkrax
         headers = parser.export_headers
         expect(headers).to include('id')
         expect(headers).to include('model')
-        expect(headers).to include('collections')
```
alishaevn, line 262: this spec doesn't have mappings for `collection` or `file` so the numerated versions aren't in the headers either.

PR #358, opened 2021-09-21
custom-model-header-name-on-export

alishaevn:
# Expected behavior
- the header that refers to the Model on the csv will be "model" unless there is a field mapping for "model"
- if there's a field mapping for "model", the header will use that field mapping

# Demo
field mapping: `'model' => { from: ['work_type'] }`

csv:
[export 3.csv](https://github.com/samvera-labs/bulkrax/files/7200251/export.3.csv)
image

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -76,7 +76,7 @@ module Bulkrax
       self.parsed_metadata = {}
       self.parsed_metadata['id'] = hyrax_record.id
       self.parsed_metadata[source_identifier] = hyrax_record.send(work_identifier)
-      self.parsed_metadata['model'] = hyrax_record.has_model.first
+      self.parsed_metadata[mapping['model']['from'].first || 'model'] = hyrax_record.has_model.first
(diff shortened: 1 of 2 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -76,7 +76,7 @@ def build_export_metadata
       self.parsed_metadata = {}
       self.parsed_metadata['id'] = hyrax_record.id
       self.parsed_metadata[source_identifier] = hyrax_record.send(work_identifier)
-      self.parsed_metadata['model'] = hyrax_record.has_model.first
+      self.parsed_metadata[mapping['model'] & ['from']&.first || 'model'] = hyrax_record.has_model.first
```
alishaevn, line 79: without the spaces around the ampersand I got this rubocop error: "Surrounding space missing for operator &" 🤷🏾‍♀️

alishaevn:
after explaining the difference /similarities between model and work_type and how model will relate to future work, sara and jenny decided they're fine with keeping the exported header as "model".

PR #362, opened 2021-09-25
seperate-paths-by-tenant

alishaevn:
# Related issue
https://gitlab.com/notch8/britishlibrary/-/issues/148#note_686505543

# Expected behavior
- import and export paths will now have the tenant name for hyku apps
- imports a csv and a zip successfully
- exports a csv successfully
- previously imported csv's/zip still have a valid import path, but the csv/zip should be re-imported for consistency and to avoid future issues
- previously exported csv's don't have a valid download anymore so they need to be re-exported

# Demo
| csv import | zip import | export|
| --- | --- | --- |
| image | image |  image |

### DIFF
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -153,11 +153,15 @@ module Bulkrax
 
     # If the import data is zipped, unzip it to this path
     def importer_unzip_path
-      @importer_unzip_path ||= File.join(Bulkrax.import_path, "import_#{path_string}")
+      @importer_unzip_path ||= File.join(import_base_path, "import_#{path_string}")
     end
 
     def errored_entries_csv_path
-      @errored_entries_csv_path ||= File.join(Bulkrax.import_path, "import_#{path_string}_errored_entries.csv")
+      @errored_entries_csv_path ||= File.join(import_base_path, "import_#{path_string}_errored_entries.csv")
+    end
+
+    def import_base_path
+      ENV['HYKU_MULTITENANT'] ? File.join(Bulkrax.import_path, Site.instance.account.name) : Bulkrax.import_path
     end
 
     def path_string
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -153,15 +153,11 @@ module Bulkrax
 
     # If the import data is zipped, unzip it to this path
     def importer_unzip_path
-      @importer_unzip_path ||= File.join(import_base_path, "import_#{path_string}")
+      @importer_unzip_path ||= File.join(Bulkrax.base_path.call('import'), "import_#{path_string}")
     end
 
     def errored_entries_csv_path
-      @errored_entries_csv_path ||= File.join(import_base_path, "import_#{path_string}_errored_entries.csv")
-    end
-
-    def import_base_path
-      ENV['HYKU_MULTITENANT'] ? File.join(Bulkrax.import_path, Site.instance.account.name) : Bulkrax.import_path
+      @errored_entries_csv_path ||= File.join(Bulkrax.base_path.call('import'), "import_#{path_string}_errored_entries.csv")
     end
 
     def path_string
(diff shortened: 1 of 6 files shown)


spec/parsers/bulkrax/csv_parser_spec.rb
```diff
@@ -95,29 +95,44 @@ module Bulkrax
       let(:importer) { FactoryBot.create(:bulkrax_importer_csv_failed) }
       let(:file)     { fixture_file_upload('./spec/fixtures/csv/ok.csv') }
 
-      it 'returns the path of the partial import file' do
-        expect(subject.write_partial_import_file(file))
-          .to eq("tmp/imports/#{importer.id}_#{importer.created_at.strftime('%Y%m%d%H%M%S')}/failed_corrected_entries.csv")
-      end
+      context 'in a single tenant application' do
```
alishaevn, line 98: same specs as before, they're just wrapped in a context block.

spec/parsers/bulkrax/csv_parser_spec.rb
```diff
@@ -95,29 +95,44 @@ module Bulkrax
...
+          expect(partial_import_filename).not_to eq(uploaded_filename)
+          expect(partial_import_filename).to eq('failed_corrected_entries.csv')
+        end
       end
 
-      it 'renames the uploaded file to the original import filename + _corrected_entries' do
-        import_filename = importer.parser_fields['import_file_path'].split('/').last
-        uploaded_filename = file.original_filename
-        partial_import_filename = subject.write_partial_import_file(file).split('/').last
+      context 'in a multi tenant application' do
```
alishaevn, line 125: this wip spec is to verify that the tenant string is being added.

lib/bulkrax.rb
```diff
@@ -33,6 +34,10 @@ class << self
     self.removed_image_path = Bulkrax::Engine.root.join('spec', 'fixtures', 'removed.png').to_s
     self.server_name = 'bulkrax@example.com'
 
+    self.base_path = lambda do |type|
+      ENV['HYKU_MULTITENANT'] ? File.join(Bulkrax.send("#{type}_path"), Site.instance.account.name) : Bulkrax.send("#{type}_path")
+    end
+
```
sephirothkod, line 40: I don't think this should be here as a lambda, I think it would be better suited to existing as a regular method in the application_parser. Also that would be more readable.

app/parsers/bulkrax/application_parser.rb
```diff
@@ -95,7 +95,7 @@ def write_import_file(file)
     # Path where we'll store the import metadata and files
     #  this is used for uploaded and cloud files
     def path_for_import
-      @path_for_import = File.join(Bulkrax.import_path, importerexporter.path_string)
+      @path_for_import = File.join(Bulkrax.base_path.call('import'), importerexporter.path_string)
       FileUtils.mkdir_p(@path_for_import) unless File.exist?(@path_for_import)
       @path_for_import
     end
```
sephirothkod, line 101: add:

```
def base_path(type = 'import')
      ENV['HYKU_MULTITENANT'] ? File.join(Bulkrax.send("#{type}_path"), Site.instance.account.name) : Bulkrax.send("#{type}_path")
    end

File: app/models/bulkrax/importer.rb [hunk 5]
```diff
@@ -153,11 +153,11 @@ def import_metadata_format
 
     # If the import data is zipped, unzip it to this path
     def importer_unzip_path
-      @importer_unzip_path ||= File.join(Bulkrax.import_path, "import_#{path_string}")
+      @importer_unzip_path ||= File.join(Bulkrax.base_path.call('import'), "import_#{path_string}")
```
sephirothkod, line 156: becomes:

```
@importer_unzip_path ||= File.join(parser.base_path('import'), "import_#{path_string}")

sephirothkod (changes requested):

lib/bulkrax.rb
alishaevn: @sephirothkod I tried to put a method in application parser first, but it didn't work there. I'll try again tomorrow though.

app/models/bulkrax/importer.rb
alishaevn: ahhh. this is probably what I was missing in my first attempt. I was calling the method directly instead of from the parser. 
thanks!

sephirothkod (approved):

PR #364, opened 2021-09-30
Add ability to import Collections with metadata

bkiahstroud:
# Summary

Add ability to import collections with metadata when `:model` or model mappings are set to Collection **(CSV only)** 

| model | source_identifier | title | description |
| --- | --- | --- | --- |
| Collection | collection_1 | First Collection | Lorem ipsum |

Retains ability to import Collections by title when importing works 

| model | source_identifier | title | collection |
| --- | --- | --- | --- |
| Work | work_1 | First Work | Parent Collection |

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -159,11 +159,10 @@ module Bulkrax
 
     def member_of_collections
       ms = object.member_of_collection_ids.to_a.map { |id| find_collection(id) }
-      [:collection, :collections].each do |atat|
-        next if attributes[atat].blank?
+      if attributes[parser.collection_field_mapping].present?
         ms.concat(
           Array.wrap(
-            find_collection(attributes[atat])
+            find_collection(attributes[parser.collection_field_mapping])
           )
         )
       end
@@ -194,7 +193,7 @@ module Bulkrax
     def create_attributes
       return transform_attributes if klass == Collection
       ActiveSupport::Deprecation.warn("Passing collection or collections directly to the ObjectFactory is no longer supported. Please update your Entry class to call add_collections instead.") if attributes[:collection].present? || attributes[:collections].present?
-      transform_attributes.except(:collections, :collection)
+      transform_attributes.except(:collections, :collection, parser.collection_field_mapping)
     end
 
     # Strip out the :collection key, and add the member_of_collection_ids,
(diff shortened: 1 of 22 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -8,14 +8,26 @@ def self.export_supported?
...
+      importer.parser_fields['total'] = csv_data.count
+      importer.save
+      @records ||= csv_data.map { |record_data| entry_class.data_for_entry(record_data, nil) }
+    end
+
     def collections
-      # does the CSV contain a collection column?
-      return [] unless (import_fields & [:collection, :collections]).any?
+      col_mapping = Bulkrax.collection_field_mapping[self.entry_class.to_s]&.to_sym || :collection
+      model_mappings = Bulkrax.field_mappings[self.class.to_s]&.dig('model', :from) || []
```
alishaevn, line 22: the `self.class.to_s` call in csv_entry.rb evaluations to "Class". will this not do the same?

app/parsers/bulkrax/oai_dc_parser.rb
```diff
@@ -119,6 +119,11 @@ def collections_total
       end
     end
 
+    # TODO: change to differentiate between collection and work records when adding ability to import collection metadata
```
alishaevn, line 122: isn't this TODO outdated?

app/parsers/bulkrax/csv_parser.rb
bkiahstroud: It shouldn't. This is an instance method, so `self` in this case is an instance of `CsvParser`: 
```ruby
self
=> #<Bulkrax::CsvParser:0x00005602f005e2b0 ...
self.class
=> Bulkrax::CsvParser
```

I think the method in the `CsvEntry` as a class method, so `self` in that case would be the class: 
```ruby
self
=> Bulkrax::CsvEntry
self.class
=> Class
```

alishaevn (approved):

app/factories/bulkrax/object_factory.rb
```diff
@@ -213,6 +212,10 @@ def transform_attributes
       @transform_attributes
     end
 
+    def parser
+      Entry.find_by(identifier: source_identifier_value)&.parser || ApplicationParser
```
sephirothkod, line 216: can we lazy load this so we aren't making multiple calls?

app/models/bulkrax/csv_collection_entry.rb
```diff
@@ -7,13 +7,27 @@ def factory_class
     end
 
     def build_metadata
-      self.parsed_metadata = self.raw_metadata
+      self.parsed_metadata = {}
+      self.parsed_metadata[work_identifier] = self.identifier
```
orangewolf, line 11: where does identifier come from?

bkiahstroud: [CsvParser#unique_collection_identifier](https://github.com/samvera-labs/bulkrax/pull/364/files#diff-ccbdc46eb1fafd2a6aec08c257c2589531c46fee5d793832a85bf0653f754490R275-R284). I added a comment to the top of the new method that handles this explaining why we're using `self.identifier`

orangewolf (approved):

PR #367, opened 2021-10-08
Adds a 'Delete' status and upgrades the Pending symbol

sephirothkod:


app/views/bulkrax/importers/show.html.erb
```diff
@@ -101,6 +101,8 @@
                   <td><%= e.id %></td>
                   <% if e.status == "Complete" %>
                     <td><span class="glyphicon glyphicon-ok" style="color: green;"></span> <%= e.status %></td>
+                  <% elsif e.status == "Pending" %>
```
orangewolf, line 104: should "Deleted" also be added here somewhere?

spec/jobs/bulkrax/delete_work_job_spec.rb
```diff
@@ -34,6 +35,7 @@ module Bulkrax
       it 'increments :deleted_records' do
         expect(importer_run).to receive(:increment!).with(:deleted_records)
         expect(importer_run).to receive(:decrement!).with(:enqueued_records)
+        entry.save
```
orangewolf, line 38: ? seems like maybe there aren't any specs for this?

app/views/bulkrax/importers/show.html.erb
sephirothkod: I left deleted as the `else` category since that is the red `x` I thought it matched well. I can also add a seperate symbol for deleted though if you think that would be better.

Screen Shot 2021-10-08 at 15 31 48

spec/jobs/bulkrax/delete_work_job_spec.rb
sephirothkod: not really. I can hash out the specs for the delete job a bit more next week if you want.

app/views/bulkrax/importers/show.html.erb
orangewolf: I think having it be a green x is probably right or the same as complete.  Red X means failed here, so the delete succeeding should be different I think

PR #373, opened 2021-10-18
returning-model-factory-class-again

alishaevn:
# Issue
I kept getting "undefined method `method_defined?` for nil:NilClass" when importing a csv. I tracked the issue down to `def factory_class` in import_behvior.rb. 

# Expected behavior
- the correct value is returned for a collection or model factory class

# Demo of broken code
Screen Shot 2021-10-18 at 4 24 08 PM

### DIFF
--- a/app/models/concerns/bulkrax/import_behavior.rb
+++ b/app/models/concerns/bulkrax/import_behavior.rb
@@ -83,9 +83,12 @@ module Bulkrax
            else
              Bulkrax.default_work_type
            end
+
+      # return the name of the collection
       fc.downcase.camelcase.constantize
-    rescue NameError
-      nil
+    rescue NameError => e
+      # return the name of the model, or nil
+      fc.constantize || nil
     rescue
       Bulkrax.default_work_type.constantize
     end
--- a/app/models/concerns/bulkrax/import_behavior.rb
+++ b/app/models/concerns/bulkrax/import_behavior.rb
@@ -84,11 +84,12 @@ module Bulkrax
              Bulkrax.default_work_type
            end
 
-      # return the name of the collection
-      fc.downcase.camelcase.constantize
+      # return the name of the collection or work
+      fc.gsub!(' ', '_')
+      fc.downcase! if fc =~ /[-_]/
+      fc.camelcase.constantize
     rescue NameError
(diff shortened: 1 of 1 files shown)


app/models/concerns/bulkrax/import_behavior.rb
```diff
@@ -83,9 +83,12 @@ def factory_class
            else
              Bulkrax.default_work_type
            end
+
+      # return the name of the collection
       fc.downcase.camelcase.constantize
```
alishaevn, line 88: this line was updated from "fc.constantize" to "fc.downcase.camelcase.constantize" for the collection work. the camelcase method doesn't actually camelcase the work though, it titlecases it. (refer to screenshot in pr summary)

alishaevn: `fc.downcase.camelcase.constantize || fc.constantize` still throws the no method error

bkiahstroud: @alishaevn do you still get the no method error if you do this? 

```ruby
fc&.downcase&.camelcase&.constantize || fc.constantize
```

alishaevn: didn't try that one. I can in a bit though

bkiahstroud: Actually, I don't think my suggestion will work. 

I made the change to `fc.downcase.camelcase.constantize` in order to handle when the model isn't uppercased and/or is snakecase: 

```
work
generic_work
```

But it seems to be tripping up on classes that are already camelcase or titlecased: 

```
BookContributor
bookContributor
```

I have an idea, give me a few minutes to test it

bkiahstroud:
```ruby
fc.gsub!(' ', '_')
fc.downcase! if fc =~ /[-_]/
fc.camelcase.constantize
```

I tested this for the following values:

```
genericWork
GenericWork
generic work
Generic work
Generic Work
generic_work
Generic_Work
```

I think something like this would be preferable to the current diff because if `fc.constantize` fails, I think we would get stuck in an infinite loop (infinitely catching and throwing `NameError`s)

PR #380, opened 2021-11-22
protect exporter from halting if one record is not correct

orangewolf:


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -158,8 +158,11 @@ def create_new_entries
       current_work_ids.each_with_index do |wid, index|
         break if limit_reached?(limit, index)
         new_entry = find_or_create_entry(entry_class, wid, 'Bulkrax::Exporter')
-        entry = Bulkrax::ExportWorkJob.perform_now(new_entry.id, current_run.id)
-
+        begin
+          entry = Bulkrax::ExportWorkJob.perform_now(new_entry.id, current_run.id)
+        rescue => e
+          Rails.logger.info("#{e.message} was detected during export")
+        end
```
alishaevn, line 165: 👍🏾 retroactively approving

PR #384, opened 2021-12-07
Use parsed relationship mappings in entry parsed_metadata

bkiahstroud:
This commit was supposed to go in with #383 🤦


alishaevn (commented):
retroactively approved!

PR #387, opened 2021-12-14
Fix TypeError when parsing metadata

bkiahstroud:
# Summary

**Changes include:**
- Fix TypeError when parsing metadata
- Do not add blank values to parsed metadata
- Update specs


app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -52,11 +52,18 @@ def add_metadata(node_name, node_content, index = nil)
                   single_metadata(node_content)
                 end
 
-        set_parsed_data(object_multiple, object_name, name, index, value)
+        object_name.present? ? set_parsed_object_data(object_multiple, object_name, name, index, value) : set_parsed_data(name, value)
       end
     end
 
-    def set_parsed_data(object_multiple, object_name, name, index, value)
+    def set_parsed_data(name, value)
```
alishaevn, line 59: is this separation just so the methods are shorter?

alishaevn (approved):

app/models/concerns/bulkrax/has_matchers.rb
bkiahstroud: That, and all the extra arguments (`object_name`, `index`, etc) aren't necessary for non-object parsed metadata

PR #388, opened 2021-12-20
Add ability to import FileSet metadata using the CSV Importer

bkiahstroud:
# Summary

All changes only apply to CSV imports.

Changes include: 

- Adds the concept of a file set entry (similar to collection entry) 
- Extends `ObjectFactory` to handle importing file sets 
- Extracts dynamic record searching from `CreateRelationshipsJob` to new `DynamicRecordLookup` concern 
- `ImporterRun` counters: 
  - Count each record type separately (`processed_works`, `failed_collections`, etc)
  - All `*_records` counters are now generic (counts works + collections + file sets) 
- Adds new `ImportFileSetJob` to handle creation of file sets 
- Add file set entry information to importer index and show pages 
- Test coverage 
- [New documentation](https://github.com/samvera-labs/bulkrax/wiki/CSV-Importer#method-2) in Wiki

PR #403, opened 2022-01-28
i401 i402 fix bugs with collection_field_mapping and parent-child jobs

### LINKED ISSUE #401: collection_field_mapping support v.2.0
Until support for collection_field_mapping is deprecated, we need to support this configuration. 

With the current changes, the importer will fail because the collection doesn't have a work identifier. 

In the older version of bulkrax, there were [default settings](https://github.com/samvera-labs/bulkrax/blob/v1.0.0/app/parsers/bulkrax/csv_parser.rb#L56-L69) when creating a collection.

Bulkrax v2.0 [removes](https://github.com/samvera-labs/bulkrax/blob/fbfe3960e444babfde2e84f017ab98775e34525b/app/parsers/bulkrax/csv_parser.rb#L73-L86) the default settings. Therefore when the entry checks for missing required keys, an exception is raised.

### LINKED ISSUE #402: v2.0 empty parent/child identifiers
v2.0 

Do not call [CreateRelationshipsJob](https://github.com/samvera-labs/bulkrax/blob/main/app/models/concerns/bulkrax/import_behavior.rb#L27-L37) if the parent or child identifiers is an empty string

Doing so causes an [exception](https://github.com/samvera-labs/bulkrax/blob/main/app/jobs/bulkrax/create_relationships_job.rb#L44) to be raised.

ShanaLMoore:
Resolves #401 
Resolves #402 

- [x] Add required metadata to collections being imported using the `collection_field_mapping` so they do not throw errors  
- [x] Don't throw an error when trying to create a parent-child relationships with a blank identifier 
- [x] Don't try to create parent-child relationships when running importer in validate only mode 

Verified that I was able to created collections and associated them to works using the old and new ways.

### DIFF
--- a/app/models/bulkrax/csv_collection_entry.rb
+++ b/app/models/bulkrax/csv_collection_entry.rb
@@ -9,7 +9,7 @@ module Bulkrax
     # Use identifier set by CsvParser#unique_collection_identifier, which falls back
     # on the Collection's first title if record[source_identifier] is not present
     def add_identifier
-      self.parsed_metadata[work_identifier] = [self.identifier]
+      self.parsed_metadata[work_identifier] = [self.identifier].flatten
     end
 
     def add_collection_type_gid
--- a/app/models/bulkrax/csv_collection_entry.rb
+++ b/app/models/bulkrax/csv_collection_entry.rb
@@ -9,7 +9,7 @@ module Bulkrax
     # Use identifier set by CsvParser#unique_collection_identifier, which falls back
     # on the Collection's first title if record[source_identifier] is not present
     def add_identifier
-      self.parsed_metadata[work_identifier] = self.identifier
+      self.parsed_metadata[work_identifier] = [self.identifier]
     end
 
     def add_collection_type_gid
--- a/app/models/concerns/bulkrax/import_behavior.rb
+++ b/app/models/concerns/bulkrax/import_behavior.rb
@@ -11,9 +11,9 @@ module Bulkrax
         unless self.importerexporter.validate_only
           raise CollectionsCreatedError unless collections_created?
           @item = factory.run!
+          parent_jobs if self.parsed_metadata[related_parents_parsed_mapping].present?
+          child_jobs if self.parsed_metadata[related_children_parsed_mapping].present?
         end
-        parent_jobs if self.parsed_metadata[related_parents_parsed_mapping].present?
-        child_jobs if self.parsed_metadata[related_children_parsed_mapping].present?
       rescue RSolr::Error::Http, CollectionsCreatedError => e
         raise e
       rescue StandardError => e
@@ -26,13 +26,17 @@ module Bulkrax
 
     def parent_jobs
       self.parsed_metadata[related_parents_parsed_mapping].each do |parent_identifier|
-        CreateRelationshipsJob.perform_later(entry_identifier: self.identifier, parent_identifier: parent_identifier, importer_run: self.last_run) if parent_identifier.present?
+        next if parent_identifier.blank?
+
+        CreateRelationshipsJob.perform_later(entry_identifier: self.identifier, parent_identifier: parent_identifier, importer_run: self.last_run)
       end
     end
 
     def child_jobs
       self.parsed_metadata[related_children_parsed_mapping].each do |child_identifier|
-        CreateRelationshipsJob.perform_later(entry_identifier: self.identifier, child_identifier: child_identifier, importer_run: self.last_run) if child_identifier.present?
+        next if child_identifier.blank?
+
+        CreateRelationshipsJob.perform_later(entry_identifier: self.identifier, child_identifier: child_identifier, importer_run: self.last_run)
       end
     end
 
--- a/app/models/concerns/bulkrax/import_behavior.rb
+++ b/app/models/concerns/bulkrax/import_behavior.rb
@@ -26,13 +26,13 @@ module Bulkrax
 
     def parent_jobs
       self.parsed_metadata[related_parents_parsed_mapping].each do |parent_identifier|
-        CreateRelationshipsJob.perform_later(entry_identifier: self.identifier, parent_identifier: parent_identifier, importer_run: self.last_run)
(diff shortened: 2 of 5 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -74,8 +74,22 @@ def create_collections
       collections.each_with_index do |collection, index|
         next if collection.blank?
         break if records.find_index(collection).present? && limit_reached?(limit, records.find_index(collection))
+        ActiveSupport::Deprecation.warn(
+          'Creating Collections using the collection_field_mapping will no longer be supported as of Bulkrax version 3.0.' \
+          ' Please configure Bulkrax to use related_parents_field_mapping and related_children_field_mapping instead.'
+        )
+
+        metadata = if collection.key?(:title) && collection.keys.count == 1
```
bkiahstroud, line 82: @ShanaLMoore can we add a comment above this explaining why it's here and that it should be removed when `collection_field_mapping` is removed please?

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -74,8 +74,22 @@ def create_collections
...
+          {
+            title: [collection[:title]],
+            source_identifier: [collection[:title]],
+            visibility: 'open',
+            collection_type_gid: Hyrax::CollectionType.find_or_create_default_collection_type.gid
+          }
+        end
 
-        new_entry = find_or_create_entry(collection_entry_class, unique_collection_identifier(collection), 'Bulkrax::Importer', collection.to_h)
+        collection_hash = metadata.present? ? metadata : collection
```
alishaevn, line 91:
```suggestion
        collection_hash = metadata.presence || collection
```

app/models/concerns/bulkrax/import_behavior.rb
```diff
@@ -26,13 +26,13 @@ def build_for_importer
 
     def parent_jobs
       self.parsed_metadata[related_parents_parsed_mapping].each do |parent_identifier|
-        CreateRelationshipsJob.perform_later(entry_identifier: self.identifier, parent_identifier: parent_identifier, importer_run: self.last_run)
+        CreateRelationshipsJob.perform_later(entry_identifier: self.identifier, parent_identifier: parent_identifier, importer_run: self.last_run) if parent_identifier.present?
```
bkiahstroud, line 29: Would you mind also including [this change](https://github.com/samvera-labs/bulkrax/pull/388/files#diff-b28d7bc5e71422d28ccdcbda7f6f294700f0f12553cc2820e0cf02cca25cb35cR14-L16) in this PR? I should have put it in a smaller bug fix PR like this and it's related

app/models/bulkrax/csv_collection_entry.rb
```diff
@@ -9,7 +9,7 @@ def factory_class
     # Use identifier set by CsvParser#unique_collection_identifier, which falls back
     # on the Collection's first title if record[source_identifier] is not present
     def add_identifier
-      self.parsed_metadata[work_identifier] = self.identifier
+      self.parsed_metadata[work_identifier] = [self.identifier]
```
bkiahstroud, line 12: I haven't run into any issues with this line, which means my `self.identifier` is probably already an array. Can you add `.flatten` to the end of this line just so we don't end up with a nested array please?

bkiahstroud (changes requested):

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -74,8 +74,22 @@ def create_collections
...
         break if records.find_index(collection).present? && limit_reached?(limit, records.find_index(collection))
+        ActiveSupport::Deprecation.warn(
+          'Creating Collections using the collection_field_mapping will no longer be supported as of Bulkrax version 3.0.' \
+          ' Please configure Bulkrax to use related_parents_field_mapping and related_children_field_mapping instead.'
+        )
+
+        metadata = if collection.key?(:title) && collection.keys.count == 1
+          {
+            title: [collection[:title]],
+            source_identifier: [collection[:title]],
```
bkiahstroud, line 85: Look up dynamic `source_identifier` from `field_mappings`

bkiahstroud: Actually, I believe this is [supposed to be work_identifier](https://github.com/samvera-labs/bulkrax/blob/v1.0.0/app/parsers/bulkrax/csv_parser.rb#L61)

bkiahstroud (approved):

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -74,8 +74,28 @@ def create_collections
...
-        new_entry = find_or_create_entry(collection_entry_class, unique_collection_identifier(collection), 'Bulkrax::Importer', collection.to_h)
+        ActiveSupport::Deprecation.warn(
+          'Creating Collections using the collection_field_mapping will no longer be supported as of Bulkrax version 3.0.' \
+          ' Please configure Bulkrax to use related_parents_field_mapping and related_children_field_mapping instead.'
+        )
+
+        ## BEGIN
+        # Add required metadata to collections being imported using the collection_field_mapping, which only have a :title
+        # TODO: Remove once collection_field_mapping is removed
+        metadata = if collection.delete(:from_collection_field_mapping)
```
alishaevn, line 85: @ShanaLMoore @bkiahstroud can you explain this?

bkiahstroud: There's a new way and an [old way](https://github.com/samvera-labs/bulkrax/wiki/Configuring-Bulkrax#collection-field-mapping-) of importing collections. The new way works, but the old way was failing because it didn't have the minimum required metadata. This conditional checks if we're using the old way of importing collections, and if we are, it adds the missing metadata

alishaevn: should have been more specific. I thought that's what adding this "metadata" variable was for...but beforehand it was being added regardless. now it's only being added if we're trying to delete a collection or something?

alishaevn: nevvvvvermind. shana and I talked it out. lol

alishaevn (approved):

PR #408, opened 2022-02-07
Add Collections and FileSets to CSV exports

bkiahstroud:
When exporting "All" or from an Importer, include Collection and FileSet records in the export. 

**NOTE:** when exporting from an importer, Collections that were imported using the deprecated `collection_field_mapping` are not included in the export

### DIFF
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -176,19 +176,24 @@ module Bulkrax
       output
     end
 
-    def current_work_ids
+    def current_record_ids
+      @work_ids, @collection_ids, @file_set_ids = []
+
       case importerexporter.export_from
       when 'all'
-        ActiveFedora::SolrService.query("has_model_ssim:(#{Hyrax.config.curation_concerns.join(' OR ')}) #{extra_filters}", rows: 2_147_483_647).map(&:id)
+        @work_ids = ActiveFedora::SolrService.query("has_model_ssim:(#{Hyrax.config.curation_concerns.join(' OR ')}) #{extra_filters}", rows: 2_147_483_647).map(&:id)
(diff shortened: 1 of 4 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -176,41 +176,70 @@ def extra_filters
...
       when 'importer'
-        entry_ids = Bulkrax::Importer.find(importerexporter.export_source).entries.pluck(:id)
-        complete_statuses = Bulkrax::Status.latest_by_statusable
-                                           .includes(:statusable)
-                                           .where('bulkrax_statuses.statusable_id IN (?) AND bulkrax_statuses.statusable_type = ? AND status_message = ?', entry_ids, 'Bulkrax::Entry', 'Complete')
+        set_ids_for_exporting_from_importer
+      end
+
+      @work_ids + @collection_ids + @file_set_ids
+    end
```
sephirothkod, line 198: Can we alias current_work_ids to current_record_ids so that we don't risk breaking someones custom code? or maybe do this:
```
def current_work_ids
    ActiveSupport::Deprication.warn('current_work_ids will be replaced with current_record_ids in 3.0')
    current_record_ids
end
```

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -176,41 +176,70 @@ def extra_filters
       output
     end
 
-    def current_work_ids
+    def current_record_ids
+      @work_ids = []
+      @collection_ids = []
+      @file_set_ids = []
```
sephirothkod, line 182: I like this set up. Very smart!

sephirothkod (changes requested):

sephirothkod (approved):

Issue #350, opened 2021-09-10
UI Bugs


ShanaLMoore:
**Demo:** https://share.getcloudapp.com/mXuKDkYK

**Steps to reproduce:** 
Create an importer
Go to the importers index page
Click into the created importer
Click the back button (see screenshot)
<img width="1414" alt="Screen Shot 2021-09-10 at 3 43 23 PM" src="https://user-images.githubusercontent.com/10081604/132925111-ecc1b67d-349a-43e8-aca9-2c601779b10f.png">

Notice the following UI issues:
<img width="1182" alt="Screen Shot 2021-09-10 at 3 13 43 PM" src="https://user-images.githubusercontent.com/10081604/132924874-326e152e-c62f-480a-b85b-ec44f4f7f63d.png">

Also this: 
<img width="721" alt="Screen Shot 2021-09-30 at 11 21 46 AM" src="https://user-images.githubusercontent.com/10081604/135510053-499c0d09-6743-494e-a3c2-7d5d23ff626b.png">

bkiahstroud:
This is due to the Turbolinks gem and the DataTables jQuery library not playing nice with each other

ShanaLMoore:
https://github.com/samvera-labs/bulkrax/pull/414

PR #414, opened 2022-02-11
i350 - fix for multile ui elements on index page

ShanaLMoore:
# Bug: 
Clicking on the back button on an importer's show page, to return to the index pages, would produce duplicate elements within the UI. 

# BEFORE 
image

# Summary:
- [ ] tell turbolinks to not cache the JS sections or we get multiple ui elements when clicking the back button

# ref: 
https://github.com/samvera-labs/bulkrax/issues/350

# AFTER: 

i350-demo

### DIFF
--- a/app/views/bulkrax/importers/index.html.erb
+++ b/app/views/bulkrax/importers/index.html.erb
@@ -5,10 +5,10 @@
       <span class="fa fa-edit" aria-hidden="true"></span> <%= t(:'helpers.action.importer.new') %>
     <% end %>
   </div>
-  
-  <% content_for(:head) do %>
-    <meta name="turbolinks-cache-control" content="no-cache">
-  <% end %>
+<% end %>
+
+<% content_for(:head) do %>
+  <meta name="turbolinks-cache-control" content="no-cache">
 <% end %>
 
 <div class="panel panel-default">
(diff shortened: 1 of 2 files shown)


app/views/bulkrax/importers/index.html.erb
```diff
@@ -7,6 +7,10 @@
   </div>
 <% end %>
 
+<% content_for(:head) do %>
+  <meta name="turbolinks-cache-control" content="no-cache">
+<% end%>
```
bkiahstroud, line 12: A few things: 

1. Generally speaking, I think `<meta>` tags are expected at the top of files. Can we move this to the top of this file? 
2. Does this `<meta>` tag stick around on other pages? I'm not sure if it gets cached at all, which could dramatically slow down other pages like the Dashboard > Collections or Dashboard > Works pages 
3. Assuming 2 ☝️ is false, can we also apply this change to the Exporters index page? It also uses DataTables

ShanaLMoore: Sounds good, I'll play around with this on our staging and report back.

bkiahstroud (approved):

app/views/bulkrax/importers/index.html.erb
bkiahstroud: Not necessarily. As long as the `meta` tag doesn't persist on those pages I think it should be fine

sephirothkod (approved):

bkiahstroud:
@ShanaLMoore is this ready to be merged / released?

Issue #213, opened 2020-07-10
Remove "true" option from split in #field_mappings and rename split to split_on


bkiahstroud:
The "default" set of split characters goes back and forth and differs from implementation to implementation; we should remove the default list (aka `split: true`), effectively necessitating that they always have to be specified. 

Also, renaming `split` to `split_on` might be more explicit. E.g. 

```ruby
split_on: ['|', ';', ':']
```

jeremyf:
To preserve backwards compatibility, I'm leaving the `split` parameter and allowing for the value of `true`.  When true it will use the new configuration option on split values.

PR #419, opened 2022-02-16
417 Do not stop all collections from importing when one fails

### LINKED ISSUE #417: If a Collection entry fails during Import, it halts the processing of all following Collection entries

bkiahstroud:
Closes #417 

Use `#perform_later` for collections that are not imported using the `collection_field_mapping`

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -3,7 +3,10 @@
 require 'csv'
 
 module Bulkrax
-  class CsvEntry < Entry
+  # TODO: We need to rework this class some to address the Metrics/ClassLength rubocop offense.
+  # We do too much in these entry classes. We need to extract the common logic from the various
+  # entry models into a module that can be shared between them.
+  class CsvEntry < Entry # rubocop:disable Metrics/ClassLength
     serialize :raw_metadata, JSON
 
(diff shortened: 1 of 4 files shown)


sephirothkod (changes requested):
This looks good for the most part, I would like to change that split so that it runs off of a set variable or defaults to :;| . since if a collection has a colon in the title (ex. "bobs collection: the beginning") it will always get split. Please add a todo or make a ticket for this.

Also there is a better way to do the depreciation, make it conditional with the line that call that specific key ie make it use the same unless statement so it only prints out when that key is called.

bkiahstroud:
> I would like to change that split so that it runs off of a set variable or defaults to :;|

@sephirothkod there's [this ticket](https://github.com/samvera-labs/bulkrax/issues/213), does that still cover what you had in mind or should it be modified?

sephirothkod:
That seems like it is referencing the field mappings, this isn't in the same area and so I am worried it would be missed. We should either create a new ticket for this or just add this specific location to the existing ticket (maybe start a checklist).

bkiahstroud:
@sephirothkod how's this? https://github.com/samvera-labs/bulkrax/issues/429

sephirothkod:
Looks good! Thanks!

sephirothkod (approved):

PR #420, opened 2022-02-16
i421 - enqueued entries count bug

ShanaLMoore:
# Summary

- [ ] Entries Enqueued count would remain at 1, even though the job completed (and sidekiq had no enqueued jobs remaining) (sample had 2 works and 1 collection. The collection had 1 work)
- [ ] status remains as "pending" even though the job completes (when importing a larger sample ie: 2 collections, 4 works, 1 work in one collection, 3 in the other)

co-author @sephirothkod 

## Solution
Even though [current_run](https://github.com/samvera-labs/bulkrax/pull/420/files#diff-4eda1ce8a9d73a758fd0c6bf408cc877e883155d5a3472d372d13c3adab36ec0R25) was being saved, we needed to find it again for it to have the latest count updates from the database on each iteration. 

ref: https://github.com/samvera-labs/bulkrax/issues/421

## BEFORE
Screen Shot 2022-02-16 at 9 23 34 AM
Screen Shot 2022-02-16 at 10 45 20 AM

## AFTER
Screen Shot 2022-02-16 at 9 11 50 AM
Screen Shot 2022-02-16 at 11 31 08 AM

### DIFF
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -85,7 +85,7 @@ module Bulkrax
     end
 
     def create_collections
-      collections.each_with_index do |collection|
+      collections.each_with_index do |collection, index|
         next if collection.blank?
         break if records.find_index(collection).present? && limit_reached?(limit, records.find_index(collection))
         ActiveSupport::Deprecation.warn(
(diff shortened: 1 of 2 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -85,7 +85,7 @@ def valid_import?
     end
 
     def create_collections
-      collections.each_with_index do |collection, index|
+      collections.each_with_index do |collection|
```
alishaevn, line 88:
```suggestion
      collections.each do |collection|
```

alishaevn (changes requested):

app/parsers/bulkrax/csv_parser.rb
ShanaLMoore: changes were reverted so this change no longer applies.

alishaevn (approved):

labradford (approved):

PR #431, opened 2022-03-04
fix attachment concurrency issues and order issues for child objects

sephirothkod:
fixes issues with concurrency on attaching child objects and having the ordered list overwrite itself.


app/jobs/bulkrax/schedule_relationships_job.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Bulkrax
+  class ScheduleRelationshipsJob < ApplicationJob
+    def perform(importer_id:)
+      importer = Importer.find(importer_id)
+      pending_num = Bulkrax::Status.where(statusable_type: "Bulkrax::Entry",
```
sephirothkod, line 7: @orangewolf this is what I cam up with to deal with the the statuses issue for now.

orangewolf: this never runs if any of the entries Fail. that doesn't seem right to me. seems like it should run if all entries are not pending (which is not failed and not complete)

sephirothkod: I saw this when going over things with Kiah, then forgot to change it. It's fixed now.

app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -21,63 +21,60 @@ class CreateRelationshipsJob < ApplicationJob
...
-        @child_record = find_record(child_identifier)
-      else
-        raise ::StandardError, %("#{entry_identifier}" needs either a child or a parent to create a relationship)
+    def perform(parent_identifier:, importer_run_id:)
+      pending_relationships = Bulkrax::PendingRelationship.where(bulkrax_importer_run_id: importer_run_id,
+                                                                 parent_id: parent_identifier).sort_by(&:order)
+      @importer_run_id = importer_run_id
+      @parent_record = find_record(parent_identifier)
+      @child_records = { works: [], collections: [] }
+      pending_relationships.each do |rel|
```
orangewolf, line 40: this should be find_each so it doesn't load them all in memory first (find_each does pages of 500 to prevent memory bloat)

app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -87,18 +84,20 @@ def collection_parent_work_child
...
-        source_identifier_value: nil, # sending the :id in the attrs means the factory doesn't need a :source_identifier_value
-        work_identifier: base_entry.parser.work_identifier,
-        collection_field_mapping: base_entry.parser.collection_field_mapping,
-        replace_files: false,
-        user: user,
-        klass: child_record.class
-      ).run
-      # TODO: add counters for :processed_parents and :failed_parents
-      importer_run.increment!(:processed_relationships) # rubocop:disable Rails/SkipsModelValidations
+      child_records[:works].each do |child_record|
```
orangewolf, line 87: find_each (though less critical here)

app/jobs/bulkrax/schedule_relationships_job.rb
sephirothkod: ahh I didn't realize that... so something like:

```
pending_num = importer.entries.left_outer_joins(:latest_status).where('bulkrax_statuses.status_message <> ?', %w[Complete Failed]).count
```

that should grab all entries without a complete or failed status

sephirothkod: or should I just use the NULL one and assume that we are only looking for `pending` statuses?

orangewolf: I think the null one is probably safe

PR #449, opened 2022-03-30
Fix NoMethodError on ImporterRun#parents

bkiahstroud:
## Intended for v3.0 major release

After merging, release v3.0.0.beta3

## Summary 

`ImporterRun#parents` was being set to a String instead of an Array: 

```ruby
i.last_run.parents
=> "[]irc1irc1" # should be ["irc1", "irc1"]
```

### Fix 

Serialize `ImporterRun#parents` as an Array. This fixes this bug: 

DEVELOPMENT  Sidekiq 2022-03-29 at 11 41 17 AM

### Why the database change? 

This bug was not a problem on projects using postgres because postgres natively supports storing data as arrays. However, other databases (e.g. mysql) do _not_ natively storing data as arrays. Because of this, `#serialize` is preferable since it has more compatibility

### DIFF
--- a/app/models/bulkrax/importer_run.rb
+++ b/app/models/bulkrax/importer_run.rb
@@ -4,5 +4,7 @@ module Bulkrax
   class ImporterRun < ApplicationRecord
     belongs_to :importer
     has_many :statuses, as: :runnable, dependent: :destroy
+
+    serialize :parents, Array
   end
 end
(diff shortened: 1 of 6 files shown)


alishaevn (approved):

alishaevn:
tested this locally on louisville to confirm that the changes in this pr will also work with a postgres db.

PR #451, opened 2022-03-31
update-collection-handling

alishaevn:
# TODO
discuss with braydon

# expected behavior
- determine possible collection ids by source_identifier and not title

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -41,6 +41,7 @@ module Bulkrax
       self.parsed_metadata = {}
       add_identifier
       add_ingested_metadata
+      # TODO(alishaevn): remove the collections stuff entirely and only reference collections via the new parents code
       add_collections
       add_visibility
       add_metadata_for_model
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -248,7 +248,7 @@ module Bulkrax
       identifiers = []
       split_references = record[parent_field_mapping].split(/\s*[;|]\s*/)
       split_references.each do |c_reference|
-        matching_collection_entries = importerexporter.entries.select { |e| e.raw_metadata['source_identifier'] == c_reference && e.is_a?(CsvCollectionEntry) }
+        matching_collection_entries = importerexporter.entries.select { |e| e.raw_metadata[work_identifier] == c_reference && e.is_a?(CsvCollectionEntry) }
         raise ::StandardError, 'Only expected to find one matching entry' if matching_collection_entries.count > 1
         identifiers << matching_collection_entries.first&.identifier
       end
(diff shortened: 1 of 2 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -41,6 +41,7 @@ def build_metadata
       self.parsed_metadata = {}
       add_identifier
       add_ingested_metadata
+      add_collections
```
alishaevn, line 44: @sephirothkod I don't think this is a good choice for bulkrax because this line seems to have been removed intentionally for some reason. however, without it `self.collection_ids` is always empty so `collections_created?` will return false and throw a CollectionCreatedError. even when the collection is in fact created.

app/models/bulkrax/csv_entry.rb
```diff
@@ -241,16 +242,17 @@ def self.matcher_class
...
-      return [] unless collection_field_mapping.present? && record[collection_field_mapping].present?
+      parent_field_mapping = self.class.parent_field(parser)
+      return [] unless parent_field_mapping.present? && record[parent_field_mapping].present?
 
       identifiers = []
-      split_titles = record[collection_field_mapping].split(/\s*[;|]\s*/)
-      split_titles.each do |c_title|
-        matching_collection_entries = importerexporter.entries.select { |e| e.raw_metadata['title'] == c_title }
+      split_references = record[parent_field_mapping].split(/\s*[;|]\s*/)
+      split_references.each do |c_reference|
```
alishaevn, line 250: more semantic names.

app/models/bulkrax/csv_entry.rb
```diff
@@ -241,16 +242,17 @@ def self.matcher_class
...
+        matching_collection_entries = importerexporter.entries.select { |e| e.raw_metadata['source_identifier'] == c_reference && e.is_a?(CsvCollectionEntry) }
```
alishaevn, line 251: we're no longer using titles to track collections. we also are making sure that we only return collection parents and not work parents so that `collections_created?` isn't incorrectly throwing false.

app/models/bulkrax/csv_entry.rb
alishaevn: @sephirothkod @bkiahstroud actually...I now needed to bring this code into LV and BL through overrides. may it be better to put this code into bulkrax proper, but with a note to reevaluate removing `add_collections` from here in the future once the updated logic is worked out?

alishaevn: @sephirothkod @bkiahstroud actually...I now needed to bring this code into LV and BL through overrides. may it be better to put this code into bulkrax proper, but with a note to reevaluate removing `add_collections` from here in the future once the updated logic is worked out?

sephirothkod: sure lets leave it for now and then I'll have a look when i have time as to how to remove the collections stuff entirely and  only reference collections via the new parents code.

sephirothkod (approved):

PR #453, opened 2022-04-05
remove-empty-id-when-setting-object-attrs

alishaevn:
# use case
british libraries has an "id" column on their csv's that isn't always filled in. as is, the code will error out at the line below if a blank id field is present in `attrs`
``` ruby
object.attributes = attrs
```

# expected behavior
- remove id when `attrs[id]` is not present so it does not throw errors

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,7 @@ GIT
 PATH
   remote: .
   specs:
-    bulkrax (3.0.0.beta2)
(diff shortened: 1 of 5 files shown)


ShanaLMoore (approved):

ShanaLMoore:
approval is contingent upon passing specs though

sephirothkod (approved):

PR #454, opened 2022-04-05
accurate-relationship-forming

alishaevn:
# use case
I imported the same collection on 3 different importers while upgrading bl. the first time the collection wasn't created. so on the 3rd importer where the collection had been created, it was still finding the entry from the first importer. and therefore `entry.factory.find` was nil when it shouldn't have been.

# expected behavior
- get the entry from the current importer and not the first entry, if applicable
- return the entry and fedora record from `find_record`
- update code around importing FileSet rows

# demo
| works | file set |
| -- | -- |
| image | image|

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -7,10 +7,10 @@ module Bulkrax
     include DynamicRecordLookup
 
     define_model_callbacks :save, :create
-    attr_reader :attributes, :object, :source_identifier_value, :klass, :replace_files, :update_files, :work_identifier, :related_parents_parsed_mapping
+    attr_reader :attributes, :object, :source_identifier_value, :klass, :replace_files, :update_files, :work_identifier, :related_parents_parsed_mapping, :importer_run_id
 
     # rubocop:disable Metrics/ParameterLists
-    def initialize(attributes:, source_identifier_value:, work_identifier:, related_parents_parsed_mapping: nil, replace_files: false, user: nil, klass: nil, update_files: false)
+    def initialize(attributes:, source_identifier_value:, work_identifier:, related_parents_parsed_mapping: nil, replace_files: false, user: nil, klass: nil, importer_run_id: nil, update_files: false)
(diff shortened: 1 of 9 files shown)


sephirothkod (approved):
good catch on this, we do need to specify the entry for the specific importer!

app/factories/bulkrax/object_factory.rb
```diff
@@ -152,7 +153,7 @@ def update_collection(attrs)
 
     # This method is heavily inspired by Hyrax's AttachFilesToWorkJob
     def create_file_set(attrs)
-      work = find_record(attributes[related_parents_parsed_mapping].first)
+      _, work = find_record(attributes[related_parents_parsed_mapping].first, importer_run_id)
```
alishaevn, line 156: `find_record` now returns two values, but we don't need the entry here

app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -37,11 +37,11 @@ def perform(parent_identifier:, importer_run_id:)
       end.sort_by(&:order)
 
       @importer_run_id = importer_run_id
-      @parent_record = find_record(parent_identifier)
+      @parent_entry, @parent_record = find_record(parent_identifier, importer_run_id)
```
alishaevn, line 40: example use case that made this necessary:
a collection was imported in importer 1, and in importer 2 we are importing a work to add to that collection. in that case, line 56 below would have failed because this importer_run doesn't have the collection.

app/jobs/bulkrax/import_file_set_job.rb
```diff
@@ -63,7 +66,8 @@ def check_parent_is_a_work!(parent_identifier)
     end
 
     def find_parent_record(parent_identifier)
-      @parent_record ||= find_record(parent_identifier)
+      @parent_record ||= find_record(parent_identifier, importer_run_id)
```
alishaevn, line 69: I tried calling `.last` on this method but it didn't work.

ShanaLMoore (approved):

PR #457, opened 2022-04-07
Bulkrax 3+ update to display collection title on importer show page w…

ShanaLMoore:
Bulkrax 3+ update to display collection title on importer show page work entries section

The current code displays the title of a collection if the csv is imported the collection_field_mapping way (w the collection header) because it takes the value off of the collections key of parsed or raw metadata, even though a Collection can no longer be created this way. 

Since that has been deprecated, this code updates how titles get retrieved.

BEFORE
image

AFTER

Screen Shot 2022-04-07 at 1 35 06 PM

### DIFF
--- a/app/views/bulkrax/importers/show.html.erb
+++ b/app/views/bulkrax/importers/show.html.erb
@@ -97,13 +97,8 @@
               <% @work_entries.each do |e| %>
                 <tr>
                   <td><%= link_to e.identifier, bulkrax.importer_entry_path(@importer.id, e.id) %></td>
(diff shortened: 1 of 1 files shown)


ShanaLMoore:
In Draft to check if a Collection can have more than 1 title?

alishaevn (approved):

alishaevn:
> In Draft to check if a Collection can have more than 1 title?

re: https://github.com/samvera/hyrax/blob/v2.9.6/app/models/concerns/hyrax/core_metadata.rb

PR #462, opened 2022-04-13
Draft: export thumbnails

ShanaLMoore:
# Summary

PALS would like to contribute back the following work: 

1. The ability to export thumbnails, similar to the way files get exported.

## Screenshots

From the export form, if a user checks to include thumbnails the csv will get exported with a "thumbnail_file" column header and filenames as its values. The actual files/thumbnails will be included in the files directory of the exported zipped file. 
Screen Shot 2022-04-12 at 4 32 41 PM
Screen Shot 2022-04-12 at 4 32 12 PM
Screen Shot 2022-04-12 at 4 24 29 PM


ShanaLMoore:
TODO: add specs

PR #463, opened 2022-04-13
streamline parents lookup per run.

sephirothkod:
remove database column for parents and just use relationships look up each time.

### DIFF
--- a/app/models/bulkrax/importer_run.rb
+++ b/app/models/bulkrax/importer_run.rb
@@ -4,5 +4,9 @@ module Bulkrax
   class ImporterRun < ApplicationRecord
     belongs_to :importer
     has_many :statuses, as: :runnable, dependent: :destroy
+
+    def parents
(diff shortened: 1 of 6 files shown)


app/models/bulkrax/importer_run.rb
```diff
@@ -5,6 +5,8 @@ class ImporterRun < ApplicationRecord
     belongs_to :importer
     has_many :statuses, as: :runnable, dependent: :destroy
 
-    serialize :parents, Array
+    def parents
```
alishaevn, line 8: does this get called now with `importer.last_run.parents` in the schedule relationships job?

alishaevn, line 8: does this get called now with `importer.last_run.parents` in the schedule relationships job?

sephirothkod: yes; that is the idea at least.

alishaevn (approved):
this is working for me locally in louisville

PR #465, opened 2022-04-13
identifier-mixup

alishaevn:
# expected behavior
- use `source_identifier` instead of `work_identifier`
- `source_identifier` is the value on the csv and the "from" in bulkrax.rb
- `work_identifier` is the "to" value in bulkrax.rb
- these are not always the same

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -249,7 +249,7 @@ module Bulkrax
       identifiers = []
       split_references = record[parent_field_mapping].split(/\s*[;|]\s*/)
       split_references.each do |c_reference|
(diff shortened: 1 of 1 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -249,7 +249,7 @@ def possible_collection_ids
       identifiers = []
       split_references = record[parent_field_mapping].split(/\s*[;|]\s*/)
       split_references.each do |c_reference|
-        matching_collection_entries = importerexporter.entries.select { |e| e.raw_metadata[work_identifier] == c_reference && e.is_a?(CsvCollectionEntry) }
+        matching_collection_entries = importerexporter.entries.select { |e| e.raw_metadata[source_identifier] == c_reference && e.is_a?(CsvCollectionEntry) }
```
alishaevn, line 252: use the `source_identifier` value since that's what's on the csv and we're looking at the raw metadata.

sephirothkod (approved):
good catch!

Issue #475, opened 2022-04-19
Unable to add collection to collections


ShanaLMoore:
# Summary

ActiveFedora::ObjectNotFoundError is received when trying to create a collection and add it to another collection, on the fly.

Sample data: 
<img width="325" alt="Screen Shot 2022-04-18 at 2 53 04 PM" src="https://user-images.githubusercontent.com/10081604/164113810-9c95df62-1eab-4fe4-902a-73d0025b45f4.png">

Sample error: 
<img width="1154" alt="Screen Shot 2022-04-19 at 3 48 40 PM" src="https://user-images.githubusercontent.com/10081604/164113865-2063424f-a7d2-4c8a-89e8-cb9e39836e77.png">
<img width="671" alt="Screen Shot 2022-04-18 at 2 09 39 PM" src="https://user-images.githubusercontent.com/10081604/164113908-d5c2deb6-9578-4484-9f13-b4bf9ca319f4.png">

ShanaLMoore:
solved and merged per [PR](https://github.com/samvera-labs/bulkrax/pull/508)

Issue #471, opened 2022-04-19
unable to reference a parent work/collection by id in 3.0+


alishaevn:
# summary
at least as of `v.3.0.0.beta3`, and potentially as far back as `v.3.0.0.beta1`, the "parents" and "children" columns are expecting a `source_identifier`. this means that users cannot add children to previously created or imported works/collections with just an id.

# acceptance criteria
- [ ] users can add a child work to a preexisting collection using the collection's id
- [ ] users can add a child work to a preexisting work using the preexisting work's id

PR #477, opened 2022-04-19
unable to reference a parent work/collection by id in 3.0+

alishaevn:
ref: #471 && #475

# expected behavior
- users can add a child work to a preexisting collection using the collection's id
- users can add a child work to a preexisting work using the preexisting work's id
- users can add a child collection to a preexisting collection using the preexisting collections's id

# demo
| work to collection | work to work |  collection to collection |
| --- | --- | --- |
| image | image | image |

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -140,14 +140,14 @@ module Bulkrax
     def create_collection(attrs)
       attrs = clean_attrs(attrs)
       attrs = collection_type(attrs)
-      persist_collection_memberships(parent: find_collection(attributes[related_parents_parsed_mapping]), child: object) if attributes[related_parents_parsed_mapping].present?
+      pass_collection_to_be_persisted(parent: find_collection(attributes[related_parents_parsed_mapping]), child: object) if attributes[related_parents_parsed_mapping].present?
       object.attributes = attrs
       object.apply_depositor_metadata(@user)
       object.save!
     end
 
     def update_collection(attrs)
-      persist_collection_memberships(parent: find_collection(attributes[related_parents_parsed_mapping]), child: object) if attributes[related_parents_parsed_mapping].present?
+      pass_collection_to_be_persisted(parent: find_collection(attributes[related_parents_parsed_mapping]), child: object) if attributes[related_parents_parsed_mapping].present?
       object.attributes = attrs
       object.save!
     end
@@ -182,12 +182,20 @@ module Bulkrax
       actor.update_metadata(file_set_attrs)
     end
 
+    def pass_collection_to_be_persisted(parent:, child:)
+      if parent.is_a? Array
+        parent.each { |par| persist_collection_memberships(par, child) }
+      else
+        persist_collection_memberships(parent, child)
+      end
+    end
+
     # Add child to parent's #member_collections
     # Add parent to child's #member_of_collections
-    def persist_collection_memberships(parent:, child:)
+    def persist_collection_memberships(parent, child)
       parent.reject!(&:blank?) if parent.respond_to?(:reject!)
       child.reject!(&:blank?) if child.respond_to?(:reject!)
-      return if parent.blank? || child.blank?
+      return if parent.invalid? || child.invalid?
 
       ::Hyrax::Collections::NestedCollectionPersistenceService.persist_nested_collection_for(parent: parent, child: child)
     end
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -140,16 +140,16 @@ module Bulkrax
     def create_collection(attrs)
       attrs = clean_attrs(attrs)
       attrs = collection_type(attrs)
-      pass_collection_to_be_persisted(parent: find_collection(attributes[related_parents_parsed_mapping]), child: object) if attributes[related_parents_parsed_mapping].present?
       object.attributes = attrs
       object.apply_depositor_metadata(@user)
       object.save!
+      pass_collection_to_be_persisted(parent: find_collection(attributes[related_parents_parsed_mapping]), child: object) if attributes[related_parents_parsed_mapping].present?
     end
 
     def update_collection(attrs)
-      pass_collection_to_be_persisted(parent: find_collection(attributes[related_parents_parsed_mapping]), child: object) if attributes[related_parents_parsed_mapping].present?
       object.attributes = attrs
       object.save!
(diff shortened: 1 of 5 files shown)


app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -87,8 +87,8 @@ def collection_parent_work_child
         ObjectFactory.new(
           attributes: attrs,
           source_identifier_value: nil, # sending the :id in the attrs means the factory doesn't need a :source_identifier_value
-          work_identifier: parent_entry.parser.work_identifier,
-          related_parents_parsed_mapping: parent_entry.parser.related_parents_parsed_mapping,
+          work_identifier: parent_entry&.parser&.work_identifier,
```
alishaevn, line 90: when adding a work to an existing work or collection, the parent_entry doesn't exist. these values being nil didn't cause any errors for me though. 🙌🏾

alishaevn:
@bkiahstroud can you test this out on ucsc when you get a chance?

alishaevn:
confirming that this also worked on louisville.

app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -41,7 +41,7 @@ def perform(parent_identifier:, importer_run_id:)
       @child_records = { works: [], collections: [] }
       pending_relationships.each do |rel|
         raise ::StandardError, %("#{rel}" needs either a child or a parent to create a relationship) if rel.child_id.nil? || rel.parent_id.nil?
-        _, child_record = find_record(rel.child_id, importer_run_id)
+        @child_entry, child_record = find_record(rel.child_id, importer_run_id)
```
orangewolf, line 44: does this need to be an instance variable?

alishaevn: mmm, nope!

app/factories/bulkrax/object_factory.rb
```diff
@@ -182,12 +182,20 @@ def update_file_set(attrs)
       actor.update_metadata(file_set_attrs)
     end
 
+    def pass_collection_to_be_persisted(parent:, child:)
```
alishaevn, line 185: I created this method because the `parent` that gets returned from the `find_collection` method down below has always been an array.

I'm not sure when it will be a hash or a string (except when an array of strings is being mapped over), but the options are there so I didn't wanna discount them.

@orangewolf @sephirothkod @bkiahstroud have any of you recently seen where the id being passed into the `find_collection` method is a hash or string? do we still need to account for "when Hash"?

app/factories/bulkrax/object_factory.rb
```diff
@@ -182,12 +182,20 @@ def update_file_set(attrs)
...
+    end
+
     # Add child to parent's #member_collections
     # Add parent to child's #member_of_collections
-    def persist_collection_memberships(parent:, child:)
+    def persist_collection_memberships(parent, child)
       parent.reject!(&:blank?) if parent.respond_to?(:reject!)
       child.reject!(&:blank?) if child.respond_to?(:reject!)
-      return if parent.blank? || child.blank?
+      return if parent.invalid? || child.invalid?
```
alishaevn, line 198: this method is only called when a collection has a parent collection listed. the child at this time is a new instance of a collection that hasn't been saved so the next line fails in [active fedora](https://github.com/samvera/active_fedora/blob/v12.1.1/lib/active_fedora/relation/finder_methods.rb#L182) for having no id. (see below)

changing the method to `invalid?` means the method is exited before failing at the "::Hyrax::Collections::NestedCollectionPersistenceService..." line below.

the new collection is then properly saved and the relationship to its parent collection is formed, but idk if this is the best way to go about it.

@orangewolf @sephirothkod @bkiahstroud thoughts?
Screen Shot 2022-04-21 at 5 56 33 PM

alishaevn:
when we discussed this pr the other day, a concern was brought up that the `log_created`, `log_updated` and `log_deleted_fs` methods wouldn't log correctly anymore now that `work_identifier` and  `related_parents_parsed_mapping` could be nil.

it appears however that this isn't the case. 🎉 

image

app/jobs/bulkrax/create_relationships_job.rb
alishaevn: yes. otherwise `child_entry` in the rescue below will be nil.

app/factories/bulkrax/object_factory.rb
bkiahstroud: > I'm not sure when it will be a hash or a string (except when an array of strings is being mapped over) 

For this reason, I'm confident that both Array and String are valid here. 

As far as Hash goes, I've never seen this used. Personally, I'd remove it unless someone knows for sure if it gets used by something I'm not super familiar with (i.e. the OAI or Bagit parsers). If we need to handle Hashes in the future, we can always add it back in.

app/factories/bulkrax/object_factory.rb
bkiahstroud: Hmm... this confuses me for a few reasons: 

1. This was working as is in v2.3.0 when I tested it 
2. > the new collection is then properly saved and the relationship to its parent collection is formed

  2a. The fact that the relationship still gets created properly despite bailing out of this method is odd because, as far as I was aware, this was the only thing that actually created the relationship between the collections, so I'm not sure what's going on there. 

Ultimately, I don't think we should just no-op out of this method 100% of the time. If it requires the objects to be persisted, maybe we could call it after the `object.save!` lines in the `#create_collection` and `#update_collection` methods?

alishaevn: thanks kiah. I considered calling the method after the objects are saved too if it's necessary that it be called despite the relationships forming without it.

sephirothkod: After looking through I agree with moving this after the object.save call. If that causes issues for some reason, we can move it back and do a check on the child collection and save if it hasn't been saved yet.

app/factories/bulkrax/object_factory.rb
sephirothkod: I feel like I remember bagit having id's in hashes as a possibility, let's keep accounting for this.

sephirothkod (changes requested):

app/factories/bulkrax/object_factory.rb
alishaevn: changed the code back to using `.blank?` and moved persisting the collection to after the new collection is saved. this still works successfully.

app/factories/bulkrax/object_factory.rb
sephirothkod: Whoops, I didn't mean to request a change here. I meant it to be for the call being moved to after the object.save sorry for the confusion!

sephirothkod (approved):

PR #494, opened 2022-04-26
Random nomethoderror import

selamhabteab:
### Summary
Inconsistent error on first run of import: NoMethodError - undefined method '[]' for nil. 
Error clears after a re-run.

Found that the raw metadata of imported entries returns a nil value due to import jobs still loading, hence the NoMethodError on the first run and not the rerun. Issue is resolved with `&.` that prevents block from erroring out


bkiahstroud:
Is this supposed to be going into the `export-enhancements` branch? If so, should it also go into `main`?

alishaevn (approved):

Issue #495, opened 2022-04-27
Bug: Create and Validate w Collections causes error


ShanaLMoore:
# Summary
User receives an error message when selecting "create and validate" on a new bulkrax csv importer. 

The error is valid because "create and validate" does not import anything. Therefore, when the code checks for the collection's id, it is correct that it does not exist. 

The code should not be checking for the existence of a collection at this point, when "create and validate" is chosen. 

**Current bulkrax version**: 3.0.0beta5

# Acceptance Criteria

* [ ] selecting "create and validate" on a csv bulkrax importer should not throw an error. 
* Csv should be set up to create works and collections the 3.0+ way.

# Screenshots or Video

URL (Staging): https://lv-hyku-staging.notch8.cloud/importers/26?locale=en#collection-entries

DEMO: https://share.getcloudapp.com/RBun27ZP

<img width="1425" alt="Screen Shot 2022-04-27 at 9 04 25 AM" src="https://user-images.githubusercontent.com/10081604/165562292-2c47d92e-90b8-4ce1-bc4b-ac166d825ffa.png">

# Testing Instructions

* Create and Validate the sample file using a csv importer
* sample file: [shanas-rhino.zip](/uploads/fc8addddaea5ea78716980c4c3ac4552/shanas-rhino.zip)

ShanaLMoore:
solved and merged per [PR](https://github.com/samvera-labs/bulkrax/pull/506)

PR #506, opened 2022-05-05
Don't create permission template accesses for collections on validate only

ShanaLMoore:
# Summary

ref: https://github.com/samvera-labs/bulkrax/issues/495

Don't create permission template accesses for collections on validate only

authored by braydon j.


ShanaLMoore (commented):
@sephirothkod  tested on pals and lv ✅ create and validate no longer errors out. 

But testing further for thumbnail issues.

sephirothkod (approved):

PR #508, opened 2022-05-05
Remove legacy collection relationships logic from object factory

kirkkwang:
Separated relationships handling away from `ObjectFactory` and into `CreateRelationshipsJob` that was creating an `ActiveFedora::ObjectNotFoundError`

Relationships include:
- Collection Parent -> Collection Child
- Collection Parent -> Work Child
- Work Parent -> Work Child


sephirothkod (approved):
I made one change to clarify a spec (just the descriptor)

kirkkwang:
> I made one change to clarify a spec (just the descriptor)

@sephirothkod perfect thank you for all your help!

PR #511, opened 2022-05-10
Propertly differential CSV Entry types

bkiahstroud:
skip if model_mapping is blank to avoid shoveling everything into `@works` when there are multiple model_mappings


alishaevn (approved):
this is working for bl

PR #515, opened 2022-05-16
Account for commons mistakes when importing controlled Questioning Authority fields

bkiahstroud:
# Summary

For default Hyrax local Questioning Authority controlled fields (`:rights_statement` and `:license`): 

- Use appropriate protocol (`https` or `http`) depending on IDs in `config/authorities/` YAML files 
- Add missing trailing slashes
- Remove duplicate `:removed_image_path` from `Bulkrax` config

### DIFF
--- a/app/models/concerns/bulkrax/import_behavior.rb
+++ b/app/models/concerns/bulkrax/import_behavior.rb
@@ -5,8 +5,6 @@ module Bulkrax
   module ImportBehavior
     extend ActiveSupport::Concern
 
-    LOCAL_QA_FIELDS = %w[rights_statement license].freeze
-
     def build_for_importer
       begin
         build_metadata
@@ -109,7 +107,7 @@ module Bulkrax
 
     # TODO: documentation
     def sanitize_qa_uri_values!
-      LOCAL_QA_FIELDS.each do |field|
+      Bulkrax.qa_controlled_properties.each do |field|
         next if parsed_metadata[field].blank?
 
         parsed_metadata[field].each_with_index do |value, i|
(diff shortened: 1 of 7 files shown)


app/models/concerns/bulkrax/import_behavior.rb
```diff
@@ -103,6 +105,39 @@ def add_collections
...
+    # TODO: documentation
+    def sanitize_qa_uri_values!
+      LOCAL_QA_FIELDS.each do |field|
+        next if parsed_metadata[field].blank?
+
+        parsed_metadata[field].each_with_index do |value, i|
+          next if value.blank?
+
+          if value.match?(::URI::DEFAULT_PARSER.make_regexp)
+            value = value.strip.chomp.sub('https', 'http')
```
sephirothkod, line 117: Maybe a second value in the initializer or create a hash for which ones need to be converted to http.

bkiahstroud: Is there possibly a more programmatic solution? What if we checked the service for both http and https versions and then did a sub accordingly? @sephirothkod

bkiahstroud: Oh I see, by config option you mean declaring whether they should ALL be http or ALL be https?

sephirothkod: checking the service is probably the best way to go. check the raw link, if it doesn't exist in the QA do something like:
```
if value.contain?("https")
  value = value.sub('https', 'http')
else
  value = value.sub('http', 'https')
end
```
then try again. error if it still doesn't exist in the QA

sephirothkod: you might be able to do a ternary of the above.

sephirothkod (approved):

PR #516, opened 2022-05-17
added missing method for BagIt parser

kirkkwang:


app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -23,6 +23,10 @@ def collection_entry_class
       Entry
     end
 
+    def file_set_entry_class
+      CsvFileSetEntry
+    end
```
bkiahstroud, line 28: I think this method should look more like `#collection_entry_class`, except instead of subbing with `'CollectionEntry'` we sub with `'FileSetEntry'`. I don't think using a CSV entry in the Bagit parser is correct

bkiahstroud (changes requested):

app/parsers/bulkrax/bagit_parser.rb
kirkkwang: Ah I see, but if I'm importing a bag with a `metadata.csv` file, would that still be able to get parsed?

kirkkwang: That's a good point, according to the [wiki](https://github.com/samvera-labs/bulkrax/wiki/BagIt-Importer), the metadata either can be an RDF or CSV.  Looks like there needs to be a determination of what the metadata is supplied, then if it's as `.csv` would we go with `CsvFileSetEntry` and if it's a `.rdf` use `RdfEntry`?  Something to that extent?

bkiahstroud: That sounds right to me!

alishaevn: potential solution 
``` ruby
 Bulkrax::Importer.last.parser_fields['metadata_format'] == "Bulkrax::CsvEntry" ? CsvFileSetEntry : <something? I don't see an rdf file set entry class>
```

alishaevn (approved):

bkiahstroud (approved):

alishaevn:
I tested creating a csv bagit importer with the new changes that reference `FileSetEntryBehavior` on pals and it worked for me.

image

PR #517, opened 2022-05-17
Erroneous complete status

kirkkwang:
Ref: https://gitlab.com/notch8/palni-palci/-/issues/468

Before: Although it says "Complete", it is actually not complete and that is why the download button is not there.
image

After:
| while pending | upon completion |
| --- | --- |
| image | image|

### DIFF
--- a/app/models/bulkrax/exporter.rb
+++ b/app/models/bulkrax/exporter.rb
@@ -96,7 +96,8 @@ module Bulkrax
     end
 
     def current_run
-      @current_run ||= self.exporter_runs.create!(total_work_entries: self.limit || parser.total, enqueued_records: self.limit || parser.total)
+      total = self.limit || parser.total
+      @current_run ||= self.exporter_runs.create!(total_work_entries: total, enqueued_records: total)
     end
 
     def last_run
--- a/app/models/bulkrax/exporter.rb
+++ b/app/models/bulkrax/exporter.rb
@@ -96,7 +96,7 @@ module Bulkrax
     end
 
     def current_run
-      @current_run ||= self.exporter_runs.create!(total_work_entries: self.limit || parser.total)
+      @current_run ||= self.exporter_runs.create!(total_work_entries: self.limit || parser.total, enqueued_records: self.limit || parser.total)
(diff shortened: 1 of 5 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -240,7 +240,7 @@ def create_new_entries
         new_entry = find_or_create_entry(this_entry_class, id, 'Bulkrax::Exporter')
 
         begin
-          entry = ExportWorkJob.perform_now(new_entry.id, current_run.id)
+          entry = ExportWorkJob.perform_now(new_entry.id, current_run.id, current_record_ids.length)
```
bkiahstroud, line 243: This third argument can be removed now

app/models/bulkrax/exporter.rb
```diff
@@ -96,7 +96,7 @@ def importers_list
     end
 
     def current_run
-      @current_run ||= self.exporter_runs.create!(total_work_entries: self.limit || parser.total)
+      @current_run ||= self.exporter_runs.create!(total_work_entries: self.limit || parser.total, enqueued_records: self.limit || parser.total)
```
bkiahstroud, line 99: Since we're setting both counters to the same value, this could be made slightly cleaner by doing something like this: 

```ruby
    def current_run
      total = self.limit || parser.total
      @current_run ||= self.exporter_runs.create!(total_work_entries: total, enqueued_records: total)
```

kirkkwang: Awesome thanks, made the changes!

bkiahstroud:
@alishaevn @sephirothkod I'm requesting reviews from one/both of you since @kirkkwang and I paired on this for the most part

bkiahstroud (commented):
@kirkkwang also, can we add some test coverage for these changes to make sure they don't break the same way in the future?

alishaevn (approved):

sephirothkod (approved):

bkiahstroud (approved):
@kirkkwang looks great to me!

PR #519, opened 2022-05-18
Fix Collection-to-Work relationship assigning

bkiahstroud:
# Summary 

`Hyrax::Collections::NestedCollectionPersistenceService#persist_nested_collection_for` is only meant to be used for Collection-to-Collection relationships. Because of this, Collection-to-Work relationships were broken. 

Instead, use [Collection#add_member_objects](https://github.com/samvera/hyrax/blob/v2.9.6/app/controllers/hyrax/dashboard/collection_members_controller.rb#L29), which is used in the `Hyrax::Dashboard::CollectionMembersController`

### DIFF
--- a/app/models/concerns/bulkrax/import_behavior.rb
+++ b/app/models/concerns/bulkrax/import_behavior.rb
@@ -21,6 +21,8 @@ module Bulkrax
         status_info(e)
       else
         status_info
+      ensure
+        self.save!
       end
       return @item
     end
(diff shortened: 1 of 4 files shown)


app/models/concerns/bulkrax/import_behavior.rb
```diff
@@ -21,6 +21,8 @@ def build_for_importer
         status_info(e)
       else
         status_info
+      ensure
+        self.save!
```
bkiahstroud, line 25: Occasionally, Entries were being processed without saving at the end. This would lead to imported records and "Complete" statuses without the entry having any parsed metadata set

sephirothkod (approved):

PR #524, opened 2022-05-20
Erroneous pending status on import

kirkkwang:
`#record_status` which updates the status wasn't being called when a collection finishes however it was called when a work was completed.

This seemingly meant that if the works finish faster than the collections the status gets stuck on "Pending".

### DIFF
--- a/app/views/bulkrax/importers/index.html.erb
+++ b/app/views/bulkrax/importers/index.html.erb
@@ -38,7 +38,7 @@
             <% @importers.each do |importer| %>
               <tr>
                 <th scope="row"><%= link_to importer.name, importer_path(importer) %></th>
-                <td><%= importer.status %></td>
+                <td><%= importer.last_run&.enqueued_records&.zero? ? 'Complete' : importer.status %></td>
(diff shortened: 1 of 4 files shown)


app/views/bulkrax/importers/index.html.erb
```diff
@@ -38,7 +38,7 @@
             <% @importers.each do |importer| %>
               <tr>
                 <th scope="row"><%= link_to importer.name, importer_path(importer) %></th>
-                <td><%= importer.status %></td>
+                <td><%= importer.last_run&.enqueued_records&.zero? ? 'Complete' : importer.status %></td>
```
alishaevn, line 41: is this still needed now that `record_status` is being called on collection entries?

kirkkwang: I believe so, sometimes the works finish first and sometimes the collections finish first. I think that's why it felt so random when we were testing it.

kirkkwang: Ohh that part no

alishaevn (approved):

PR #527, opened 2022-05-20
patch version bump v3.1.1

bkiahstroud:


alishaevn:
do you mind holding off on this until https://github.com/samvera-labs/bulkrax/pull/516 gets merged in too? hopefully it'll also be today and we won't need to turn around and do another version update in a few hours.

(might also be able to sneak in #524 if the specs act right.

bkiahstroud:
@alishaevn is this good to merge now?

alishaevn:
> @alishaevn is this good to merge now?

good to go! thanks for waiting!

PR #530, opened 2022-05-23
Stop gt 255 filenames breaking export

cziaarm:
Conditionally truncate filenames to stop them going going over 255 chars when they have fileset_ids prepended to them.

### DIFF
--- a/app/models/concerns/bulkrax/export_behavior.rb
+++ b/app/models/concerns/bulkrax/export_behavior.rb
@@ -57,7 +57,7 @@ module Bulkrax
       end
       # Remove extention truncate and reattach
       ext=File.extname(filename)
-      filename = "#{File.basename(filename, ext)[0...(255-ext.length)]}#{ext}"
+      "#{File.basename(filename, ext)[0...(255-ext.length)]}#{ext}"
     end
   end
 end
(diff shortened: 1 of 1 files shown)


app/models/concerns/bulkrax/export_behavior.rb
```diff
@@ -42,19 +42,22 @@ def write_files
...
       if fn.include?(file_set.id) || importerexporter.metadata_only?
-        return fn if mime.to_s == ext_mime.to_s
-        return "#{fn}.#{mime.to_sym}"
+        filename = "#{fn}.#{mime.to_sym}"
+        filename = fn if mime.to_s == ext_mime.to_s
       else
-        return "#{file_set.id}_#{fn}" if mime.to_s == ext_mime.to_s
-        return "#{file_set.id}_#{fn}.#{mime.to_sym}"
+        filename = "#{file_set.id}_#{fn}.#{mime.to_sym}"
+        filename = "#{file_set.id}_#{fn}" if mime.to_s == ext_mime.to_s
```
alishaevn, line 56: same question here as the above.

cziaarm: same answer

cziaarm:
I don't have perms to add a label, which is the only failing check left

kirkkwang:
> I don't have perms to add a label, which is the only failing check left

I can add it for you, which label?

cziaarm:
Thanks Kirk, I was thinking bug-fix?

alishaevn (approved):

PR #536, opened 2022-05-25
Cache CsvParser#records method

bkiahstroud:
If we don't cache `CsvParser#records`, it does the following every time it gets called: 

- Reads all the data from the CSV 
- Saves the Importer object


bkiahstroud:
@sephirothkod are there any reasons we _shouldn't_ make this change? (Related thread [in Slack](https://assaydepot.slack.com/archives/G0311DN5MGA/p1653436936706459))

sephirothkod (approved):
This seems like a big oversight on our part. Should help with performance. Good catch Kiah!!

PR #537, opened 2022-05-26
Stop gt 255 filenames breaking export

cziaarm:
OK so even if Bulkrax is suggesting a filename that is guaranteed to be <=255 chars, zip uses tmpfile which uses tmpdir which uses prefixes and suffixes that will take it back over 255 chars. After looking at the tmpdir source code (!) my understanding is that it'll get `date-number-random_number_base64` appended to what we've supplied from bulkrax (or more importantly `8chars-1char-6chars`) so reducing the limiting we do in bulkrax (`[0...(255-ext.length)]`) by 17 would work... but let's do 35 for safety's sake...

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,7 @@ GIT
 PATH
   remote: .
   specs:
-    bulkrax (3.0.1)
+    bulkrax (3.1.0)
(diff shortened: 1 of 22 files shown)


cziaarm:
label: bug-fix please

Issue #444, opened 2022-03-22
Extend Exporter show page to include Collection and FileSet entries


bkiahstroud:

alishaevn:
released in https://github.com/samvera-labs/bulkrax/releases/tag/v4.3.0.

PR #542, opened 2022-06-01
444-exporter-show-page

alishaevn:
ref #444 

# expected behavior

# demo
| | exporter show page | 
| --- | --- |
| works |image |
| collections | image |
| file sets | image|


ShanaLMoore (approved):

ShanaLMoore:
The exporters show page would be easier to read if the tab sections were separated into partials. However I see that this is how the importers show page is done too.

alishaevn:
> @alishaevn The exporters show page would be easier to read if the tab sections were separated into partials. However I see that this is how the importers show page is done too.

great idea! I've updated this. can you review again @ShanaLMoore?

app/views/bulkrax/exporters/show.html.erb
```diff
@@ -87,49 +87,25 @@
...
-            <th>Identifier</th>
-            <th>Entry ID</th>
-            <th>Status</th>
-            <th>Errors</th>
-            <th>Status Set At</th>
-            <th>Actions</th>
-          </tr>
-        </thead>
-        <tbody>
-          <% @work_entries.each do |e| %>
```
alishaevn, line 109: moved the entry tabs into partials.

app/views/bulkrax/importers/show.html.erb
```diff
@@ -80,120 +80,9 @@
       </ul>
       <!-- Tab panes -->
       <div class="tab-content outline">
-        <div role="tabpanel" class="tab-pane active bulkrax-nav-tab-table-left-align" id="work-entries">
```
alishaevn, line 83: moved the entry tabs into partials.

ShanaLMoore: thanks for doing so, although both worked this looks so much better and less overwhelming to read :)

ShanaLMoore (approved):

PR #553, opened 2022-06-09
Pending relationships do not cause errors when destroying Importers

bkiahstroud:
# Summary 

Let Rails handle database table prefixes with its magic.

An Importer's `PendingRelationship`s (through `ImporterRun`) will no longer cause attempting to destroy the Importer to throw an error. Instead, the Importer's pending relationships will be destroyed alongside it automatically

### DIFF
--- a/app/models/bulkrax/importer_run.rb
+++ b/app/models/bulkrax/importer_run.rb
@@ -7,7 +7,7 @@ module Bulkrax
     has_many :pending_relationships, dependent: :destroy
 
     def parents
-      PendingRelationship.where(bulkrax_importer_run_id: id).pluck(:parent_id).uniq
+      pending_relationships.pluck(:parent_id).uniq
     end
   end
 end
--- a/app/models/bulkrax/importer_run.rb
+++ b/app/models/bulkrax/importer_run.rb
@@ -4,6 +4,7 @@ module Bulkrax
   class ImporterRun < ApplicationRecord
     belongs_to :importer
     has_many :statuses, as: :runnable, dependent: :destroy
+    has_many :pending_relationships, dependent: :destroy
 
     def parents
       PendingRelationship.where(bulkrax_importer_run_id: id).pluck(:parent_id).uniq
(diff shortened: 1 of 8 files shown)


bkiahstroud:
The failed "PR has required labels" CI check can be ignored in anticipation of [this PR](https://github.com/samvera-labs/bulkrax/pull/555)

app/models/bulkrax/importer_run.rb
```diff
@@ -4,9 +4,10 @@ module Bulkrax
   class ImporterRun < ApplicationRecord
     belongs_to :importer
     has_many :statuses, as: :runnable, dependent: :destroy
+    has_many :pending_relationships, dependent: :destroy
 
     def parents
-      PendingRelationship.where(bulkrax_importer_run_id: id).pluck(:parent_id).uniq
+      pending_relationships.pluck(:parent_id).uniq
```
alishaevn, line 10: will this not give us relationships unrelated to our current importer?

alishaevn: which could be problematic when the same work is imported on different importers. especially if the metadata is different. (I vaguely feel like this is a use case I've come up against)

bkiahstroud: Possibly? I'm not sure off the top of my head. Right now though, `PendingRelationship`s aren't associated directly with an importer, they're associated with the `ImporterRun`, so if anything the bug has been maintained by this change instead of introduced. I'm pretty sure that these two lines do the exact same thing (`PendingRelationship.where(bulkrax_importer_run_id: id) == pending_relationships`)

bkiahstroud: > will this not give us relationships unrelated to our current importer?

The more I think about this, the more confident I am that this isn't the case. The returned parent ids will only come from "this" imported run, which will only ever be associated with "this" importer

alishaevn: yeah, that's what I meant. the importer run is how it searches. which means you could have the same parent id across different importers. so you might want importer run 6, but end up with 2 because we're no longer adding that specificity right? unless I'm missing how `pending_relationships` is created?

alishaevn: we discussed the above over a call. 🙂

alishaevn (approved):

Issue #398, opened 2022-01-25
Better error handling for CSV::MalformedCSVErrors


bkiahstroud:
# Summary

When a `CSV::MalformedCSVError` is thrown, the importer just gets "stuck" in a "Pending" state with no clear reason why. This is because the `Bulkrax::ImporterJob` infinitely retries 

Example error (seen in Sidekiq retries dashboard): 

```
CSV::MalformedCSVError: Invalid byte sequence in UTF-8 in line 3.
```

Capture these errors and present a helpful message to the user

PR #558, opened 2022-06-10
#398 Handle malformed CSV errors better

bkiahstroud:
Closes https://github.com/samvera-labs/bulkrax/issues/398 

# Summary 

Do not infinitely retry importer job when CSV is malformed

### DIFF
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -96,16 +96,19 @@ module Bulkrax
     end
 
     def current_run
-      @current_run ||= if file? && zip?
-                         self.importer_runs.create!
-                       else
-                         entry_counts = {
-                           total_work_entries: self.limit || parser.works_total,
-                           total_collection_entries: parser.collections_total,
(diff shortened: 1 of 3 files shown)


app/models/bulkrax/importer.rb
```diff
@@ -96,16 +96,19 @@ def schedulable?
     end
 
     def current_run
-      @current_run ||= if file? && zip?
-                         self.importer_runs.create!
-                       else
-                         entry_counts = {
-                           total_work_entries: self.limit || parser.works_total,
-                           total_collection_entries: parser.collections_total,
-                           total_file_set_entries: parser.file_sets_total
```
bkiahstroud, line 105: Attempting to read the CSV throws the `CSV::MalformedCSVError`, which causes the `ImporterRun` creation to fail, which means that a status cannot be logged. This change adds an additional write to the db, which isn't ideal, but it ensures the `ImporterRun` will always be created successfully

alishaevn (approved):

PR #565, opened 2022-06-14
Remove willow_sword as a soft dependency, make rubyzip a full dependency

orangewolf:
Export is a first class feature, so we should support its dependencies directly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ coverage/**
 .DS_Store
 tmp/**
 .env.*
+*~undo-tree~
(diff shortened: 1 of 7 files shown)


kirkkwang:
Does this mean WillowSword can be removed from hyku as a dependency as well?

orangewolf:
@ShanaLMoore @kirkkwang any thoughts on why these specs are failing?

kirkkwang:
> @ShanaLMoore @kirkkwang any thoughts on why these specs are failing?

That's strange, I wonder if we need to add `require 'bagit'` to the spec file? I can look into it

Issue #566, opened 2022-06-14
expand collection metadata exporting


alishaevn:
# summary
currently we only export collection metadata when we select "all" or an importer. (related to: https://gitlab.com/notch8/palni-palci/-/issues/478)

# acceptance criteria
- [ ] we can export the metadata of all collections in their own csv (does not apply to the bagit parser)
- [ ] we can export collection metadata along with the works in that collection when we choose the "collection" option

alishaevn:
ref #597

Issue #563, opened 2022-06-13
import/export fileset metadata with bagit


alishaevn:
ref: https://gitlab.com/notch8/palni-palci/-/issues/478

# Acceptance Criteria
- [ ] a user should be able to export a work and its filesets + their metadata in bag format
- [ ] a user should be able to import a work and its filesets + their metadata in bag format

PR #567, opened 2022-06-14
566-collections-metadata-only

alishaevn:
ref: #566 

# expected behavior
- we can export the metadata of all collections in their own csv (does not apply to the bagit parser)
- we can export collection metadata along with the works in that collection when we choose the "collection" option (does not apply to the bagit parser)

# known gotchas
- this will not work for bagit parsing because bags require a file to be present and collections don't have files. this work is actually because of that fact though. exporting all work and fileset metadata is being worked on in #563 if a user wants checksum verification as they export their entire tenant. then they can export all collection metadata with this feature.
- the exporter show page only shows work ids, although the "total works" count is correct. this is being handled on #444 

# demo
|| new/edit page | show page | exported file |
| --- | --- | --- | --- |
| "collections metadata" | image |  image |[export__from_collections metadata 3.csv](https://github.com/samvera-labs/bulkrax/files/8903798/export__from_collections.metadata.3.csv)  |
| collections | image | image |  [export_3_7.zip](https://github.com/samvera-labs/bulkrax/files/8903814/export_3_7.zip) |

### DIFF
--- a/app/models/bulkrax/exporter.rb
+++ b/app/models/bulkrax/exporter.rb
@@ -14,13 +14,15 @@ module Bulkrax
     validates :name, presence: true
     validates :parser_klass, presence: true
 
-    delegate :write, :create_from_collection, :create_from_importer, :create_from_worktype, :create_from_all, to: :parser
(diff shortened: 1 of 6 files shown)


kirkkwang (approved):
Looks good to me!

PR #568, opened 2022-06-15
563-bagit-export-fileset-metadata

alishaevn:
ref #563 

# expected behavior
- exporting all now exports the file set metadata that's related to the work all on the same csv

# demo
[export_6_18.zip](https://github.com/samvera-labs/bulkrax/files/8905284/export_6_18.zip)

(the entire database consists of 2 collections, one work, with one fileset attached. we would expect one bag that has a work row and the related fileset row. no reference to the collections at all)
image


kirkkwang (approved):
Cool, looks good to me

PR #569, opened 2022-06-16
handles bagit bag file exists error and prepends id to data filenames

kirkkwang:
Prepending the id onto the filename to avoid duplication errors.  Though this changes the filename they are still coming out as valid bags so it should still satisfy requirements.  Added error handling just in case something really weird happens.

**Work with duplicate filesets**
Screen Shot 2022-06-15 at 10 09 23 PM

**Error in console**
Screen Shot 2022-06-15 at 8 05 36 PM

**Failed status on Export**
Screen Shot 2022-06-15 at 10 08 04 PM

**Failed status on Entry**
Screen Shot 2022-06-15 at 10 08 52 PM

**Resulting bag**
Screen Shot 2022-06-15 at 10 22 12 PM

### DIFF
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -189,7 +189,7 @@ module Bulkrax
     alias create_from_worktype create_new_entries
     alias create_from_all create_new_entries
 
-    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
     def write_files
       require 'open-uri'
       require 'socket'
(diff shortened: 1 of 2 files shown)


app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -211,17 +211,22 @@ def write_files
           file = Tempfile.new([file_name, File.extname(file_name)], binmode: true)
           file.write(io.read)
           file.close
-          bag.add_file(file_name, file.path)
+          begin
+            bag.add_file(file_name, file.path)
+          rescue => e
+            entry.status_info(e)
+            status_info(e)
+          end
```
alishaevn, line 219: 👍🏾

PR #571, opened 2022-06-16
563 import fileset metadata

alishaevn:
ref #563 

# expected behavior
- able to round trip objects when they have a value in the "children" and "parents" columns
- able to import a work, its child file sets and any parent collections each in their own row in a single bag
- handling nil when the parent or child record can't be found (the identifier passed in was an object.id so the entry couldn't be retrieved, and the object wasn't a collection)

# demo
### sample files
- [bag_with_all_relationships.zip](https://github.com/samvera-labs/bulkrax/files/8960969/bag_with_all_relationships.zip)
- [multiple_bags_with_fileset_rows.zip](https://github.com/samvera-labs/bulkrax/files/8960974/multiple_bags_with_fileset_rows.zip)
- [multiple_models_csv.zip](https://github.com/samvera-labs/bulkrax/files/8960973/multiple_models_csv.zip)
- [csv_with_all_relationships.zip](https://github.com/samvera-labs/bulkrax/files/8960976/csv_with_all_relationships.zip)

image

image


app/models/concerns/bulkrax/import_behavior.rb
```diff
@@ -12,8 +12,8 @@ def build_for_importer
           raise CollectionsCreatedError unless collections_created?
           @item = factory.run!
           add_user_to_permission_templates! if self.class.to_s.include?("Collection")
-          parent_jobs if self.parsed_metadata[related_parents_parsed_mapping].present?
-          child_jobs if self.parsed_metadata[related_children_parsed_mapping].present?
+          parent_jobs if self.parsed_metadata[related_parents_parsed_mapping]&.join.present?
+          child_jobs if self.parsed_metadata[related_children_parsed_mapping]&.join.present?
```
alishaevn, line 16: the join takes care of `['']`

app/models/concerns/bulkrax/dynamic_record_lookup.rb
```diff
@@ -12,15 +12,15 @@ def find_record(identifier, importer_run_id = nil)
...
-          record ||= work_type.where(id: identifier).first # rubocop:disable Rails/FindBy
-        end
+      begin
+        # the identifier parameter can be a :source_identifier or the id of an object
+        record = Entry.find_by(default_scope.merge({ importerexporter_id: importer_id })) || Entry.find_by(default_scope)
+        record ||= ActiveFedora::Base.find(identifier)
+      rescue NameError # if ActiveFedora isn't installed
+        record = nil
+      rescue ActiveFedora::ObjectNotFoundError
+        record = nil
```
sephirothkod, line 23:
```suggestion
      # NameError for if ActiveFedora isn't installed
      rescue NameError, ActiveFedora::ObjectNotFoundError
        record = nil
```

I believe you can do this so you don't need two rescues doing the same thing

alishaevn: ahh. I wasn't sure about chaining in rescue's. thanks!

app/parsers/bulkrax/application_parser.rb
```diff
@@ -104,20 +104,111 @@ def visibility
       @visibility ||= self.parser_fields['visibility'] || 'open'
     end
 
+    def build_records
```
alishaevn, line 107: moved this method here from csv_parser.rb

app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -36,39 +44,41 @@ def import_fields
...
+      else
+        data = entry_class.data_for_entry(data, source_identifier, self)
+        data[:file] = bag.bag_files.join('|') unless importerexporter.metadata_only?
       end
+
+      data
     end
 
     def create_works
+      entry_class == CsvEntry ? super : create_rdf_works
```
alishaevn, line 78: we want to be able to create more than works with a bagit csv.

app/parsers/bulkrax/oai_dc_parser.rb
```diff
@@ -61,6 +61,11 @@ def import_fields
 
     delegate :list_sets, to: :client
 
+    # TODO: not yet supported
```
alishaevn, line 64: the empty methods in this file are to override the now defaults in application_parser.rb

app/parsers/bulkrax/xml_parser.rb
```diff
@@ -12,11 +12,23 @@ def collection_entry_class; end
     # @todo not yet supported
     def create_collections; end
 
+    def create_file_sets; end
```
alishaevn, line 15: the empty methods in this file are to override (or set back) the now defaults in application_parser.rb.

app/models/concerns/bulkrax/file_set_entry_behavior.rb
```diff
@@ -8,10 +8,14 @@ def factory_class
 
     def add_path_to_file
       parsed_metadata['file'].each_with_index do |filename, i|
-        path_to_file = ::File.join(parser.path_to_files, filename)
+        next if filename.blank?
+
+        path_to_file = ::File.join(parser.path_to_files(parent_id: parsed_metadata['parents']), filename)
```
bkiahstroud, line 13: The "parents" key is dynamic / isn't guaranteed to be `"parents"`. [This method](https://github.com/samvera-labs/bulkrax/blob/main/app/parsers/bulkrax/application_parser.rb#L63-L65) fetches the dynamic key and falls back on `"parents"`

app/parsers/bulkrax/bagit_parser.rb
bkiahstroud: (Disclaimer: I know very little about bagit) -- if there's this much overlap between the bagit and csv parsers, should the bagit parser inherit from the csv parser..? 🤔  Instead of pushing "csv behavior" into the application parser?

app/parsers/bulkrax/oai_dc_parser.rb
bkiahstroud: Instead of "unsetting" defaults, should the bagit parser just inherit from the csv parser? (see [related comment](https://github.com/samvera-labs/bulkrax/pull/571#discussion_r902779855))

app/parsers/bulkrax/xml_parser.rb
bkiahstroud: Instead of "unsetting" defaults, should the bagit parser just inherit from the csv parser? (see [related comment](https://github.com/samvera-labs/bulkrax/pull/571#discussion_r902779855))

app/parsers/bulkrax/bagit_parser.rb
alishaevn: I thought about this too. bagit itself doesn't inherit so much from the csv parser, it's just the csv format of bagit that inherits from it. the rdf portion does it's own thing...which is why stuff like this exists.

although moving the csv only stuff back into the csv parser and having bagit inherit from it would mean that overrides only need to exist in the bagit parser. so that may be worth a shot.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -382,10 +382,11 @@ def file_paths
...
 
     # Retrieve the path where we expect to find the files
-    def path_to_files
+    def path_to_files(**args)
+      filename = args.fetch(:filename, '')
+
       @path_to_files ||= File.join(
-        zip? ? importer_unzip_path : File.dirname(import_file_path),
-        'files'
+        zip? ? importer_unzip_path : File.dirname(import_file_path), 'files', filename
```
bkiahstroud, line 389: @alishaevn If I'm reading this correctly, `filename` is effectively an optional argument. However, if `filename` ends up being an empty string, `File.join` will add a `/` to the end of the result: 

```ruby
irb(main):003:0> File.join('hello', 'world')
=> "hello/world"
irb(main):002:0> File.join('hello', 'world', '')
=> "hello/world/"
```

1. Do you know if this is problematic? Or does a trailing `/` not change the behavior? 
2. Was this change only made for bagit importers/exporters, or is the `filename` argument also used for CSV importer/exporters?

bkiahstroud: > 2. the change was made for the bagit parser because a user can import multiple bags in a single import. so we needed a way to determine the correct path. in doing so, I needed to update this version of the method in the csv_parser to ensure it still worked.

Why make the change in the csv parser and not the bagit parser then?

alishaevn: it was in the bagit parser, but now the bagit parser is inheriting from the csv parser. plus, we needed access to the file name in the `path_to_files` method in the bagit parser. which meant that result would already have the file name appended to it, but get it appended again in the previous version of the `add_path_to_file` method, which I also updated. 

it was either do some conditionals in the `add_path_to_file` method based on the parser, or just update the `path_to_files` methods that already existed in the csv and bagit parsers. I felt the latter was cleaner and more semantic.

bkiahstroud: Ahh, I think I get it now. Thanks!

bkiahstroud (approved):
I love PRs that remove more code than they add :shipit:

PR #572, opened 2022-06-16
Bagit export status and total fix

kirkkwang:
Bagit exports were erroneously updating the status to 'Complete' when it was not.  That was because the enqueued records were never set correctly and was stuck on zero.  Pulled in `CsvParser#total` to `BagitParser` and found when `limit` is 0, it did not respect it as the same as nil.

image

This change might need to be reflected in `CsvParser` or better yet, move all duplicate code from each individual parser up to `ApplicationParser`.

### DIFF
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -100,8 +100,17 @@ module Bulkrax
       total
     end
 
+    # TODO @kirkkwang: look into if CsvParser#total needs this same logic
     def total
-      importerexporter.entries.count
+      @total = importer.parser_fields['total'] || 0 if importer?
+
+      @total = if exporter?
+                 limit.nil? || limit.zero? ? current_record_ids.count : limit
+               end
+
+      return @total || 0
+    rescue StandardError
+      @total = 0
     end
 
     def extra_filters
@@ -132,7 +141,6 @@ module Bulkrax
       when 'importer'
         set_ids_for_exporting_from_importer
       end
-
       @work_ids + @collection_ids + @file_set_ids
     end
 
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -100,7 +100,6 @@ module Bulkrax
       total
     end
 
-    # TODO: look into if CsvParser#total needs this same logic
     def total
       @total = importer.parser_fields['total'] || 0 if importer?
 
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -189,7 +189,7 @@ module Bulkrax
     alias create_from_worktype create_new_entries
     alias create_from_all create_new_entries
 
-    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
     def write_files
       require 'open-uri'
       require 'socket'
@@ -226,7 +226,7 @@ module Bulkrax
         bag.manifest!(algo: 'sha256')
       end
     end
-    # rubocop:enable Metrics/AbcSize
+    # rubocop:enable Metrics/MethodLength, Metrics/AbcSize
 
     def setup_csv_metadata_export_file(id)
       File.join(importerexporter.exporter_export_path, id, 'metadata.csv')
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -193,11 +193,11 @@ module Bulkrax
     def write_files
       require 'open-uri'
       require 'socket'
-      importerexporter.entries.where(identifier: current_record_ids)[0..limit || total].each do |e|
-        work = ActiveFedora::Base.find(e.identifier)
+      importerexporter.entries.where(identifier: current_record_ids)[0..limit || total].each do |entry|
+        work = ActiveFedora::Base.find(entry.identifier)
         next unless Hyrax.config.curation_concerns.include?(work.class)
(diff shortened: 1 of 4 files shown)


app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -100,8 +100,17 @@ def works_total
       total
     end
 
+    # TODO: look into if CsvParser#total needs this same logic
```
alishaevn, line 103:
```suggestion
```

alishaevn: this is already the exact logic in the csv parser.

app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -189,15 +197,15 @@ def create_new_entries
     alias create_from_worktype create_new_entries
     alias create_from_all create_new_entries
 
-    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength, Metrics/AbcSize
```
alishaevn, line 200: the changes from lines 200-237 were done in https://github.com/samvera-labs/bulkrax/pull/569. if you merge in main to this branch, this diff should go away.

spec/parsers/bulkrax/bagit_parser_spec.rb
```diff
@@ -5,6 +5,34 @@
...
 module Bulkrax
   RSpec.describe BagitParser do
+    describe '#total' do
+      context 'while exporting' do
+        subject { described_class.new(exporter) }
+        let(:exporter) { FactoryBot.create(:bulkrax_exporter_worktype_bagit) }
+        let(:bulkrax_exporter_run) { FactoryBot.create(:bulkrax_exporter_run, exporter: exporter) }
+        let(:work_ids_solr) { [OpenStruct.new(id: SecureRandom.alphanumeric(9)), OpenStruct.new(id: SecureRandom.alphanumeric(9))] }
+        let(:collection_ids_solr) { [OpenStruct.new(id: SecureRandom.alphanumeric(9))] }
+        let(:file_set_ids_solr) { [OpenStruct.new(id: SecureRandom.alphanumeric(9)), OpenStruct.new(id: SecureRandom.alphanumeric(9)), OpenStruct.new(id: SecureRandom.alphanumeric(9))] }
```
alishaevn, line 15: this logic is already in the `describe '#create_new_entries'...` section of this file. can we just move it out so that it's usable more than once, instead of duplicating it?

spec/parsers/bulkrax/bagit_parser_spec.rb
```diff
@@ -5,6 +5,34 @@
...
+
+        before do
+          allow(subject).to receive(:current_record_ids).and_return(work_ids_solr + collection_ids_solr + file_set_ids_solr)
```
alishaevn, line 18: we don't export collections with the bagit parser. so the totals below would all need to be minus 1.

app/parsers/bulkrax/bagit_parser.rb
kirkkwang: This was copied over but had to be changed because if `limit` was 0, then `@total` would come out as 0, if `limit` was nil then `@total` would be whatever `current_record_ids.count` is.  The user chooses the limit, but it is indicated that 0 or blank _should_ do the same thing.  It is seemingly not, at least in the `BagitParser`.

app/parsers/bulkrax/bagit_parser.rb
kirkkwang: That sounds good i'll take a look

spec/parsers/bulkrax/bagit_parser_spec.rb
```diff
@@ -5,6 +5,34 @@
 
 module Bulkrax
   RSpec.describe BagitParser do
+    describe '#total' do
+      context 'while exporting' do
```
alishaevn, line 9: since we can export with various restrictions (all, importer, collection...) I think this context would do well with some greater specificity. 🙂

spec/parsers/bulkrax/bagit_parser_spec.rb
kirkkwang: Oh good catch

app/parsers/bulkrax/bagit_parser.rb
alishaevn: mmm. I actually wouldn't expect a limit of 0 and a "missing' limit to do the same. to me "0" does not equal "none". 

I would expect a limit to be enforced only if a value were placed in the "limit" field. even if that value were 0. in which case my logic is in line with the logic in the csv parser. 

perhaps we should get another opinion?

spec/parsers/bulkrax/bagit_parser_spec.rb
alishaevn: and the "collection_ids_solr" wouldn't need to be returned in this before block.

app/parsers/bulkrax/bagit_parser.rb
kirkkwang: Ah gotcha, I was going off of what the UI says, but I admit I haven't explicitly tested what if I put a 0 in there.
image

I can just take it out for now until further testing.

alishaevn: you know what.....idk if I ever looked at that message. 😅  that's not at all how I'd expect it to work. if it is then you're implementation is what should happen.

alishaevn (approved):

app/parsers/bulkrax/bagit_parser.rb
alishaevn: @kirkkwang I approved this pr...but we should still check with others what the intention here should be.

kirkkwang: Thanks @alishaevn and agreed!

Issue #538, opened 2022-05-27
Exporter From type field as a required field


kirkkwang:
As a user, I should not be able to create an export if the "Export From" source is not filled out.

On `/exporters/new`, when the `Export From` field is selected, a new field with the same name as the selected entry shows up.  That field does not have a required label nor a validation.  The user can create the export without choosing that option which results in an error.
| Current | Should be | And have |
|---|---|---|
|<img width="400" alt="image" src="https://user-images.githubusercontent.com/19597776/170733743-01a7e056-8c5e-4eb2-a751-4bdd49f16706.png">|<img width="400" alt="image" src="https://user-images.githubusercontent.com/19597776/170734152-6c65ba82-ffaa-4748-8e3a-ba4076de760d.png">|<img width="400" alt="image" src="https://user-images.githubusercontent.com/19597776/170734926-9b2947d2-e08a-40e5-b1b5-a54020fa0f35.png">

These changes should be applied to the `Importer`, `Collection`, and `Work Type` options.
NOTE: Screen shots were created by manipulating the DOM through the browser, work has not begun.

Testing instructions:

- [ ] Go to Dashboard > Exporters > New
- [ ] Fill out:
  - [ ] `Name`
  - [ ] `Type of Export`
  - [ ] `Export From`
  - [ ] `Export Format`
- [ ] Check that if `Importer`, `Collection`, or `Work Type` is chosen from the `Export From` dropdown, there is a `required` badge on the label
- [ ] Without selecting anything in `Work Type` (should be on 'Select from the list'), click `Create and Export`
- [ ] Confirm that the `Please select and item in the list.` alert pops up in the `Work Type` field

laritakr:
refer to https://stackoverflow.com/questions/22148080/an-invalid-form-control-with-name-is-not-focusable

solution needs to mimic what is being done in the importer re: hidden fields.

PR #580, opened 2022-06-27
Undo require form fields

laritakr:
References[ bulkrax 538](https://github.com/samvera-labs/bulkrax/issues/538). Hidden fields cannot be required.

reverse https://github.com/samvera-labs/bulkrax/pull/575 to be handled at a later point.

### DIFF
--- a/app/views/bulkrax/exporters/_form.html.erb
+++ b/app/views/bulkrax/exporters/_form.html.erb
@@ -29,7 +29,7 @@
 
   <%= form.input :export_source_importer,
     label: t('bulkrax.exporter.labels.importer'),
-    required: true,
+    # required: true,
     prompt: 'Select from the list',
     label_html: { class: 'importer export-source-option hidden' },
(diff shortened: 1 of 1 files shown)


alishaevn:
ref https://github.com/samvera-labs/bulkrax/issues/538

bkiahstroud:
Not saying this needs to happen in this PR, but I think a potential solution might be disabling the hidden fields. If the inputs are disabled, no params are sent, and I think the requirement is disabled

alishaevn:
> Not saying this needs to happen in this PR, but I think a potential solution might be disabling the hidden fields. If the inputs are disabled, no params are sent, and I think the requirement is disabled

thanks kiah! larita and selam did have an idea of how to solve this, but this is holding up the qa of 3 other tickets on pals so I advised to unblock our qa first, and then come back to trying to make this work. 🙂

bkiahstroud (approved):

PR #582, opened 2022-06-30
Split downloads by size

alishaevn:
ref: https://gitlab.com/notch8/palni-palci/-/issues/466

# expected behavior
### csv
- csv formatted exports will have a maximum of 1000 records per csv
- if there are multiple csv's, each one along with its related "files" folder, will be in its own zip file

### bagit
- bagit formatted exports will have a maximum of 1000 records per zip. 
  - there will still only be 1 work per csv, along with its child file sets
  - a single zip will contain multiple bags

### ui
- now that there are potentially several zip files created from a single export, there is a drop down of files on the index page
- selecting a zip file and then pressing "download" will download that particular zip file
- the same applies to the exporter show page

# demo
<details>
<summary> exporter index page </summary>

image
</details>

<details>
<summary> exporter show page </summary>

image
</details>

<details>
<summary> new file structure </summary>

image
</details>

### DIFF
--- a/app/controllers/bulkrax/exporters_controller.rb
+++ b/app/controllers/bulkrax/exporters_controller.rb
@@ -127,7 +127,7 @@ module Bulkrax
     # Download methods
 
     def file_path
-      @exporter.exporter_export_zip_path
+      "#{@exporter.exporter_export_zip_path}/#{params['exporter']['exporter_export_zip_files']}"
     end
   end
 end
--- a/app/models/concerns/bulkrax/export_behavior.rb
+++ b/app/models/concerns/bulkrax/export_behavior.rb
@@ -7,9 +7,6 @@ module Bulkrax
 
     def build_for_exporter
       build_export_metadata
-      # TODO(alishaevn): determine if the line below is still necessary
-      # the csv and bagit parsers also have write_files methods
-      write_files if export_type == 'full' && !importerexporter.parser_klass.include?('Bagit')
     rescue RSolr::Error::Http, CollectionsCreatedError => e
       raise e
     rescue StandardError => e
@@ -24,25 +21,6 @@ module Bulkrax
 
     def hyrax_record
       @hyrax_record ||= ActiveFedora::Base.find(self.identifier)
-    end
-
-    def write_files
-      return if hyrax_record.is_a?(Collection)
-
-      file_sets = hyrax_record.file_set? ? Array.wrap(hyrax_record) : hyrax_record.file_sets
-      file_sets << hyrax_record.thumbnail if hyrax_record.thumbnail.present? && hyrax_record.work? && exporter.include_thumbnails
-      file_sets.each do |fs|
-        path = File.join(exporter_export_path, 'files')
-        FileUtils.mkdir_p(path)
-        file = filename(fs)
-        require 'open-uri'
-        io = open(fs.original_file.uri)
-        next if file.blank?
-        File.open(File.join(path, file), 'wb') do |f|
-          f.write(io.read)
-          f.close
-        end
-      end
     end
 
     # Prepend the file_set id to ensure a unique filename and also one that is not longer than 255 characters
--- a/app/parsers/bulkrax/application_parser.rb
+++ b/app/parsers/bulkrax/application_parser.rb
@@ -262,10 +262,20 @@ module Bulkrax
     end
 
     def zip
-      FileUtils.rm_rf(exporter_export_zip_path)
-      Zip::File.open(exporter_export_zip_path, create: true) do |zip_file|
-        Dir["#{exporter_export_path}/**/**"].each do |file|
-          zip_file.add(file.sub("#{exporter_export_path}/", ''), file)
+      if exporter_export_zip_path.exclude?('zip')
+        Dir["#{exporter_export_path}/**"].each do |folder|
+          zip_path = "#{exporter_export_zip_path.split('/').last}_#{folder.split('/').last}.zip"
+          Zip::File.open(File.join("#{exporter_export_zip_path}/#{zip_path}"), create: true) do |zip_file|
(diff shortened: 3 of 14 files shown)


app/controllers/bulkrax/exporters_controller.rb
```diff
@@ -127,7 +127,7 @@ def add_exporter_breadcrumbs
     # Download methods
 
     def file_path
-      @exporter.exporter_export_zip_path
+      "#{@exporter.exporter_export_zip_path}/#{params['exporter']['exporter_export_zip_files']}"
```
alishaevn, line 130: this will return something like "tmp/exports/export_2_109/export_2_109_1.zip"

app/models/concerns/bulkrax/export_behavior.rb
```diff
@@ -26,25 +23,6 @@ def hyrax_record
       @hyrax_record ||= ActiveFedora::Base.find(self.identifier)
     end
 
-    def write_files
```
alishaevn, line 29: exporting the file sets is now done from within the `write_files` method in the csv_parser (same as the bagit parser). it was necessary because each zip will need it's own "files" folder, instead of adding them after the fact.

app/parsers/bulkrax/application_parser.rb
```diff
@@ -262,10 +262,14 @@ def unzip(file_to_unzip)
     end
 
     def zip
-      FileUtils.rm_rf(exporter_export_zip_path)
-      Zip::File.open(exporter_export_zip_path, create: true) do |zip_file|
-        Dir["#{exporter_export_path}/**/**"].each do |file|
-          zip_file.add(file.sub("#{exporter_export_path}/", ''), file)
+      FileUtils.mkdir_p(exporter_export_zip_path)
```
alishaevn, line 265: refer to the pr comment under "demo" for what this new file structure looks like.

app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -122,18 +123,27 @@ def current_record_ids
     def write_files
       require 'open-uri'
       require 'socket'
+
```
alishaevn, line 126: what the new file structure looks like when bags are being created for export
image

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -307,9 +308,33 @@ def retrieve_cloud_files(files)
     # export methods
 
     def write_files
-      CSV.open(setup_export_file, "w", headers: export_headers, write_headers: true) do |csv|
-        importerexporter.entries.where(identifier: current_record_ids)[0..limit || total].each do |e|
-          csv << e.parsed_metadata
+      require 'open-uri'
```
alishaevn, line 311: what the new file structure looks like when csv's are being created for export
image

app/views/bulkrax/exporters/show.html.erb
```diff
@@ -135,10 +136,6 @@
       <%= page_entries_info(@work_entries) %><br>
       <%= paginate(@work_entries, param_name: :work_entries_page) %>
       <br>
-      <% if File.exist?(@exporter.exporter_export_zip_path) %>
```
alishaevn, line 138: now that the download is a dropdown, it felt too clunky to keep at the bottom.

lib/tasks/bulkrax_tasks.rake
```diff
@@ -1,6 +1,30 @@
 # frozen_string_literal: true
 
-# desc "Explaining what the task does"
-# task :bulkrax do
-#   # Task goes here
-# end
+namespace :bulkrax do
```
alishaevn, line 3: this rake task needs to be run on all projects that update to this version!!

the exporter paths have all changed, so existing exported files can't be reached anymore.

spec/jobs/bulkrax/importer_job_spec.rb
```diff
@@ -30,7 +30,7 @@ module Bulkrax
         importer_job.perform(importer.id)
 
         expect(importer.current_run.total_work_entries).to eq(10)
-        expect(importer.current_run.total_collection_entries).to eq(427)
+        expect(importer.current_run.total_collection_entries).to eq(428)
```
kirkkwang, line 33: Curious how come this number ticked up by 1?

kirkkwang (commented):
When I run the CSV (I'm splitting it by 5 vs 1000 for testing purposes), this is what I'm getting.  There are 8 works that I'm exporting so I'm expecting two zips, one with five and another with three.  Instead I'm getting four zip files.

image

image

I then tried the same thing with BagIt instead and it's erroring out:

image

Is this related to the rake task?

I'm still seeing a zip available for download:

image

But when I click download I get:

image

Also, it's the only zip that shows up in the drop down menu, while I expected two.

image

bkiahstroud (commented):
1. The fact that this will break all previous exports makes me think it should be a major release, not a minor one. Can we update the PR labels for the Release Notes? 
2. Can we add test coverage for these changes? They're significant and doing a merge/release without them makes me nervous

spec/jobs/bulkrax/importer_job_spec.rb
bkiahstroud: Specs that reach out to external sources should always be stubbed. This one isn't, so the external count it's testing against changes sometimes. It needs to be properly stubbed, but no one has gotten around to doing it yet.

alishaevn:
> 1. The fact that this will break all previous exports makes me think it should be a major release, not a minor one. Can we update the PR labels for the Release Notes?
> 
>     2. Can we add test coverage for these changes? They're significant and doing a merge/release without them makes me nervous

@bkiahstroud added specs. mind taking another look? @sephirothkod you too?

sephirothkod (approved):

PR #584, opened 2022-07-05
Fix wiki links, remove CLA mention

rotated8:
A couple links to the Bulkrax wiki did not resolve, so I replaced them with what seemed like the most obvious candidates. Also, since Samvera has dropped the CLA requirement, I removed the mention of it I found.


bkiahstroud (approved):
Thanks for doing this!

PR #597, opened 2022-07-16
export-child-collections

alishaevn:
# expected behavior
- do not include child collections as part of the @work_ids
- include child collections as part of the @collection_ids
- remove remaining/returned references to "collections metadata"
- return guard clause for bagit files on export

# demo
- 1 parent collection
- 1 child collection
- 2 child works, with 6 file sets between them
(the bag image only shows one of the child works with its associated files)

csv: [export_2_19_1.zip](https://github.com/samvera-labs/bulkrax/files/9134433/export_2_19_1.zip)
bag: [export_3_21_1.zip](https://github.com/samvera-labs/bulkrax/files/9134394/export_3_21_1.zip)

| csv | bag|
| --- | --- |
| image | image |

### DIFF
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -140,7 +140,7 @@ module Bulkrax
           file.write(io.read)
           file.close
           begin
-            bag.add_file(file_name, file.path) unless bag.bag_files.select { |b| b.include?(file_name) }.present?
+            bag.add_file(file_name, file.path) if bag.bag_files.select { |b| b.include?(file_name) }.blank?
           rescue => e
             entry.status_info(e)
             status_info(e)
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -140,7 +140,7 @@ module Bulkrax
           file.write(io.read)
           file.close
           begin
-            bag.add_file(file_name, file.path)
+            bag.add_file(file_name, file.path) unless bag.bag_files.select { |b| b.include?(file_name) }.present?
           rescue => e
             entry.status_info(e)
             status_info(e)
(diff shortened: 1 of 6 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -195,8 +195,9 @@ def current_record_ids
         @collection_ids = ActiveFedora::SolrService.query("has_model_ssim:Collection #{extra_filters}", method: :post, rows: 2_147_483_647).map(&:id)
         @file_set_ids = ActiveFedora::SolrService.query("has_model_ssim:FileSet #{extra_filters}", method: :post, rows: 2_147_483_647).map(&:id)
       when 'collection'
-        @work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source + extra_filters}", method: :post, rows: 2_000_000_000).map(&:id)
-        @collection_ids = ActiveFedora::SolrService.query("id:#{importerexporter.export_source} #{extra_filters}", method: :post, rows: 2_147_483_647).map(&:id)
+        @work_ids = ActiveFedora::SolrService.query("member_of_collection_ids_ssim:#{importerexporter.export_source + extra_filters} AND has_model_ssim:(#{Hyrax.config.curation_concerns.join(' OR ')})", method: :post, rows: 2_000_000_000).map(&:id)
```
alishaevn, line 198: added `AND has_model_ssim:(#{Hyrax.config.curation_concerns.join(' OR ')})` so we stop returning child collections as a work.

app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -140,7 +140,7 @@ def write_files
           file.write(io.read)
           file.close
           begin
-            bag.add_file(file_name, file.path)
+            bag.add_file(file_name, file.path) if bag.bag_files.select { |b| b.include?(file_name) }.blank?
```
alishaevn, line 143: because of the work I was doing on https://github.com/samvera-labs/bulkrax/pull/596 on this tenant, I happened to import duplicate files. (known issue on that branch) that work is on hold and we won't come across it in live apps that are using bulkrax.

however, figured we'd put the guard back since it doesn't hurt. and won't throw off our dev testing in the meantime.

bkiahstroud: @alishaevn if this is a temporary guard clause, should we add a `TODO:` to remove it later?

alishaevn: we've put this clause up and removed it about 3 times over the last 2 months. since we keep coming across use cases for it...I wasn't sure if it should be temporary, or just leave it here since it doesn't mess anything up either.

each time we find a use case, we find a way around it too so it _could_ be temporary. and currently, I'm the only one who will have the problem unless someone else works on the linked pr above. but leaving it here isn't problematic either.

I'm fine either way.

bkiahstroud (approved):

PR #599, opened 2022-07-18
undefined-method-length-for-nil-class

alishaevn:
# issue
ran into this issue while running this [rake task](https://github.com/samvera-labs/bulkrax/blob/main/lib/tasks/bulkrax_tasks.rake) on pals for all exporters that were exporting from a collection after updating to `v4.1.0`

sentry error: https://sentry.notch8.com/sentry/pals/issues/142067/
image

# expected behavior
- account for when `importerexporter.field_mapping` is `nil` in the "get_field_mapping_hash_for" method

# demo
image

### DIFF
--- a/app/parsers/bulkrax/application_parser.rb
+++ b/app/parsers/bulkrax/application_parser.rb
@@ -75,10 +75,9 @@ module Bulkrax
     def get_field_mapping_hash_for(key)
       return instance_variable_get("@#{key}_hash") if instance_variable_get("@#{key}_hash").present?
 
-      mapping = importerexporter.field_mapping == [{}] ? {} : importerexporter.field_mapping
       instance_variable_set(
         "@#{key}_hash",
-        mapping&.with_indifferent_access&.select { |_, h| h.key?(key) }
+        mapping&.with_indifferent_access&.select { |_, h| h.key?(key) } || {}
       )
       raise StandardError, "more than one #{key} declared: #{instance_variable_get("@#{key}_hash").keys.join(', ')}" if instance_variable_get("@#{key}_hash").length > 1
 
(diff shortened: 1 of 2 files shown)


app/parsers/bulkrax/application_parser.rb
```diff
@@ -75,10 +75,9 @@ def related_children_parsed_mapping
     def get_field_mapping_hash_for(key)
       return instance_variable_get("@#{key}_hash") if instance_variable_get("@#{key}_hash").present?
 
-      mapping = importerexporter.field_mapping == [{}] ? {} : importerexporter.field_mapping
       instance_variable_set(
         "@#{key}_hash",
-        mapping&.with_indifferent_access&.select { |_, h| h.key?(key) }
```
alishaevn, line 81: because `Bulkrax::Exporter.find(2).field_mapping` was returning `nil`, line 81 returned `nil`. this resulted in an error on line 83 because `instance_variable_get("@#{key}_hash")` evaluated to `nil` and `nil.length` throws an error.

app/parsers/bulkrax/application_parser.rb
```diff
@@ -75,10 +75,9 @@ def related_children_parsed_mapping
...
+        importerexporter.field_mapping&.with_indifferent_access&.select { |_, h| h.key?(key) } || {}
```
alishaevn, line 80: either return the valid field mapping, or `{}` in all cases.

app/parsers/bulkrax/application_parser.rb
```diff
@@ -75,7 +75,7 @@ def related_children_parsed_mapping
     def get_field_mapping_hash_for(key)
       return instance_variable_get("@#{key}_hash") if instance_variable_get("@#{key}_hash").present?
 
-      mapping = importerexporter.field_mapping == [{}] ? {} : importerexporter.field_mapping
+      mapping = importerexporter.field_mapping.is_a?(Hash) ? importerexporter.field_mapping : {}
```
alishaevn, line 78: either return the valid field mapping, or {} in all cases.

sephirothkod (approved):

PR #600, opened 2022-07-19
 ignore-misidentified-collection-entries

alishaevn:
# issue
ran into this issue while running this [rake task](https://github.com/samvera-labs/bulkrax/blob/main/lib/tasks/bulkrax_tasks.rake) on pals for all exporters that were exporting from a collection after updating to `v4.1.0`

sentry error: https://sentry.notch8.com/sentry/pals/issues/142068/
image

# demo
before https://github.com/samvera-labs/bulkrax/pull/597, child collections were exported as a CsvEntry class instead of CsvCollectionEntry. this means that when that importer/exporter is run again, the old entry still exists, but a new one is created as well. the newer entry is at index 0, so calling `uniq` on the identifier value (as seen below), returns index 0.

Screen Shot 2022-07-18 at 8 37 59 PM

# expected behavior
- ignore misidentified collection entries when writing files from the csv parser

### DIFF
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -324,7 +324,7 @@ module Bulkrax
     def write_files
       require 'open-uri'
       folder_count = 0
-      sorted_entries = sort_entries(importerexporter.entries)
+      sorted_entries = sort_entries(importerexporter.entries.uniq(&:identifier))
 
       sorted_entries[0..limit || total].in_groups_of(records_split_count, false) do |group|
         folder_count += 1


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -324,7 +324,7 @@ def retrieve_cloud_files(files)
     def write_files
       require 'open-uri'
       folder_count = 0
-      sorted_entries = sort_entries(importerexporter.entries)
+      sorted_entries = sort_entries(importerexporter.entries.uniq(&:identifier))
```
alishaevn, line 327: before #597, child collections were exported as a `CsvEntry` class instead of `CsvCollectionEntry`. this means that when that importer/exporter is run again, the old entry still exists, but a new one is created as well. we want to get the newest one.

``` bash
irb(main):148:0> current_record_ids.count
=> 13

irb(main):149:0> Bulkrax::Exporter.find(59).entries.count
=> 14

Bulkrax::Exporter.find(59).entries.uniq(&:identifier).count
=> 13
```

sephirothkod (approved):

PR #604, opened 2022-07-19
escape-nil-comparisons

alishaevn:
# issue
sentry: https://sentry.notch8.com/sentry/pals/issues/142087/
image

# expected behavior
- account for custom work types in #sort_entries

### DIFF
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -391,12 +391,12 @@ module Bulkrax
       # always export models in the same order: work, collection, file set
       entries.sort_by do |entry|
         case entry.type
-        when 'Bulkrax::CsvEntry'
-          '0'
         when 'Bulkrax::CsvCollectionEntry'
           '1'
         when 'Bulkrax::CsvFileSetEntry'
           '2'
+        else
+          '0'
         end
       end
     end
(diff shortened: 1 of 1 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -391,12 +396,12 @@ def sort_entries(entries)
       # always export models in the same order: work, collection, file set
       entries.sort_by do |entry|
         case entry.type
-        when 'Bulkrax::CsvEntry'
-          '0'
         when 'Bulkrax::CsvCollectionEntry'
           '1'
         when 'Bulkrax::CsvFileSetEntry'
           '2'
+        else
```
alishaevn, line 403: in apps that have custom work entry types, they should all be counted as a "Bulkrax::CsvEntry". this is in line with what we're doing in `#create_new_entries` too.

app/parsers/bulkrax/csv_parser.rb
```diff
@@ -325,6 +329,7 @@ def write_files
       require 'open-uri'
       folder_count = 0
       sorted_entries = sort_entries(importerexporter.entries.uniq(&:identifier))
+                       .select { |e| valid_entry_types.include?(e.type) }
```
alishaevn, line 332: now that we're exporting all work types as "Bulkrax::CsvEntry", we don't need to reference the other custom work entry types that may have been previously exported. the valid work itself will still be exported because of the `#create_new_entries` method. but it will be only as a "Bulkrax::CsvEntry".

Issue #611, opened 2022-07-25
export base_path per tenant


alishaevn:
# summary
current: "tmp/exports/"
proposed: "tmp/exports/`<tenant-name>`"

the import paths aren't being updated because they do not appear to have this same issue. potentially because timestamps are being used??

# issue
- the exporter paths should be filtered per tenant so that data isn't written over and/or found on a tenant it shouldn't be
- example of the issue: https://gitlab.com/notch8/palni-palci/-/issues/466#note_1037575100

alishaevn:
released in https://github.com/samvera-labs/bulkrax/releases/tag/v4.3.0.

PR #612, opened 2022-07-26
611-base-path-per-tenant

alishaevn:
ref #611 

# expected behavior
- use the tenant name in the parser export base_path in hyku apps

# demo

| before | after |
| --- | --- |
| image | image|

# info
the spec for this work is being done in #613

### DIFF
--- a/app/parsers/bulkrax/application_parser.rb
+++ b/app/parsers/bulkrax/application_parser.rb
@@ -134,7 +134,9 @@ module Bulkrax
 
     # Base path for imported and exported files
     def base_path(type = 'import')
-      ENV['HYKU_MULTITENANT'] ? File.join(Bulkrax.send("#{type}_path"), Site.instance.account.name) : Bulkrax.send("#{type}_path")
+      # account for multiple versions of hyku
+      is_multitenant = ENV['HYKU_MULTITENANT'] || ENV['SETTINGS__MULTITENANCY__ENABLED'] == 'true'
+      is_multitenant ? File.join(Bulkrax.send("#{type}_path"), Site.instance.account.name) : Bulkrax.send("#{type}_path")
     end
 
     # Path where we'll store the import metadata and files
--- a/app/parsers/bulkrax/application_parser.rb
+++ b/app/parsers/bulkrax/application_parser.rb
@@ -134,9 +134,9 @@ module Bulkrax
 
     # Base path for imported and exported files
     def base_path(type = 'import')
-      # account for multiple versions of hyku
-      is_multitenant = ENV['HYKU_MULTITENANT'] || ENV['SETTINGS__MULTITENANCY__ENABLED'] == 'true'
-      is_multitenant ? File.join(Bulkrax.send("#{type}_path"), Site.instance.account.name) : Bulkrax.send("#{type}_path")
+      # account for multiple versions of hyku, but don't change the import paths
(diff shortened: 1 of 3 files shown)


lib/tasks/bulkrax_tasks.rake
```diff
@@ -5,26 +5,27 @@ namespace :bulkrax do
...
       end
     else
-      rerun_exporters_and_delete_zips
+      make_new_exports
     end
   end
 
-  def rerun_exporters_and_delete_zips
+  def make_new_exports
+    # delete the existing folders and zip files
```
alishaevn, line 22: deletes everything in "tmp/exports" and creates new paths.

left it in here even for hyrax apps because if the app is upgrading to 4.0+, it will still need the new file structure outlined in #582.

app/parsers/bulkrax/application_parser.rb
```diff
@@ -134,7 +134,9 @@ def write_import_file(file)
 
     # Base path for imported and exported files
     def base_path(type = 'import')
-      ENV['HYKU_MULTITENANT'] ? File.join(Bulkrax.send("#{type}_path"), Site.instance.account.name) : Bulkrax.send("#{type}_path")
```
alishaevn, line 137: this code was moved to the application parser in [this commit](https://github.com/samvera-labs/bulkrax/commit/65b2ef282aa546b29fa886c6ee23b2e84ba2b196#diff-fb85ab6a3ae748c61b9e6ca84ae5ab375b13470522978ecdd561b699967cfdf1), sept 2021.

unsure if it ever worked as intended though because `ENV['HYKU_MULTITENANT']` would have returned `'true'` or `'false'`. both of which would have evaluated to true since they're not empty strings.

app/parsers/bulkrax/application_parser.rb
```diff
@@ -134,7 +134,9 @@ def write_import_file(file)
...
+      # account for multiple versions of hyku, but don't change the import paths
+      is_multitenant = ENV['HYKU_MULTITENANT'] == 'true' || ENV['SETTINGS__MULTITENANCY__ENABLED'] == 'true'
```
alishaevn, line 138: pals is using `ENV['SETTINGS__MULTITENANCY__ENABLED']`

louisville, british libraries and utk are all using `ENV['HYKU_MULTITENANT']`. I believe the difference is the hyku version. so I'm accounting for both so multiple versions of hyku still work as expected.

app/parsers/bulkrax/application_parser.rb
kirkkwang: ran into the same issue in hyku, kept evaluating to true because it was actually `'false'` vs `false`

app/parsers/bulkrax/application_parser.rb
kirkkwang: What would happen if an application started as single tenant, then later decided to go multitenant?  Is that ever an edge to worry about?

alishaevn: the data would still exist in the db, they just wouldn't be able to access the path anymore.

for importing -- this shouldn't really be a problem. once the data is imported...I don't see a great need to access the imported file again.

for exporting -- re run the exporter, or create a new one.

kirkkwang (approved):

PR #613, opened 2022-07-26
611 base_path spec

alishaevn:
ref #612 

# expected behavior
- add specs for the #base_path method

### DIFF
--- a/spec/parsers/bulkrax/application_parser_spec.rb
+++ b/spec/parsers/bulkrax/application_parser_spec.rb
@@ -4,15 +4,15 @@ require 'rails_helper'
 
 module Bulkrax
   RSpec.describe ApplicationParser do
-    describe '#get_field_mapping_hash_for' do
-      let(:importer) { FactoryBot.create(:bulkrax_importer) }
-      let(:exporter_with_no_field_mapping) { FactoryBot.create(:bulkrax_exporter) }
-      let(:exporter_with_field_mapping) do
-        FactoryBot.create(:bulkrax_exporter, field_mapping: {
-                            "bulkrax_identifier" => { "from" => ["source_identifier"], "source_identifier" => true }
-                          })
-      end
+    let(:importer) { FactoryBot.create(:bulkrax_importer) }
+    let(:exporter_with_no_field_mapping) { FactoryBot.create(:bulkrax_exporter) }
+    let(:exporter_with_field_mapping) do
+      FactoryBot.create(:bulkrax_exporter, field_mapping: {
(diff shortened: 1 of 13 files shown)


spec/parsers/bulkrax/application_parser_spec.rb
```diff
@@ -37,5 +39,40 @@ module Bulkrax
...
+
+    describe '#base_path' do
+      before do
+        allow(Site).to receive(:instance).and_return(site)
+        allow(Site.instance).to receive(:account).and_return(account)
+      end
+
+      context 'in a hyku enabled app' do
+        before do
+          ENV['SETTINGS__MULTITENANCY__ENABLED'] = 'true'
```
kirkkwang, line 51: would it be worth it to do this test again but with `ENV['HYKU_MULTITENANT'] = 'true'`?

alishaevn: I thought about that...but it would end in the same result. plus to me the purpose of the test was to just make sure that if we're in a multi tenant app, that we get the right path and in a non multi tenant app, we get the right path. 
not to check per se all the ways we can get to those results.

kirkkwang: ah yes agreed then

kirkkwang (approved):

spec/parsers/bulkrax/application_parser_spec.rb
```diff
@@ -37,5 +39,40 @@ module Bulkrax
...
       end
     end
+
+    describe '#base_path' do
+      before do
+        allow(Site).to receive(:instance).and_return(site)
+        allow(Site.instance).to receive(:account).and_return(account)
+      end
+
+      context 'in a hyku enabled app' do
```
alishaevn, line 49:
```suggestion
      context 'in a multi tenant app' do
```

spec/parsers/bulkrax/application_parser_spec.rb
```diff
@@ -37,5 +39,40 @@ module Bulkrax
...
+        it 'sets the import path correctly' do
+          expect(importer.parser.base_path).to eq('tmp/imports/bulkrax')
+        end
+
+        it 'sets the export path correctly' do
+          expect(importer.parser.base_path('export')).to eq('tmp/exports/bulkrax')
+        end
+      end
+
+      context 'in a hyrax app' do
```
alishaevn, line 63:
```suggestion
      context 'in a non multi tenant app' do
        # this includes single tenant hyku apps

```

PR #620, opened 2022-07-27
move migration files to test_app/db

ShanaLMoore:
- [ ] move migrations to test_app/db to not affect the main migration stack.

### DIFF
--- a/spec/test_app/db/migrate/20220726162605_create_site.rb
+++ b/spec/test_app/db/migrate/20220726162605_create_site.rb
@@ -1,0 +1,7 @@
+class CreateSite < ActiveRecord::Migration[5.2]
+  def change
+    create_table :sites do |t|
+      t.integer :account_id
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


spec/test_app/db/migrate/20220726162605_create_site.rb
```diff
@@ -0,0 +1,7 @@
+class CreateSite < ActiveRecord::Migration[5.2]
```
alishaevn, line 1: don't we need to delete the version of this file that exists in db/migrate too?

spec/test_app/db/migrate/20220726164334_create_account.rb
```diff
@@ -0,0 +1,7 @@
+class CreateAccount < ActiveRecord::Migration[5.2]
```
alishaevn, line 1: same as above.

alishaevn (approved):

PR #621, opened 2022-07-27
account-for-missing-original-file-on-export

alishaevn:
# issue
this might be the fix to the issue with pals constantly retrying 2 exporters. per the screen shot below, `.uri` is getting called on a nil class. `.uri` only shows up 3 times in the code. the bagit parser, csv parser and object factory inside of the `#create_file_set` method. we aren't creating, so I figure that's not the problem.

I'd think we should always have a file...but we're doing a check for "if file_set.original_file.blank?" in the `#filename` method in export_behavior.rb too. so I figured there must be a case where it doesn't exist.

Screen Shot 2022-07-27 at 2 07 33 PM

# expected behavior
- account for a missing original_file on csv and bagit export

# demo
Screen Shot 2022-07-27 at 6 04 47 PM

### DIFF
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -134,7 +134,8 @@ module Bulkrax
 
         record.file_sets.each do |fs|
           file_name = filename(fs)
-          next if file_name.blank?
+          next if file_name.blank? || fs.original_file.blank?
+
           io = open(fs.original_file.uri)
           file = Tempfile.new([file_name, File.extname(file_name)], binmode: true)
           file.write(io.read)
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -200,13 +200,12 @@ module Bulkrax
         # get the parent collection and child collections
         @collection_ids = ActiveFedora::SolrService.query("id:#{importerexporter.export_source} #{extra_filters}", method: :post, rows: 2_147_483_647).map(&:id)
         @collection_ids += ActiveFedora::SolrService.query("has_model_ssim:Collection AND member_of_collection_ids_ssim:#{importerexporter.export_source}", method: :post, rows: 2_147_483_647).map(&:id)
+        find_child_file_sets(@work_ids)
       when 'worktype'
         @work_ids = ActiveFedora::SolrService.query("has_model_ssim:#{importerexporter.export_source + extra_filters}", method: :post, rows: 2_000_000_000).map(&:id)
       when 'importer'
         set_ids_for_exporting_from_importer
       end
-
-      find_child_file_sets(@work_ids) if importerexporter.export_from == 'collection'
 
       @work_ids + @collection_ids + @file_set_ids
     end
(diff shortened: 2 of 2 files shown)


app/parsers/bulkrax/bagit_parser.rb
```diff
@@ -134,7 +134,8 @@ def write_files
 
         record.file_sets.each do |fs|
           file_name = filename(fs)
-          next if file_name.blank?
+          next if file_name.blank? || fs.original_file.blank?
```
alishaevn, line 137: if we exit the loop here, the FileSet does not get written to the csv.

alishaevn:
@orangewolf @sephirothkod @bkiahstroud do either of you see any problems with the description under "issue" above?

idk the record that caused the issue because sidekiq doesn't give any details. I'm running the exporter through the rancher console now, but so far I'm getting no errors. this pr is more of using inferences and trying to figure out the cause of the error. 

have any of you encountered a case where a file set doesn't have the `original_file` on it?

alishaevn:
related [sentry error](https://sentry.notch8.com/sentry/pals/issues/142420/) and [slack thread](https://assaydepot.slack.com/archives/G0311DNF3MG/p1658964889983179).

app/parsers/bulkrax/bagit_parser.rb
kirkkwang: When this happens and the FileSet does not get written to the CSV, then because of this line https://github.com/samvera-labs/bulkrax/blob/f81734b00f9e7aa91301ac39e23a748986b5d252/app/parsers/bulkrax/bagit_parser.rb#L127
it leads me to think it still would be counted resulting in a split that's instead of 1000 records, something like 999 records.

Would a `records_in_folder - 1` or something to that extend be appropriate here?

kirkkwang: maybe something like?

```ruby
if file_name.blank? || fs.original_file.blank?
  records_in_folder - 1
  next
end
```

alishaevn: great suggestion! thanks!

kirkkwang: Yeah, i was looking at that... didn't know how to get around that 😓

alishaevn: I considered weeding out the bad files after the `#current_record_ids` method, and before the `#write_files` methods. that would require finding each file set and then removing the id if `original_file` doesn't exist though.

in this case where we are exporting the entire tenant, there are over 2000 files. plus `current_record_ids` isn't an instance variable and we call it a few times so it could result in different values. I don't think the operational cost of removing a few missing files so that we can have 1000 records exactly (or however many they deem once that value becomes dynamic) on a csv is worth it.

kirkkwang: Agreed, I gave it some thought as well and I don't think it'll be worth it either.

kirkkwang (approved):

alishaevn:
I dug some more into this issue (it's recorded in the slack thread linked above if you want details).

I found a file where `original_file.blank?` returned true. the file set exists, but there is no image attached to it. this feels like a valid error to check for.

Issue #635, opened 2022-08-17
update the default "split" value


alishaevn:
# summary
now that allinson_flex can be used with bulkrax, we should consider changing the default `split` value to "true" in [default_field_mapping](https://github.com/samvera-labs/bulkrax/blob/main/lib/bulkrax.rb#L89).

since allinson_flex is supposed to handle adding/removing property's on the fly, we don't need to define those properties in "config/initializers/bulkrax.rb" in each client app. however, many users import a multi valued property using a split character and not with numerated headers.

this change would allow for that continued behavior for users using allinson_flex and bulkrax together. it should also not mess up existing behavior

alishaevn:
@orangewolf @sephirothkod @bkiahstroud @ShanaLMoore @kirkkwang opinions?

alishaevn:
hmmm. yeah those would be useful too. thanks!

kirkkwang:
in UTK there's a MR that checks for AllinsonFlex profiles, maybe something like that and set it to true if conditions are met?

https://gitlab.com/notch8/utk-hyku/-/merge_requests/83/diffs#3ebc2864d890372340663389de3c9fe45fc07fe2_0_9

alishaevn:
> in UTK there's a MR that checks for AllinsonFlex profiles, maybe something like that and set it to true if conditions are met?
> 

that could be a possibility, but I'm thinking more of changing the `split` value to be more accessible/dynamic overall....not just when AF is installed. having AF installed however is one of the use cases that makes a more dynamic approach necessary.

PR #636, opened 2022-08-17
635 update default split value

alishaevn:
# expected behavior
- the default split value is now "true" do to reasons listed [in the issue](https://github.com/samvera-labs/bulkrax/issues/635)
- split will default to a pipe ("|") only because the semi-colon (";") is a valid character in the `remote_url` header. also removed the colon (":") as an option because I think it could mess up in the same place.

this is still a draft because I haven't tested this code, but these are the changes that felt right to me by just looking at the code.

### DIFF
--- a/app/matchers/bulkrax/application_matcher.rb
+++ b/app/matchers/bulkrax/application_matcher.rb
@@ -30,7 +30,7 @@ module Bulkrax
 
     def process_split
       if self.split.is_a?(TrueClass)
-        @result = @result.split(/\s*[|]\s*/) # default split by "|""
+        @result = @result.split(/\s*[|]\s*/) # default split by "|"
       elsif self.split
         result = @result.split(Regexp.new(self.split))
         @result = result.map(&:strip)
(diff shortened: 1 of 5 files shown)


app/matchers/bulkrax/application_matcher.rb
```diff
@@ -30,7 +30,7 @@ def result(_parser, content)
 
     def process_split
       if self.split.is_a?(TrueClass)
-        @result = @result.split(/\s*[:;|]\s*/) # default split by : ; |
+        @result = @result.split(/\s*[|]\s*/) # default split by "|""
```
alishaevn, line 33:
```suggestion
        @result = @result.split(/\s*[|]\s*/) # default split by "|"
```

spec/matchers/bulkrax/application_matcher_spec.rb
```diff
@@ -22,12 +22,6 @@ module Bulkrax
         result = matcher.result(nil, " hey ; how : are | you")
         expect(result).to eq("hey ; how : are | you")
       end
-
-      it 'custom split' do
-        matcher = described_class.new(split: '\|')
-        result = matcher.result(nil, " hey ; how : are | you")
-        expect(result).to eq(["hey ; how : are", "you"])
-      end
```
alishaevn, line 30: this is now the same as `split: true` which is tested above.

alishaevn:
closing this for the moment rather than leaving it as a draft. the direction on handling this may change (refer to the ticket).

PR #643, opened 2022-08-24
i-638-remote-files-import-with-FileSet

alishaevn:
# expected behavior
- we should be able to import a FileSet row with a file using the `remote_files` header

# demo
sample file: [daily_bugle.csv](https://github.com/samvera-labs/bulkrax/files/9421496/daily_bugle.csv)
image


app/models/concerns/bulkrax/file_set_entry_behavior.rb
```diff
@@ -6,7 +6,17 @@ def factory_class
       ::FileSet
     end
 
+    def file_reference
+      if parsed_metadata&.[]('file')&.map(&:present?)&.any?
+        return 'file'
+      elsif parsed_metadata&.[]('remote_files')&.map(&:present?)&.any?
```
alishaevn, line 12: also account for `parsed_metadata['remote_files']`.

app/models/concerns/bulkrax/file_set_entry_behavior.rb
```diff
@@ -6,7 +6,17 @@ def factory_class
...
+        return 'remote_files'
+      end
+    end
+
     def add_path_to_file
+      return unless file_reference == 'file'
```
alishaevn, line 18: we don't have a local file when using `remote_url`.

ShanaLMoore (approved):

PR #645, opened 2022-08-25
I 638 remote files import with file set (part 2)

alishaevn:
# fixes the incorrect merge conflict from #643


kirkkwang (approved):
looks good!

Issue #638, opened 2022-08-24
import FileSet row with `remote_files` header


alishaevn:
ref: https://gitlab.com/notch8/utk-hyku/-/issues/153

# Summary
bulkrax is not currently set up to allow files to uploaded using the `remote_files` header when also using the FileSet row with metadata. bulkrax will need to be updated to handle this for the dc tenant.

ref: [the bulkrax wiki](https://github.com/samvera-labs/bulkrax/wiki/CSV-Importer#method-2)

# Acceptance Criteria
- [ ] import a FileSet as a model with metadata and a file, using the `remote_files` header, successfully

# Screenshots or Video
![image](/uploads/a96527afa2e562389765ec8e8aa76461/image.png)

# Testing Instructions
- import the file below through bulkrax successfully
  - [daily_bugle.csv](/uploads/0ef6f36b6646b63486c8554c5f8dc7a2/daily_bugle.csv)

# Notes
- one piece of code that would need to be updated: https://github.com/samvera-labs/bulkrax/blob/main/app/models/concerns/bulkrax/file_set_entry_behavior.rb#L24-L28

alishaevn:
closed the pr above since we have not returned to this work. if this ticket is picked up again, the pr can be reopened if desired.

PR #651, opened 2022-09-07
i638-create-content-for-remote-files

alishaevn:
ref: https://gitlab.com/notch8/utk-hyku/-/issues/173#note_1090318824

# expected behavior
- call `#create_content` in the file set actor method for remote files too

# demo
image

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -174,11 +174,12 @@ module Bulkrax
 
     def create_file_set_actor(attrs, work, work_permissions, file, type:)
       actor = ::Hyrax::Actors::FileSetActor.new(object, @user)
-      file&.update(file_set_uri: actor.file_set.uri) if type = 'uploaded'
+      # TODO(alishaevn): confirm that the line below doesn't need to be applied to remote files too
+      file&.update(file_set_uri: actor.file_set.uri) if type == 'uploaded'
       actor.file_set.permissions_attributes = work_permissions
       actor.create_metadata(attrs)
       actor.create_content(file)
-      handle_remote_file(remote_file: remote_file, actor: actor, update: false) if type = 'remote'
+      handle_remote_file(remote_file: remote_file, actor: actor, update: false) if type == 'remote'
       actor.attach_to_work(work, attrs)
     end
 
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -163,22 +163,22 @@ module Bulkrax
         uploaded_file = ::Hyrax::UploadedFile.find(uploaded_file_id)
         next if uploaded_file.file_set_uri.present?
 
-        create_file_set_actor(attrs, work, work_permissions, uploaded_file)
+        create_file_set_actor(attrs, work, work_permissions, uploaded_file, type: 'uploaded')
       end
       attrs['remote_files']&.each do |remote_file|
-        create_file_set_actor(attrs, work, work_permissions, nil, remote_file)
+        create_file_set_actor(attrs, work, work_permissions, remote_file, type: 'remote')
       end
 
       object.save!
     end
 
-    def create_file_set_actor(attrs, work, work_permissions, uploaded_file, remote_file = nil)
+    def create_file_set_actor(attrs, work, work_permissions, file, type:)
       actor = ::Hyrax::Actors::FileSetActor.new(object, @user)
-      uploaded_file&.update(file_set_uri: actor.file_set.uri)
+      file&.update(file_set_uri: actor.file_set.uri) if type = 'uploaded'
       actor.file_set.permissions_attributes = work_permissions
       actor.create_metadata(attrs)
-      actor.create_content(uploaded_file) if uploaded_file
-      handle_remote_file(remote_file: remote_file, actor: actor, update: false) if remote_file
+      actor.create_content(file)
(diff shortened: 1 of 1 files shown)


app/factories/bulkrax/object_factory.rb
```diff
@@ -163,22 +163,22 @@ def create_file_set(attrs)
...
     end
 
-    def create_file_set_actor(attrs, work, work_permissions, uploaded_file, remote_file = nil)
+    def create_file_set_actor(attrs, work, work_permissions, file, type:)
       actor = ::Hyrax::Actors::FileSetActor.new(object, @user)
-      uploaded_file&.update(file_set_uri: actor.file_set.uri)
+      file&.update(file_set_uri: actor.file_set.uri) if type = 'uploaded'
       actor.file_set.permissions_attributes = work_permissions
       actor.create_metadata(attrs)
-      actor.create_content(uploaded_file) if uploaded_file
```
alishaevn, line 180: this line caused the error below because `parent` came back as `nil` in [this line ](https://github.com/samvera/hyrax/blob/92686adc219ddcf0ec9e1deece47723c4e7c7718/app/actors/hyrax/actors/file_set_actor.rb#L33)before being passed to the VisibilityCopyJob.

Screen Shot 2022-09-06 at 6 18 50 PM

app/factories/bulkrax/object_factory.rb
```diff
@@ -163,22 +163,22 @@ def create_file_set(attrs)
...
-      handle_remote_file(remote_file: remote_file, actor: actor, update: false) if remote_file
+      actor.create_content(file)
```
alishaevn, line 180: always pass the file to `#create_content`

app/factories/bulkrax/object_factory.rb
```diff
@@ -163,22 +163,22 @@ def create_file_set(attrs)
         uploaded_file = ::Hyrax::UploadedFile.find(uploaded_file_id)
         next if uploaded_file.file_set_uri.present?
 
-        create_file_set_actor(attrs, work, work_permissions, uploaded_file)
+        create_file_set_actor(attrs, work, work_permissions, uploaded_file, type: 'uploaded')
       end
       attrs['remote_files']&.each do |remote_file|
-        create_file_set_actor(attrs, work, work_permissions, nil, remote_file)
+        create_file_set_actor(attrs, work, work_permissions, remote_file, type: 'remote')
```
alishaevn, line 169: passing whatever file type we're looping over and handling it as `file` in `#create_file_set_actor` below. also passing ta `type` parameter that we can  check against instead of setting "uploaded_file" or "remote_file" to nil.

this also sets us up for the possible future case that we also handle the `file_url` header on a FileSet row.

ShanaLMoore:
@alishaevn this looks like it's mainly a refactor. Just making sure you're aware that create_content is getting called by remote files here. With the refactor you may want to remove or change [this line](https://github.com/samvera-labs/bulkrax/blob/f137db7d5d92487ffd527d001d03f775433f6bf6/app/factories/bulkrax/object_factory.rb#L208)

alishaevn:
> @alishaevn this looks like it's mainly a refactor. Just making sure you're aware that create_content is getting called by remote files here. With the refactor you may want to remove or change [this line](https://github.com/samvera-labs/bulkrax/blob/f137db7d5d92487ffd527d001d03f775433f6bf6/app/factories/bulkrax/object_factory.rb#L208)

hmm. for some reason it wasn't working for me last night when I was trying to test importing a a csv with only a file set that would be related to an already existing work. however, right now my pry's and other things aren't playing nice so I'll make this a draft until I can come back and look at it further.

ShanaLMoore:
> > @alishaevn this looks like it's mainly a refactor. Just making sure you're aware that create_content is getting called by remote files here. With the refactor you may want to remove or change [this line](https://github.com/samvera-labs/bulkrax/blob/f137db7d5d92487ffd527d001d03f775433f6bf6/app/factories/bulkrax/object_factory.rb#L208)
> 
> hmm. for some reason it wasn't working for me last night when I was trying to test importing a a csv with only a file set that would be related to an already existing work. however, right now my pry's and other things aren't playing nice so I'll make this a draft until I can come back and look at it further.

Was it because you were trying to import a single line (FileSet only)? I actually didn't solve for that at the time but I think I recall trying it once before and it said it needed a parent and didn't import. 

Although a single fileset line should ref the parent work (which I'm pretty sure I had) so ya, something to look into. Either way, if it matters, the case that I solved for before had two rows (parent work and FileSet). 

[cats-2.csv](https://github.com/samvera-labs/bulkrax/files/9529585/cats-2.csv)
[update-cats-2.csv](https://github.com/samvera-labs/bulkrax/files/9529584/update-cats-2.csv)

alishaevn:
> Was it because you were importing a single line (FileSet only)? I actually didn't solve for that at the time but I think I recall trying it once before and it said it needed a parent and didn't import.
> ...

@ShanaLMoore https://gitlab.com/notch8/utk-hyku/-/issues/173#note_1090318824 << that's the issue I was trying to solve for. (I've now included it in this initial pr comment for better referencing)

tldr; the parent row wouldn't import, but what I was actually trying to test was the metadata on the fileset. so I switched up the test to only import a fileset, but attach it to an existing work.

alishaevn:
closing the pr for now since we have not returned to this work. if https://github.com/samvera-labs/bulkrax/issues/638 is picked up again, the pr can be reopened if desired.

PR #652, opened 2022-09-08
Adding Bulkrax config for multi-value split/join delimiters

### LINKED ISSUE #429: Change hard-coded `split` characters
### Depends on #213 -- that ticket should be done first 

--- 

# Summary

Change places in the code where `split` characters are hard-coded to use the dynamic values brought in by #213. 

# Acceptance Criteria 

- `Bulkrax::CsvEntry` 
  - [ ] `#possible_collection_ids` 
- `Bulkrax::CsvParser` 
  - [ ] `#file_paths` 
  - [ ] `#unique_collection_identifier` 
  - [ ] `#collections` 
- `Bulkrax::BagitParser` 
  - [ ] `#collections` 
- `Bulkrax::ApplicationMatcher` 
  - [ ] `#process_split`

jeremyf:
Prior to this commit, we had "magic" strings/regexs sprinkled throughout the code.  This change now moves that towards a consistent and configurable interface.

There are two nuances:

1. The collections title split now includes a new default (e.g. `;`) in the split.  That may introduce a bug.
2. This preserves the ability to specify `split: true` and use the default; which creates the least invasive change for downstream adopters.

Closes #429, #213

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -81,7 +81,7 @@ module Bulkrax
     def add_file
       self.parsed_metadata['file'] ||= []
       if record['file']&.is_a?(String)
-        self.parsed_metadata['file'] = record['file'].split(/\s*[;|]\s*/)
+        self.parsed_metadata['file'] = record['file'].split(Bulkrax.multi_value_element_split_on)
       elsif record['file'].is_a?(Array)
         self.parsed_metadata['file'] = record['file']
       end
@@ -176,7 +176,7 @@ module Bulkrax
       data = hyrax_record.send(key.to_s)
       if data.is_a?(ActiveTriples::Relation)
         if value['join']
-          self.parsed_metadata[key_for_export(key)] = data.map { |d| prepare_export_data(d) }.join(' | ').to_s # TODO: make split char dynamic
+          self.parsed_metadata[key_for_export(key)] = data.map { |d| prepare_export_data(d) }.join(Bulkrax.multi_value_element_join_on).to_s
         else
           data.each_with_index do |d, i|
             self.parsed_metadata["#{key_for_export(key)}_#{i + 1}"] = prepare_export_data(d)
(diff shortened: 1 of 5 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -81,7 +87,7 @@ def add_ingested_metadata
     def add_file
       self.parsed_metadata['file'] ||= []
       if record['file']&.is_a?(String)
-        self.parsed_metadata['file'] = record['file'].split(/\s*[;|]\s*/)
+        self.parsed_metadata['file'] = record['file'].split(MULTI_VALUE_SPLIT_DELIMITER_REGEXP)
```
alishaevn, line 90: looks like this change is related to #429. although eventually we want this value to be set by the user in their "config/bulkrax.rb" file, and then we pull that value to set the split value.

after this pr is merged, can you link it to that issue using the "development" section in either the pr or issue as described [here](https://github.com/samvera-labs/bulkrax/blob/main/CONTRIBUTING.md#submitting-changes)? (linking the issue after the pr is merged won't close the issue)

README.md
```diff
@@ -178,6 +178,15 @@ To delete an importer or exporter, select the delete (x) icon.
...
 Once your the exporter has run, a download icon will apear on the exporters menu page.
 
+## Testing
+Bulkrax comes with an extensive test suite; you can run the test suite by:
+
+- Cloning this repository
+- Changing into that directory (e.g. `cd /path/to/bulkrax`)
+- Then running the `rake` command
+
+The first time running the tests you may need to run `bin/rails db:migrate RAILS_ENV=test` to ensure that test database is up to date.
```
alishaevn, line 188: info about the test suite is in the [contributing doc](https://github.com/samvera-labs/bulkrax/blob/main/CONTRIBUTING.md#running-the-spec-suite). I like having mention of testing in the readme, but what do you think about pointing users there instead so that the information is in one place? less chance for divergence and keeps it DRY.

alishaevn (changes requested):

README.md
```diff
@@ -178,6 +178,15 @@ To delete an importer or exporter, select the delete (x) icon.
...
+
```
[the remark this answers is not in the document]
jeremyf: Agreed that this should dry up; I missed the contributing document but will go there.

PR #653, opened 2022-09-08
Naming arguments and reducing duplication in import_work_job

jeremyf:
Instead of re-finding the ImporterRun, store that run in a local variable.  Also instead of relying on the opaque arg order, name the args and use those (instead of `arg[0]` and `arg[1]`).

### DIFF
--- a/app/jobs/bulkrax/import_work_job.rb
+++ b/app/jobs/bulkrax/import_work_job.rb
@@ -5,25 +5,27 @@ module Bulkrax
     queue_as :import
 
     # rubocop:disable Rails/SkipsModelValidations
-    def perform(*args)
-      entry = Entry.find(args[0])
+    def perform(entry_id, run_id, *)
+      entry = Entry.find(entry_id)
+      importer_run = ImporterRun.find(run_id)
       entry.build
       if entry.status == "Complete"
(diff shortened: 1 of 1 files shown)


app/jobs/bulkrax/import_work_job.rb
```diff
@@ -5,25 +5,27 @@ class ImportWorkJob < ApplicationJob
...
         # do not retry here because whatever parse error kept you from creating a work will likely
         # keep preventing you from doing so.
-        ImporterRun.find(args[1]).increment!(:failed_records)
-        ImporterRun.find(args[1]).increment!(:failed_works)
-        ImporterRun.find(args[1]).decrement!(:enqueued_records) unless ImporterRun.find(args[1]).enqueued_records <= 0 # rubocop:disable Style/IdenticalConditionalBranches
+        importer_run.increment!(:failed_records)
+        importer_run.increment!(:failed_works)
       end
+      # Regardless of completion or not, we want to decrement the enqueued records.
+      importer_run.decrement!(:enqueued_records) unless importer_run.enqueued_records <= 0
```
alishaevn, line 22: 👍🏾

alishaevn (approved):

PR #658, opened 2022-10-20
No-op refactoring ImportWork/CollectionJob

jeremyf:
As part of reviewing how we move from `Bulkrax::ImporterJob#perform` into the Hyrax stack, I wanted to understand how both the `Bulkrax::ImportCollectionJob` and `Bulkrax::ImportWorkJob` "worked".

This refactor removes the magic positional arguments and names them. Further it adds documentation regarding the parameters.  And last it creates a local variable instead of relying on multiple calls to `ActiveRecord::Base.find`, which while cached, are unnecessary.

### DIFF
--- a/app/jobs/bulkrax/import_collection_job.rb
+++ b/app/jobs/bulkrax/import_collection_job.rb
@@ -5,21 +5,26 @@ module Bulkrax
     queue_as :import
 
     # rubocop:disable Rails/SkipsModelValidations
-    def perform(*args)
-      entry = Entry.find(args[0])
+    # @param entry_id [Object] the key for the Bulkrax::Entry to build and save.
+    # @param run_id [Object] the key for the Bulkrax::ImporterRun in which we're running this entry.
+    #
+    # @return [String]
+    def perform(entry_id, run_id, *)
+      entry = Entry.find(entry_id)
+      importer_run = ImporterRun.find(run_id)
       begin
         entry.build
         entry.save!
-        ImporterRun.find(args[1]).increment!(:processed_records)
-        ImporterRun.find(args[1]).increment!(:processed_collections)
-        ImporterRun.find(args[1]).decrement!(:enqueued_records) unless ImporterRun.find(args[1]).enqueued_records <= 0 # rubocop:disable Style/IdenticalConditionalBranches
+        importer_run.increment!(:processed_records)
+        importer_run.increment!(:processed_collections)
+        importer_run.decrement!(:enqueued_records) unless importer_run.enqueued_records <= 0 # rubocop:disable Style/IdenticalConditionalBranches
       rescue => e
-        ImporterRun.find(args[1]).increment!(:failed_records)
-        ImporterRun.find(args[1]).increment!(:failed_collections)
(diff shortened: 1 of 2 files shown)


app/jobs/bulkrax/import_collection_job.rb
```diff
@@ -5,21 +5,26 @@ class ImportCollectionJob < ApplicationJob
...
-        ImporterRun.find(args[1]).increment!(:processed_records)
-        ImporterRun.find(args[1]).increment!(:processed_collections)
-        ImporterRun.find(args[1]).decrement!(:enqueued_records) unless ImporterRun.find(args[1]).enqueued_records <= 0 # rubocop:disable Style/IdenticalConditionalBranches
+        importer_run.increment!(:processed_records)
+        importer_run.increment!(:processed_collections)
+        importer_run.decrement!(:enqueued_records) unless importer_run.enqueued_records <= 0 # rubocop:disable Style/IdenticalConditionalBranches
       rescue => e
-        ImporterRun.find(args[1]).increment!(:failed_records)
-        ImporterRun.find(args[1]).increment!(:failed_collections)
-        ImporterRun.find(args[1]).decrement!(:enqueued_records) unless ImporterRun.find(args[1]).enqueued_records <= 0 # rubocop:disable Style/IdenticalConditionalBranches
```
bkiahstroud, line 19: @jeremyf I know previously we had done this because it was the only way we could get the counts to be accurate. I think due to the async nature of imports, the counters would always end up being off by 1 or more from reality. Accurate counters has been an ongoing challenge from the beginning.

bkiahstroud: @orangewolf might be able to add more details, I think he's grappled with this more than anyone else thus far.

bkiahstroud: Because of this, if we do merge this, I think we should at least do a patch release.

jeremyf: I am happy to back off the refactor of a local variable, but my understanding (which could be faulty) is that we'd still be hitting the active record caching of that object within the "request" cycle.

bkiahstroud: >but my understanding (which could be faulty) is that we'd still be hitting the active record caching of that object within the "request" cycle.

That makes sense to me, but I'm also not positive. Honestly I'd like to rework the counters to be more reliable at some point. For now though, problems with the counters are so difficult to find and nail down that I get nervous. 

@orangewolf and/or @sephirothkod, do either of you have thoughts?

sephirothkod: Yes @bkiahstroud  you are correct we have had issues with inaccurate counts when using a local variable rather than calling the run each time. I would advise against this change.

jeremyf:
Given the past experience regarding inaccurate counts, I'm going to close this PR.

orangewolf:
the code comments are good. We had race condition issues with the counts. However, it's the increment! that really makes the difference there. I'm not sure fetching each time is 
a) happening (see active record cache)
b) needed

this could probably be determined via experimentation. Jeremy, I think you can see you hit a bit of a callous where we've worn on this issue before.

PR #662, opened 2022-10-27
Adding Bulkrax::RemoveRelationshipsForImporter

### LINKED ISSUE #661: In testing and verifying environments we need a way to clean-up relationships
As folks explore how to create parsers and mappings for their local Bulkrax implementations, in our experience (Software Services by Scientists.com) they need a way to clean-up relationships.

To resolve this ticket, we should add the following:

- A job to clean up all of tenant's relationships
- A job that clean's up one importer's relationship

These could initially called via the command line.

jeremyf:
See the inline documentation for details

Closes #661

### DIFF
--- a/app/services/bulkrax/remove_relationships_for_importer.rb
+++ b/app/services/bulkrax/remove_relationships_for_importer.rb
@@ -1,0 +1,107 @@
+# frozen_string_literal: true
+module Bulkrax
+  # This module is rather destructive; it will break relationships between the works, file sets, and
+  # collections that were imported via an importer.  You probably don't want to run this on your
+  # data, except in the case where you have been testing a Bulkrax::Importer, the parsers and
+  # mappings.  Then, you might have relationships that you want to remove.
+  #
+  # tl;dr - Caution this will break things!
+  class RemoveRelationshipsForImporter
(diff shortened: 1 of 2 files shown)


app/services/bulkrax/remove_relationships_from_importer.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+module Bulkrax
+  # This module is rather destructive; it will break relationships between the works, file sets, and
```
bkiahstroud, line 3: Technically, this doesn't break relationships for file sets (see line 55): 

```ruby
next if obj.is_a?(FileSet) # FileSets must be attached to a Work
```

Whether we _should_ break file set relationships, I'm open to discussion on.

bkiahstroud (commented):
@jeremyf one comment. Also, it looks like RSpec is failing to run in the CI for a name error.

app/services/bulkrax/remove_relationships_from_importer.rb
jeremyf: I am amending the inline docs to remove the FileSet reference.

PR #663, opened 2022-11-02
Refactoring to remove conditional

jeremyf:
Prior to this commit, we had an if statement that would check if the given parser was a `Bulkrax::CsvParser`, if so it would go down one logic pathway.  If not, it would call a different pathway.  Each pathway required knowledge of the implementation details of the parser.  The `Bulkrax::CsvParser` required knowledge of a non-public API setup (e.g. a method not defined on `Bulkrax::ApplicationParser`).

With this commit, we're favoring a single entry point for all parsers. Moving the logic down to the parser and allowing for polymorphism of the `Bulkrax::CsvParser` to provide it's different implementation.

One additional change is calling save on the importer before we create the objects.  This was something not called in the `Bulkrax::CsvParser` and based on the comment could have created a problem.

### DIFF
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -143,17 +143,13 @@ module Bulkrax
       import_objects(['relationship'])
     end
 
+    DEFAULT_OBJECT_TYPES = %w[collection work file_set relationship].freeze
+
     def import_objects(types_array = nil)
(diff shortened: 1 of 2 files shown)


alishaevn (approved):
thanks for the explanation in the initial comment on this pr!

PR #664, opened 2022-11-02
Merge default field mapping in with field mapping

ShanaLMoore:
A bug was found with the following scenario:

A client only had the necessary field mappings set (bulkrax_identifier, children, and parents). 

Because of this, [#mapping](https://github.com/samvera-labs/bulkrax/blob/main/app/models/bulkrax/importer.rb#L60-L72) would only return the field mapping. This caused an issue because [matcher](https://github.com/samvera-labs/bulkrax/blob/e88d0ea349afda5b3068836aa36ef86c93d4c62d/app/models/concerns/bulkrax/has_matchers.rb#L32) would return nil for all other imported values.

Because [it](https://github.com/samvera-labs/bulkrax/blob/e88d0ea349afda5b3068836aa36ef86c93d4c62d/app/models/concerns/bulkrax/has_matchers.rb#L44) was nil at the time of assigning the value, if the property was singular or multiple it would never process the split method to make [Subject A ; Subject B] into [Subject A, Subject B], for example.

This MR merges the default field mapping in with the field mapping, so that by default the splits would still be honored. It solves a bug found in bulkrax, but it also works for allinson flex's dynamic properties.

This was also tested using an importer with headers that didn't exist on any of the work types. Although it can be seen in the raw_metadata, parsed_metadata properly filters it out because #field_supported? returns false.

**NOTE**: in the following example donkey is an invalid field and should not be in the parsed metadata. And the client's default field mapping sets split to be a pipe. 

Screen Shot 2022-11-02 at 12 31 11 PM
Screen Shot 2022-11-02 at 12 32 24 PM

### DIFF
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -66,7 +66,7 @@ module Bulkrax
                    else
                      default_field_mapping.merge(self.field_mapping)
                    end
-                   
+
       # rubocop:enable Style/IfUnlessModifier
     end
 
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -63,7 +63,7 @@ module Bulkrax
                       default_field_mapping
                      end
                    else
-                     self.field_mapping.merge(default_field_mapping)
+                     default_field_mapping.merge(self.field_mapping)
                    end
     end
 
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -58,16 +58,21 @@ module Bulkrax
 
     # If field_mapping is empty, setup a default based on the export_properties
     def mapping
+      # rubocop:disable Style/IfUnlessModifier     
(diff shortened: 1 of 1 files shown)


app/models/bulkrax/importer.rb
```diff
@@ -60,17 +60,21 @@ def record_status
...
-                       ActiveSupport::HashWithIndifferentAccess.new(
-                         parser.import_fields.reject(&:nil?).map do |m|
-                           Bulkrax.default_field_mapping.call(m)
-                         end.inject(:merge)
-                       )
+                      default_field_mapping
                      end
                    else
-                     self.field_mapping
+                     self.field_mapping.merge(default_field_mapping)
```
alishaevn, line 66:
```suggestion
                     default_field_mapping.merge(self.field_mapping)
```

alishaevn: I think the reverse is needed so that the client's mappings override the defaults.

alishaevn (approved):

Issue #667, opened 2022-11-03
Bulkrax Overrides Hyrax 3.0 Sidebar Partials


no-reply:
Hyrax supports inserting custom partials to the sidebar with `Hyrax::DashboardController.sidebar_partials[:repository_content] << 'my/partial'`. However, bulkrax overrides the partial which displays these with a hard coded version. Any application which inserts custom partials will see them disappear when Bulkrax is installed.

it looks like there was an attempt to align with Hyrax 3 in 7009686f18e0025b12d5f47c8542e32375a456ab, but this doesn't seem to have had any impact, since the partial override at https://github.com/samvera-labs/bulkrax/blob/main/app/views/hyrax/dashboard/sidebar/_repository_content.html.erb still prevents the `sidebar_partials` configuration from being used.

laritakr:
Closed/completed by $#668

PR #668, opened 2022-11-03
Support customized sidebar partials if Hyrax 3.0 config is present

no-reply:
proposing this as a possible fix for #667


laritakr:
Allinson Flex has issues when used with Bulkrax. I wonder if we need to do something in Allinson Flex as well? I am not familiar with how the customized sidebar partials work, but I know that bulkrax was causing the Allinson Flex option to not appear.

app/views/hyrax/dashboard/sidebar/_repository_content.html.erb
```diff
@@ -1,19 +1,27 @@
...
+    <span class="fa fa-folder-open" aria-hidden="true"></span> <span class="sidebar-action-text"><%= t('hyrax.admin.sidebar.collections') %></span>
+  <% end %>
 
-<%= menu.nav_link(hyrax.my_works_path,
-                  also_active_for: hyrax.dashboard_works_path) do %>
-  <span class="fa fa-file" aria-hidden="true"></span> <span class="sidebar-action-text"><%= t('hyrax.admin.sidebar.works') %></span>
-<% end %>
+  <%= menu.nav_link(hyrax.my_works_path,
+                    class: "nav-link",
+        nn            onclick: "dontChangeAccordion(event);",
```
no-reply, line 12:
```suggestion
                    onclick: "dontChangeAccordion(event);",
```

no-reply:
@laritakr i think to really solve this problem, both engines need to use the Hyrax config and stop overriding the view partials.

do you have any ideas about how to move this forward?

laritakr:
> 

I really don't know the best way to do this, but I agree that I would like to NOT be overriding the partials. For Allinson Flex, I temporarily put in code to insert the menu option we need during installation if the partial already exists in the app. 

I hope that we can find a longer term solution because I think it's likely that both will continue to be used together. I am open to hearing what you think that solution should be.

no-reply:
well, again, i think the long term solution is to change both engines to use the configuration built into Hyrax. i think that requires that both engines drop support for Hyrax 2.x and remove their overrides.

a shorter term solution might be for both engines to merge something like this, which uses the configuration if it's present. or, alternatively, either or both engines could introduce the configuration if it's not present, and then force its use?

laritakr:
I'm not familiar with the configuration in Hyrax (I have not worked on Hyrax in quite some time) but I think we do want to move to that. Dropping support for Hyrax 2.x in Allinson Flex is probably not an issue, but I believe it would be for Bulkrax, so at least in Bulkrax, your shorter term solution would be the way to go.

jeremyf:
This looks like a reasonable add and something from which we can triangulate AllinsonFlex work.

jeremyf (approved):

PR #671, opened 2022-11-16
Fix `Hyrax::VisibilityPropagator` error for remote files

kirkkwang:
In UTK, we were getting a `Hyrax::VisibilityPropagator::NullVisibilityPropogator` error for remote files that was triggered by a misspelling.  Conveniently, this led to the question of why is it even hitting the `NullVisibilityPropogator` in the first place?  Turns out, the `FileSet` that comes in did not have a parent because of the logic in Bulkrax.

This PR swaps the logic so the `FileSet` should have a parent prior to hitting the `Hyrax::VisibilityPropagator`

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -178,8 +178,8 @@ module Bulkrax
       actor.file_set.permissions_attributes = work_permissions
       actor.create_metadata(attrs)
       actor.create_content(uploaded_file) if uploaded_file
-      handle_remote_file(remote_file: remote_file, actor: actor, update: false) if remote_file
       actor.attach_to_work(work, attrs)
+      handle_remote_file(remote_file: remote_file, actor: actor, update: false) if remote_file
     end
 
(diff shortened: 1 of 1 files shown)


laritakr (approved):

ShanaLMoore:
As part of this test, did you confirm that remote filesets still work? @kirkkwang

kirkkwang:
@ShanaLMoore 

I used this CSV:
[remote_file.csv](https://github.com/samvera-labs/bulkrax/files/10024646/remote_file.csv)

And this was the result.  The visibility of `private` matches on both the `Attachment` and its `FileSet`.
image

Is that what you're referring to?

ShanaLMoore:
@kirkkwang Did this also resolve the log_profile issue?

kirkkwang:
@ShanaLMoore 

I haven't seen that error pop up when i was doing my tests, so perhaps it did fix it?  Unsure though.

laritakr:
It should resolve it, because depositor was also nil because of no parent work going into the visibility job.

PR #674, opened 2022-11-19
importers/exporters links to not show up if bulkrax is not enabled

kirkkwang:
Since Bulkrax comes default with Hyku, if someone wants to not use Bulkrax, via `HYKU_BULKRAX_ENABLED=false` environment variable for example, the sidebar will break since it won't know what `bulkrax` is.  Putting this conditional before it will take care of that and remove the `Importers` and `Exporters` links from the sidebar.

image


st1072:
Hi @kirkkwang, I noticed that this change also means that `HYKU_BULKRAX_ENABLED` must be set to true for **Hyrax** to render the links. Maybe this template could also accept the condition that `HYKU_BULKRAX_ENABLED` is not set in Hyrax? Using [include?](https://ruby-doc.org/core-2.7.0/ENV.html#method-c-include-3F) might work.

PR #678, opened 2022-11-23
change the way we lookup collection ids

sephirothkod:
updates the collection lookup to be the same for both lookups in this method

### DIFF
--- a/app/models/bulkrax/oai_entry.rb
+++ b/app/models/bulkrax/oai_entry.rb
@@ -59,12 +59,10 @@ module Bulkrax
     def find_collection_ids
       return self.collection_ids if collections_created?
       if sets.blank? || parser.collection_name != 'all'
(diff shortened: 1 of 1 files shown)


app/models/bulkrax/oai_entry.rb
```diff
@@ -64,7 +64,8 @@ def find_collection_ids
         self.collection_ids << collection.id if collection.present? && !self.collection_ids.include?(collection.id)
       else # All - collections should exist for all sets
         sets.each do |set|
-          c = Collection.find_by(work_identifier => importerexporter.unique_collection_identifier(set.content))
+          # c = Collection.find_by(work_identifier => importerexporter.unique_collection_identifier(set.content))
```
bkiahstroud, line 67: Can we remove this if we don't need it?

sephirothkod: ya I have removed it now. don't know why I kept it in in the first place.

bkiahstroud (approved):

Issue #688, opened 2022-12-15
Handle invisible characters are included in CSV column names


jeremyf:
In one project we encountered a column name having a Byte Order Mark character; hidden and vexing.

In debugging the data we had 

```ruby
> ent = Bulkrax::CsvEntry.find(id)
> ent.record
=> {"file"=>"annie-spratt-ogYRV8cfsBo-unsplash.jpg; annie-spratt-example-of-second-scan.jpg", "identifier"=>"RML_MS009_001", "identifier.ark"=>"RML_MS009_001", "title"=>"Rena and Juanita Liu seated with their dolls", "description"=>"Rena Liu and Juanita Liu, seated, holding dolls, Eleanor and Evelyn.", "creator"=>"Frost, Samuel Lilley (1884-1981)", "contributor"=>"Frost, Ella Knokey (1887-1968)", "date"=>"1939", "date.other"=>"1939", "format.extent"=>"Photograph: black and white; 7 x 5 cm", "type"=>"Image", "subject"=>"Missions -- China", "language"=>"English", "source"=>"General Conference Office of Archives, Statistics, & Research", "relation.isPartOf"=>"Samuel, Ella, and Gladys Frost Photo Collection, 1910-1975", "rights"=>"http://rightsstatements.org/vocab/NoC-US/1.0/", "coverage.spatial"=>nil}
```

Notice the `"file"` key.  When I tried `ent.record['file']`, I got `nil`.  Looking deeper at the record's `file` key I tried the following:

```ruby
> ent.raw_metadata.keys.first.chars.map(&:chr)
=> ["", "f", "i", "l", "e"]
```

There was a hidden prefix character.  Upon further digging this was a https://en.wikipedia.org/wiki/Byte_order_mark, likely injected by the tool that generated the CSV.

PR #689, opened 2022-12-15
Normalizing serialized data for BOM characters

jeremyf:
Prior to this commit, when someone would upload a CSV it might have a column name that included the tricksy, invisibile [Byte Order Mark][1]. The manifestation was that you could look at the raw metadata and see a key called `"file"` however when checking `raw_metadata.key?("file")` the result was `false`.

To find this `entry.raw_metadata.keys.first.chars.map(&:chr)` I got `["", "f", "i", "l", "e"]` where the first value of that array was a [Byte Oder Mark][1].

With this commit, we're handling both how we persist and how we load persisted serialized data.  This follows on the documentation for the `ActiveRecord::Base.serialize` method.

It's envisioned that we might have more characters to sanitize.

Closes: samvera-labs/bulkrax#688
Related to: scientist-softserv/adventist-dl#179

[1]: https://en.wikipedia.org/wiki/Byte_order_mark

### DIFF
--- a/lib/bulkrax.rb
+++ b/lib/bulkrax.rb
@@ -195,13 +195,14 @@ module Bulkrax
       returning_value
     end
 
-    # Serializes an attribute value to a string that will be stored in the database.
+    # When we write the serialized data to the database, we "dump" the value into that database
+    # column.
     def self.dump(value)
       JSON.dump(normalize_keys(value))
     end
 
-    # Deserializes a string from the database to an attribute value.
-    # rubocop:disable Security/JSONLoad
+    # When we load the serialized data from the database, we pass the database's value into "load"
+    # function.
     def self.load(string)
       normalize_keys(JSON.load(string))
     end
--- a/lib/bulkrax.rb
+++ b/lib/bulkrax.rb
@@ -4,6 +4,7 @@ require "bulkrax/version"
 require "bulkrax/engine"
 require 'active_support/all'
 
+# rubocop:disable Metrics/ModuleLength
 module Bulkrax
   class << self
     mattr_accessor :api_definition,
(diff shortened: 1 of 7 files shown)


lib/bulkrax.rb
```diff
@@ -168,4 +169,43 @@ def self.multi_value_element_split_on
...
+      hash.each_pair do |key, value|
+        returning_value[Bulkrax.normalize_string(key)] = value
+      end
+      returning_value
+    end
+
+    # Serializes an attribute value to a string that will be stored in the database.
+    def self.dump(value)
+      JSON.dump(normalize_keys(value))
+    end
```
alishaevn, line 201: 1. this comment leads me to believe we're passing in a value only, and not the entire property as a key/value paired hash. is that correct? if so, wouldn't we immediately return the value from `normalize_keys` because of line 190?
2. where is this called from? or is this setup for the future?

lib/bulkrax.rb
```diff
@@ -168,4 +169,43 @@ def self.multi_value_element_split_on
...
+
+    # Deserializes a string from the database to an attribute value.
+    # rubocop:disable Security/JSONLoad
+    def self.load(string)
+      normalize_keys(JSON.load(string))
+    end
+    # rubocop:enable Security/JSONLoad
+  end
```
alishaevn, line 209: 1. similar to the above, is this a string that contains an object?
2. also, where is this called from?

lib/bulkrax.rb
jeremyf: The Rail's [ActiveRecord::Base.serialize](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize) takes one or more arguments.

The first argument is the symbol for the attribute we are serializing (e.g. `:raw_metadata`).  If we provide a second parameter that is the `class_name_or_coder`.  In this case we are wanting to use the `Bulkrax::NormalizedJson`.  The `.dump` method is what we use to coerce input before writing that attribute to the database.  The `.load` is what we use when reading from the database.

Previously we had `serialize :raw_metadata, JSON` which loaded and dumped the input by first converting it to JSON.  With this switch, we're using `serialize :raw_metadata, Bulkrax::NormalizedJson` which will continue to use JSON as the serialization format in the database but will also ensure that the keys don't have a Byte Order Mark in them.

jeremyf: To the second question, there is not future setup.

lib/bulkrax.rb
jeremyf: See above.

lib/bulkrax.rb
alishaevn: ahhhhh. knowing that `value` = `:raw_metadata` makes sense now. thanks!

alishaevn (approved):

Issue #610, opened 2022-07-22
add limits to all job rescheduling


alishaevn:
# issue
some jobs get rescheduled forever because the requirements aren't met. we should have a max amount of times that a job gets rescheduled for all jobs. once it reaches that limit, update the status of the entry and do not reschedule

## list of jobs
- [ ] CreateRelationshipsJob 
- [ ] ScheduleRelationshipsJob 
- [ ] ImportWorkJob

## example
- https://github.com/samvera-labs/bulkrax/blob/0919acf8757ef8e549a02a7a2a635a33c6db5ab4/app/jobs/bulkrax/import_file_set_job.rb#L29-L41

PR #692, opened 2022-12-19
Removing ∞ silent retries of Bulkrax::ImportWorkJob

### LINKED ISSUE #691: Bulkrax::ImportWorkJob silently and infinitely reschedules `Bulkrax::CollectionsCreatedError` exceptions

jeremyf:
Prior to this commit, when we encountered a
`Bulkrax::CollectionsCreatedError`, we would reschedule the same job with a 1 minute delay.  However, if the underlying issue was never resolved, we would continually reschedule without fanfare this failing job.

With this commit, we provide 3 logged reschedules, then if that fails we raise an exception break the infinite cycle.  By breaking the cycle, then the upstream Sidekiq handler can determine if it should be thrown into the `Retry` queue (where the jobs that raise exceptions go).

We discovered this in Adventist watch 100 or scheduled jobs run very fast, and apparently continually rescheduling themselves.

Closes #691

### DIFF
--- a/app/jobs/bulkrax/import_work_job.rb
+++ b/app/jobs/bulkrax/import_work_job.rb
@@ -5,7 +5,7 @@ module Bulkrax
     queue_as :import
 
     # rubocop:disable Rails/SkipsModelValidations
-    def perform(entry_id, run_id, *)
+    def perform(entry_id, run_id, time_to_live = 3, *)
       entry = Entry.find(entry_id)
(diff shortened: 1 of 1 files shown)


bkiahstroud (approved):

alishaevn:
as info, #610 was created for this purpose. at a later time we should handle this type of thing the same way throughout the entire app.

PR #697, opened 2022-12-22
Moving OAI metadata capture to entry processing

jeremyf:
In writing #694 I introduced an OAI bug.  There are two OAI fetch cycles.

The first is the list of records.  These are fetched via the `list_identifiers` method.  That method returns header information without metadata.  Then, we submit one job per header.  At that point we perform the second fetch of the record.  This includes the full metadata.


kirkkwang (approved):
Cool, glad you found it and applied the fix!

PR #698, opened 2023-01-05
add errored entries tab, count, and download

kirkkwang:


orangewolf:
@kirkkwang can this PR be closed?

Issue #699, opened 2023-01-10
Exporter of type "importer" fails with solrizer error in hyrax 3+


bbpennel:
Attempting to use an Exporter with "Export From" set to "importer" is failing for us in a Hyrax 3 repository with the following error:

```
Error: NameError - uninitialized constant Solrizer Did you mean? SolrHit

Error Trace:

ruby/2.7.0/gems/bulkrax-4.4.0/app/parsers/bulkrax/csv_parser.rb:237:in `block in set_ids_for_exporting_from_importer'
ruby/2.7.0/gems/bulkrax-4.4.0/app/parsers/bulkrax/csv_parser.rb:233:in `each'
ruby/2.7.0/gems/bulkrax-4.4.0/app/parsers/bulkrax/csv_parser.rb:233:in `set_ids_for_exporting_from_importer'
ruby/2.7.0/gems/bulkrax-4.4.0/app/parsers/bulkrax/csv_parser.rb:208:in `current_record_ids'
ruby/2.7.0/gems/bulkrax-4.4.0/app/parsers/bulkrax/csv_parser.rb:247:in `create_new_entries'
ruby/2.7.0/gems/bulkrax-4.4.0/app/models/bulkrax/exporter.rb:17:in `create_from_importer'
ruby/2.7.0/gems/bulkrax-4.4.0/app/models/bulkrax/exporter.rb:25:in `export'
ruby/2.7.0/gems/bulkrax-4.4.0/app/jobs/bulkrax/exporter_job.rb:9:in `perform'
```

This appears to be caused by the removal of Solrizer in hyrax 3, which is still used here:
https://github.com/samvera-labs/bulkrax/blob/main/app/parsers/bulkrax/csv_parser.rb#L237

The recommended remediation solr_name usage (according to slack) is to either directly use the name of the solr field, or to use
`#{::ActiveFedora.index_field_mapper.solr_name(work_identifier)}`

The latter might be the right approach, but it can fail is `source_identifier` is set to a field that does not need its name converted, like `id`, which gets expanded to `id_tesim` in my testing.

bkiahstroud:
Resolved by #700

PR #700, opened 2023-01-10
Replace solrizer reference so that feature works with hyrax 3+

bbpennel:
Addresses:
https://github.com/samvera-labs/bulkrax/issues/699

This change might not work with hyrax < 2.8, since it looks like solrizer got merged into activefedora in 12.0.0, so if we need to support those versions of hyrax we will likely need to add some logic in here to decide whether to use solrizer or activefedora.

I also suspect this code might not be getting fully tested, since bulkrax is using hyrax 3, but no tests seem to be impacted by this issue.

### DIFF
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -234,13 +234,17 @@ module Bulkrax
         instance_variable_set(instance_var, ActiveFedora::SolrService.post(
           extra_filters.to_s,
           fq: [
-            %(#{::ActiveFedora.index_field_mapper.solr_name(work_identifier)}:("#{complete_entry_identifiers.join('" OR "')}")),
+            %(#{solr_name(work_identifier)}:("#{complete_entry_identifiers.join('" OR "')}")),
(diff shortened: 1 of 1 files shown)


jeremyf (approved):

bbpennel:
I don't understand what the "PR has required labels" rule requires me to change?

Also wanted to make sure we are okay with the hyrax version requirement going up, as discussed above

ShanaLMoore:
> I don't understand what the "PR has required labels" rule requires me to change?
> 
> Also wanted to make sure we are okay with the hyrax version requirement going up, as discussed above

Bulkrax has the following [requirement](https://github.com/samvera-labs/bulkrax/actions/runs/3886410433/jobs/6633177936): 

`Error: Label check failed: required 1 of patch-ver, minor-ver, major-ver, ignore-for-release, but found 0.`

bbpennel:
I've updated this PR to make it backwards compatible with older versions of hyrax. Could someone trigger the workflows again?

orangewolf (approved):

bkiahstroud (approved):
Thank you for making this change!

Issue #702, opened 2023-01-11
FactoryClass changes during the parsing of metadata


jeremyf:
During the [Bulkrax::OaiEntry#build\_metadata](https://github.com/samvera-labs/bulkrax/blob/b857a25becefc4bd960f25647033dfb2e899386e/app/models/bulkrax/oai_entry.rb#L28-L47) we check [Bulkrax::HasMatchers#field\_supported?](https://github.com/samvera-labs/bulkrax/blob/b857a25becefc4bd960f25647033dfb2e899386e/app/models/concerns/bulkrax/has_matchers.rb#L123-L129) for each field.  Note the call to the `#factory_class` method.

```ruby
def field_supported?(field)
  field = field.gsub('_attributes', '')
  
  return false if excluded?(field)
  return true if supported_bulkrax_fields.include?(field)
  return factory_class.method_defined?(field) && factory_class.properties[field].present?
end
```

For works, the [Bulkrax::ImportBehavior#factory\_class](https://github.com/samvera-labs/bulkrax/blob/b857a25becefc4bd960f25647033dfb2e899386e/app/models/concerns/bulkrax/import_behavior.rb#L180-L197) definition is contingent on the `parsed_metadata`. 

```ruby
def factory_class
  fc = if self.parsed_metadata&.[]('model').present?
    	 self.parsed_metadata&.[]('model').is_a?(Array) ? self.parsed_metadata&.[]('model')&.first : self.parsed_metadata&.[]('model')
       elsif self.mapping&.[]('work_type').present?
    	 self.parsed_metadata&.[]('work_type').is_a?(Array) ? self.parsed_metadata&.[]('work_type')&.first : self.parsed_metadata&.[]('work_type')
       else
    	 Bulkrax.default_work_type
       end
  
  # return the name of the collection or work
  fc.tr!(' ', '_')
  fc.downcase! if fc.match?(/[-_]/)
  fc.camelcase.constantize
rescue NameError
  nil
rescue
  Bulkrax.default_work_type.constantize
end
```

With the `factory_class` derived from the `parsed_metadata` this means that during an entry’s ingest the `factory_class` can and will almost certainly change.  Thus up until we’ve set the `model` in the `parsed_metadata` we are using the configured default work type.

[...]

PR #703, opened 2023-01-11
Ensuring OAI parser sets factory class then parses

jeremyf:
## Ensuring OAI parser sets factory class then parses

a086164abe2b6f6ac79d81e1ef54c65ef8071136

Let's say we have nodes with the following names (and order): `shape`, `model`,
`smell`

Prior to this commit, as we were reading through the record's metadata
when we were processing `shape` we would use the default `factory_class`
to check if `shape` was a valid field.  Then we'd use the value of
`model` to derive the `factory_class`.  And when we then processed
`smell` we'd be using the newly derived `factory_class` from the `model`
node.

With this commit, we first establish the `factory_class` from the
`model` node.  Then we again process the all the nodes to parse the
metadata.  This ensures that both `shape` and `smell` are parsed
against the expected `factory_class`.

Further, I have made a few refactorings that should ease upstream
overrides by extracting methods from the body of the `build_metadata`
method.

Related to:

- https://github.com/samvera-labs/bulkrax/issues/702
- https://github.com/scientist-softserv/adventist-dl/issues/188
- https://github.com/scientist-softserv/adventist-dl/issues/187

## Removing comment that no longer applies

41328a85e100554b0323b4c35ec347e6c734535f

## Removing `break` in loop for setting factory class

173c710893ec63238c3e762e457519cab0002fb2

In discussions with Kiah, having the `break` adds a slight performance
improvement.  However the `break` did break with the existing pattern.

We do take efforts to set the factory_class to the first encountered
`model`, so having mutliple `models` is probably okay.

### DIFF
--- a/app/models/bulkrax/oai_entry.rb
+++ b/app/models/bulkrax/oai_entry.rb
@@ -64,7 +64,6 @@ module Bulkrax
       each_candidate_metadata_node do |node|
         next unless model_field_names.include?(node.name)
         add_metadata(node.name, node.content)
-        break
       end
     end
 
--- a/app/models/bulkrax/oai_entry.rb
+++ b/app/models/bulkrax/oai_entry.rb
@@ -30,12 +30,14 @@ module Bulkrax
       self.parsed_metadata[work_identifier] = [record.header.identifier]
       self.raw_metadata = { xml: record.metadata.to_s }
 
-      record.metadata.children.each do |child|
-        child.children.each do |node|
-          add_metadata(node.name, node.content)
-        end
-      end
-      add_metadata('thumbnail_url', thumbnail_url)
+      # We need to establish the #factory_class before we proceed with the metadata.  See
+      # https://github.com/samvera-labs/bulkrax/issues/702 for further details.
+      #
+      # tl;dr - if we don't have the right factory_class we might skip properties that are
+      # specifically assigned to the factory class
+      establish_factory_class
+      add_metadata_from_record
+      add_thumbnail_url
 
       add_visibility
       add_rights_statement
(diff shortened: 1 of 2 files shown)


app/models/bulkrax/oai_entry.rb
```diff
@@ -54,6 +56,40 @@ def collections_created?
...
+    # @see https://github.com/samvera-labs/bulkrax/issues/702
+    def establish_factory_class
+      # Why send?  Because in some implementations I've found that folks have privatized the
+      # model_field_mappings method.
+      model_field_names = parser.model_field_mappings
+
+      each_candidate_metadata_node do |node|
+        next unless model_field_names.include?(node.name)
+        add_metadata(node.name, node.content)
+        break
```
bkiahstroud, line 69: Am I reading this correctly? 

> 1. Loop through each metadata node, ignoring ones that don't match the configured `model_field_names` 
> 2. Once we hit a `model_field_names` node, add its metadata 
> 3. We have what we need, `break` 

Assuming that's correct, and assuming that there could be multiple `model_field_names` (given that it's plural), wouldn't the `break` mean that we only add the first `model_field_names` and skip all ones after it? 

Example: 
```ruby
model_field_names = ['bar', 'baz']

['foo', 'bar', 'baz'].each do |node|
  next unless model_field_names.include?(node.name) # skip 'foo'
  add_metadata(node.name, node.content) # add 'bar'
  break # skip 'baz'
end
```

app/models/bulkrax/oai_entry.rb
```diff
@@ -54,6 +56,40 @@ def collections_created?
       end
     end
 
+    # To ensure we capture the correct parse data, we first need to establish the factory_class.
+    # @see https://github.com/samvera-labs/bulkrax/issues/702
+    def establish_factory_class
+      # Why send?  Because in some implementations I've found that folks have privatized the
+      # model_field_mappings method.
```
bkiahstroud, line 63: What is this comment referring to? I don't see `#send` used anywhere besides the spec

jeremyf: Oh, that was there based on needs for Adventist.  But I've made changes there.  I'll amend the comment.

bkiahstroud (approved):

app/models/bulkrax/oai_entry.rb
bkiahstroud: After discussing with @jeremyf, the more pressing concern related to this is the current behavior of `#factory_class`. I've created [this ticket](https://github.com/samvera-labs/bulkrax/issues/704) capturing that discussion

PR #705, opened 2023-01-12
Ensuring XML parser sets factory class then parses

### LINKED ISSUE #702: FactoryClass changes during the parsing of metadata

jeremyf:
This commit for changing the `Bulkrax::XmlEntry` is the sibling to the changes made to `Bulkrax::OaiEntry` (see SHA 35970cee).  The salient information is pulled as a quote from that commit:

> Prior to this commit, as we were reading through the record's metadata
> when we were processing `shape` we would use the default
> `factory_class` to check if `shape` was a valid field.  Then we'd use
> the value of `model` to derive the `factory_class`.  And when we then
> processed `smell` we'd be using the newly derived `factory_class` from
> the `model` node.
>
> With this commit, we first establish the `factory_class` from the
> `model` node.  Then we again process the all the nodes to parse the
> metadata.  This ensures that both `shape` and `smell` are parsed
> against the expected `factory_class`.

In addition, I have refactored the spec to remove global state pollution of the parser config.

I have also renamed and documented a method to be more descriptive; the previous method name of `xml_elements` was opaque, as it returned an Array of strings (and not an array of XML elements).

Further steps could be to extract the now "ubiquitous" `Avacado` class into a spec fixture.

We may want to revisit the XML parser fundamental logic; namely we only parse nodes that are explicitly declared with in the `from`.  This is a bit different than other parsers, in that they will make assumptions about each encountered column (in the case of CSV) or node (in the case of OAI).  tl;dr - Here there be dragons.

Closes #702

Related to:

- https://github.com/samvera-labs/bulkrax/pull/703
- https://github.com/scientist-softserv/adventist-dl/issues/188
- https://github.com/scientist-softserv/adventist-dl/issues/187
- https://github.com/scientist-softserv/adventist-dl/issues/186

### DIFF
--- a/app/models/bulkrax/xml_entry.rb
+++ b/app/models/bulkrax/xml_entry.rb
@@ -43,15 +43,14 @@ module Bulkrax
       raise StandardError, "Missing source identifier (#{source_identifier})" if self.raw_metadata[source_identifier].blank?
       self.parsed_metadata = {}
       self.parsed_metadata[work_identifier] = [self.raw_metadata[source_identifier]]
-      xml_elements.each do |element_name|
-        elements = record.xpath("//*[name()='#{element_name}']")
-        next if elements.blank?
-        elements.each do |el|
-          el.children.map(&:content).each do |content|
-            add_metadata(element_name, content) if content.present?
-          end
-        end
-      end
+
+      # We need to establish the #factory_class before we proceed with the metadata.  See
+      # https://github.com/samvera-labs/bulkrax/issues/702 for further details.
+      #
+      # tl;dr - if we don't have the right factory_class we might skip properties that are
+      # specifically assigned to the factory class
+      establish_factory_class
+      add_metadata_from_record
       add_visibility
       add_rights_statement
       add_admin_set_id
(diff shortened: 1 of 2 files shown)


bkiahstroud (approved):
Looks like [rubocop is failing](https://github.com/samvera-labs/bulkrax/actions/runs/3904440548/jobs/6670228028). Otherwise, approved!

PR #707, opened 2023-01-13
Add Permissions For Bulkrax

orangewolf:
# Issue #
Ref https://github.com/scientist-softserv/britishlibrary/issues/227

# Scope #
- [x] put limits around the sidebar links
- [x] put limits in the controllers
- [ ] scope exporter for allowed works

# For Release Notes #

There are 2 new Ability methods that need to be added to `ability.rb`. If you are using the generator, these methods will be added for you. However, if you are upgrading Bulkrax, you will need to add the following to your `app/models/ability.rb` file above the `can_create_any_works?` method definition. Feel free to adjust the definitions as needed.

```
    def can_import_works?
      can_create_any_work?
    end
    
    def can_export_works?
      can_create_any_work?
    end
```

### DIFF
--- a/lib/generators/bulkrax/install_generator.rb
+++ b/lib/generators/bulkrax/install_generator.rb
@@ -60,17 +60,19 @@ class Bulkrax::InstallGenerator < Rails::Generators::Base
     file_text = File.read(file)
     import_line = 'def can_import_works?'
     export_line = 'def can_export_works?'
-    if !file_text.include?(import_line)
-      insert_into_file file, before: /\s*def can_create_any_work\?/ do
-        "    def can_import_works?\n      can_create_any_work?\n    end"
+    unless file_text.include?(import_line)
+      insert_into_file file, before: /^end/ do
+        "  def can_import_works?\n    can_create_any_work?\n  end"
       end
     end
 
-    if !file_text.include?(export_line)
-      insert_into_file file, before: /\s*def can_create_any_work\?/ do
-        "    def can_export_works?\n      can_create_any_work?\n    end"
+    # rubocop:disable Style/GuardClause
+    unless file_text.include?(export_line)
+      insert_into_file file, before: /^end/ do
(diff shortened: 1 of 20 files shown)


bkiahstroud (approved):

lib/generators/bulkrax/install_generator.rb
```diff
@@ -55,6 +55,26 @@ def add_javascripts
     end
   end
 
+  def add_ability
+    file = 'app/models/ability.rb'
```
alishaevn, line 59: do we need to account for if this file doesn't exist at all?

alishaevn (changes requested):
blocking the merge per my previous comment.

lib/generators/bulkrax/install_generator.rb
bkiahstroud: [Hyrax expects the file to exist](https://github.com/samvera/hyrax/blob/bf5df6f9d47ff04dc5d103b5922bcb80a9bb821c/lib/generators/hyrax/install_generator.rb#L104-L110), so I think we would be okay to as well

bkiahstroud: Do you know of / have a case where the file doesn't exist?

alishaevn: ah ok. no, I don't know of any instance where it wouldn't...just wanted to make sure our bases were covered. 🙂

alishaevn (approved):
approving per kiah's responses to my question.

lib/generators/bulkrax/install_generator.rb
bkiahstroud: It's a completely valid question, [we just ran into that issue in the Actions repo](https://github.com/scientist-softserv/actions/commit/9aac178a68bb9d0ee5e3eccb11899b5da7025f80) last week. So, better safe than sorry 🙂

PR #374, opened 2021-10-19
Respect nil values

alishaevn:
**https://github.com/samvera-labs/bulkrax/pull/373 must be merged first or we'll have the same problem that's fixed there.**

Ref: https://gitlab.com/notch8/britishlibrary/-/issues/185

# Expected behavior
- on import, all cell values (empty or filled in) are accounted for
- if a user would like to delete data on an import, remove the value from the cell under the appropriate header in the csv, and it will be removed from the metadata for that work
- metadata and files can be removed this way

# Demo
| | csv's | before | after |
| --- | --- | --- | --- |
| metadata only | with all values: [185_bc.csv](https://github.com/samvera-labs/bulkrax/files/7376400/185_bc.csv)<br> with only required values: [185_bc_round_trip.csv](https://github.com/samvera-labs/bulkrax/files/7376402/185_bc_round_trip.csv) | image | image |
| metadata and file | with all values and a file:  [185_gw.csv](https://github.com/samvera-labs/bulkrax/files/7376409/185_gw.csv)<br> with all values, but no file: [185_gw_round_trip.csv](https://github.com/samvera-labs/bulkrax/files/7376410/185_gw_round_trip.csv) | image |  image |

PR #708, opened 2023-01-13
Scour blank strings from the creation attributes

jeremyf:
There are three changes in this commit:

1. Exposing a class attribute
2. Scouring attribute hash
3. Extracting a method

**Exposing a class attribute**

This was added to allow for easier testing of a private method.  It also "exposes" a hidden assumption in a private method; turning that assumption into a telegraphed potential configuration point.

**Scouring attribute hash**

This resolves the reported problem by @eporter23:

> When testing an import on 9/16/22, the work row failed to validate,
> returning an error regarding "invalid entry for administrative_unit".
>
> The CSV used had the header for that field, but no values entered. Our
> preprocessors output field headers for mapped fields, but sometimes
> there are no values entered for a given header.
>
> -- https://github.com/emory-libraries/dlp-curate/issues/1973

The likely most vexing problem being an array that has only empty values.  This is likely something that should be remedied upstream in the data modeling for Hyrax models; but alas we're not there yet.

**Extracting a method**

This is a happy little refactor to create symmetry between the other create/update methods for collections and file sets.  In doing this, we provide a mechanism for folks to more readily over-ride methods at a lower level.

Related to:

- https://github.com/emory-libraries/dlp-curate/issues/1973
- https://github.com/emory-libraries/dlp-curate/pull/1981

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -16,6 +16,17 @@ module Bulkrax
     class_attribute :base_permitted_attributes,
       default: %i[id edit_users edit_groups read_groups visibility work_members_attributes admin_set_id]
 
+    # @return [Boolean]
+    #
+    # @example
+    #   Bulkrax::ObjectFactory.transformation_removes_blank_hash_values = true
+    #
+    # @see #transform_attributes
+    # @see https://github.com/samvera-labs/bulkrax/pull/708 For discussion concerning this feature
+    # @see https://github.com/samvera-labs/bulkrax/wiki/Interacting-with-Metadata For documentation
+    #      concerning default behavior.
+    class_attribute :transformation_removes_blank_hash_values, default: false
+
     define_model_callbacks :save, :create
     attr_reader :attributes, :object, :source_identifier_value, :klass, :replace_files, :update_files, :work_identifier, :related_parents_parsed_mapping, :importer_run_id
 
@@ -245,7 +256,7 @@ module Bulkrax
     def transform_attributes(update: false)
       @transform_attributes = attributes.slice(*permitted_attributes)
       @transform_attributes.merge!(file_attributes(update_files)) if with_files
-      @transform_attributes = remove_blank_hash_values(@transform_attributes)
+      @transform_attributes = remove_blank_hash_values(@transform_attributes) if transformation_removes_blank_hash_values?
       update ? @transform_attributes.except(:id) : @transform_attributes
     end
 
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -6,6 +6,14 @@ module Bulkrax
     include Bulkrax::FileFactory
     include DynamicRecordLookup
 
+    # These are the attributes that we assume all "work type" classes (e.g. the given :klass) will
+    # have in addition to their specific attributes.
+    #
+    # @return [Array<Symbol>]
+    # @see #permitted_attributes
+    class_attribute :base_permitted_attributes,
+      default: %i[id edit_users edit_groups read_groups visibility work_members_attributes admin_set_id]
+
     define_model_callbacks :save, :create
     attr_reader :attributes, :object, :source_identifier_value, :klass, :replace_files, :update_files, :work_identifier, :related_parents_parsed_mapping, :importer_run_id
 
@@ -58,7 +66,7 @@ module Bulkrax
         elsif klass == FileSet
           update_file_set(attrs)
         else
-          work_actor.update(environment(attrs))
+          update_work(attrs)
         end
       end
       object.apply_depositor_metadata(@user) && object.save! if object.depositor.nil?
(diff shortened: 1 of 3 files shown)


app/factories/bulkrax/object_factory.rb
```diff
@@ -227,12 +243,32 @@ def collection_type(attrs)
...
+    def remove_blank_hash_values(attributes)
+      dupe = attributes.dup
+      dupe.each do |key, values|
+        if values.is_a?(Array) && values.all? { |value| value.is_a?(String) && value.empty? }
+          dupe[key] = []
+        elsif values.is_a?(String) && values.empty?
+          dupe[key] = nil
+        end
+      end
+      dupe
```
alishaevn, line 271: will this interfere with the intent of pr https://github.com/samvera-labs/bulkrax/pull/374?

app/factories/bulkrax/object_factory.rb
```diff
@@ -6,6 +6,14 @@ class ObjectFactory
     include Bulkrax::FileFactory
     include DynamicRecordLookup
 
+    # These are the attributes that we assume all "work type" classes (e.g. the given :klass) will
+    # have in addition to their specific attributes.
+    #
+    # @return [Array<Symbol>]
+    # @see #permitted_attributes
+    class_attribute :base_permitted_attributes,
```
bbpennel, line 14: Breaking this out into a variable is a good idea. Is it configurable via bulkrax.rb? If so, should we add a commented out example in there?

And is this list of attributes only applied to works? In my local testing I ran into issues where Collections seemed to be getting `admin_set_id` called even though they didn't support it, so I ended up [excluding it via permitted_attributes.](https://github.com/UNC-Libraries/hy-c/blob/fae572c0ba4cd448e12560b5b060b48fa304d70d/app/overrides/factories/bulkrax/object_factory_override.rb#L68)

bbpennel (approved):
This looks good to me, and functioned as expected in my local repository. I had been just starting to look into this issue, so this is very timely.

app/factories/bulkrax/object_factory.rb
jeremyf: That's a good question, and I don't know the answer.  My cursory read is that there might be a conflict.  But we'll need to discuss the spec.

@bwatson78 I'd love for you to weigh in to help us triangulate on approaches.

app/factories/bulkrax/object_factory.rb
jeremyf: @bbpennel This is one of those "quiet configuration points".  In the spec, I demonstrate that you can do the config on an instance by instance level.


But to make the change at the application level you'd likely add the following to your Bulkrax initializer:

```ruby
Bulkrax::ObjectFactory.base_permitted_attributes = %i[name title admin_set_id]
```

This is not a config that I want to consider a public API (and am updating the documentation accordingly)

app/factories/bulkrax/object_factory.rb
alishaevn: @jeremyf this [slack convo](https://assaydepot.slack.com/archives/G0313NK5NMA/p1673638808870859) may also help us. in the event that this code does make #374 irrelevant, there's still a way to delete data with bulkrax. however...it would require updating users on how deleting is done.

bkiahstroud: In addition to @alishaevn's question, there is [this section in the Wiki](https://github.com/samvera-labs/bulkrax/wiki/Interacting-with-Metadata) that I think has extra context; it lays out how Bulkrax should interpret empty cells, the absence of columns, etc. I think we should either: 

1. Make sure the behavior still matches the Wiki, as that's what is intended / what users expect, or 
2. Update Bulkrax's opinions (including the Wiki article) as to how we want to handle these use cases

app/factories/bulkrax/object_factory.rb
```diff
@@ -16,6 +16,17 @@ class ObjectFactory
...
+    #
+    # @example
+    #   Bulkrax::ObjectFactory.transformation_removes_blank_hash_values = true
+    #
+    # @see #transform_attributes
+    # @see https://github.com/samvera-labs/bulkrax/pull/708 For discussion concerning this feature
+    # @see https://github.com/samvera-labs/bulkrax/wiki/Interacting-with-Metadata For documentation
+    #      concerning default behavior.
+    class_attribute :transformation_removes_blank_hash_values, default: false
+
```
alishaevn, line 29: 👍🏾

alishaevn (approved):

app/factories/bulkrax/object_factory.rb
alishaevn: the wiki will also need to be updated to reflect this change.

bkiahstroud (approved):
Lgtm! How should we socialize this addition?

bkiahstroud:
> Lgtm! How should we socialize this addition?

Ah, I see [Alisha has already suggested updating the Wiki](https://github.com/samvera-labs/bulkrax/pull/708#discussion_r1071433502), which I'm on board with.

app/factories/bulkrax/object_factory.rb
bwatson78: @jeremyf I apologize for not responding sooner. I see that this is already merged. If the question still stands, let me know.

Issue #711, opened 2023-01-17
Source field as source_identifier export issues


bbpennel:
I have the `source` field configured as the `source_identifier` for the CSV parser, and I am finding that the source field is being returned strangely in exports which causes the field to get mangled when trying to round trip.

If I use a CSV import to create a new work with `source` field "work_20230117", it creates a fedora record with `<http://purl.org/dc/terms/source> "work_20230117"`, which seems correct. But when creating an export from this importer, the CSV comes back with the `source` column having the value `["work_20230117"]`, and a `source_1` colum with the value `work_20230117`. 

If I import this document, then fedora ends up with `<http://purl.org/dc/terms/source> "[\"work_20230117\"]"`, and doing another CSV export returns `source` with value `["[\"work_20230117\"]", "work_20230117"]`, with `source_1` containing `[\"work_20230117\"]` and `source_2` containing `work_20230117`. I assume it'll keep expanding from there.

If I edit the `source` field to remove the `[""]`, it imports fine and doesn't end up in fedora with the extra punctuation.

I'm not sure if the expected behavior is for there to just be a `source` field with value `work_20230117` or `source_1` with the same value, or if having multiple values in source_identifier should be disallowed.

bbpennel:
I opened a PR that might be one approach to resolving this. Feedback appreciated.

PR #712, opened 2023-01-17
Convert source_identifier from relationship to single value

bbpennel:
Address https://github.com/samvera-labs/bulkrax/issues/711

I'm not sure if this is the right approach or not, but it resolves issue #711 by converting the source_identifier field from a relationship to a single value during export. The result is that the export CSV has `work_20230117` as the source_identifier value, rather than `["work_20230117"]`. A `source_1` field is also still added in my testing, not sure if this is desirable or not.

The test is also not great, since its difficult to have fields return Relation without using Work.create, which requires running the whole stack.

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -107,7 +107,9 @@ module Bulkrax
     # Metadata required by Bulkrax for round-tripping
     def build_system_metadata
       self.parsed_metadata['id'] = hyrax_record.id
-      self.parsed_metadata[source_identifier] = hyrax_record.send(work_identifier)
+      source_id = hyrax_record.send(work_identifier)
+      source_id = source_id.to_a.first if source_id.is_a?(ActiveTriples::Relation)
+      self.parsed_metadata[source_identifier] = source_id
       self.parsed_metadata[key_for_export('model')] = hyrax_record.has_model.first
     end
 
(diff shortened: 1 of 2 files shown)


jeremyf (approved):
The code looks good.  However, I'd love to hear @alishaevn weigh in

app/models/bulkrax/csv_entry.rb
```diff
@@ -107,7 +107,9 @@ def build_export_metadata
     # Metadata required by Bulkrax for round-tripping
     def build_system_metadata
       self.parsed_metadata['id'] = hyrax_record.id
-      self.parsed_metadata[source_identifier] = hyrax_record.send(work_identifier)
+      source_id = hyrax_record.send(work_identifier)
+      source_id = source_id.to_a.first if source_id.is_a?(ActiveTriples::Relation)
+      self.parsed_metadata[source_identifier] = source_id
```
alishaevn, line 112: in the event that `source_id` is not an  ActiveTriples::Relation, does that mean that sometimes `self.parsed_metadata[source_identifier]` will be a string and other times it will be another data type?

that feels weird to me.

alishaevn (changes requested):

app/models/bulkrax/csv_entry.rb
bbpennel: The value was being used as-is previously, so out of caution i took the approach of checking for Relation before transforming it just in case the previous behavior was applicable for some situation I wasn't aware of. I'm okay with just calling `source_id.to_a.first` if we are sure the value will always be a Relation (assuming there is not a misconfiguration and an unsuitable field is attempting to be used for the source_identifier, like an object field or something).

alishaevn: my first thought was that this felt weird, but I've been looking at some code and things and I think this change might actually be ok.

alishaevn (approved):

PR #719, opened 2023-01-23
Ensuring CSV sets factory class before metadata parsing

### LINKED ISSUE #718: CSV Parser can raise `NoMethodError` due to factory_class not being set
This was reported in https://github.com/scientist-softserv/britishlibrary/issues/211

jeremyf:
## Ensuring CSV sets factory class before metadata parsing

dbd24b8ee1c6adc5907ef95a93dbe0fdd745ad79

Prior to this commit, if we had configured bulkrax such that we were
allowing for our model import to come from multiple fields
(e.g. `from: ['model', 'work_type']`) and the CSV had `work_type` as the
column name, we would get an error as follows:

```
NoMethodError:
  undefined method `method_defined?' for nil:NilClass
    # ./app/models/concerns/bulkrax/has_matchers.rb:128:in `field_supported?'
    # ./app/models/concerns/bulkrax/has_matchers.rb:157:in `multiple?'
    # ./app/models/concerns/bulkrax/has_matchers.rb:34:in `block in add_metadata'
    # ./app/models/concerns/bulkrax/has_matchers.rb:31:in `each'
    # ./app/models/concerns/bulkrax/has_matchers.rb:31:in `add_metadata'
    # ./app/models/bulkrax/csv_entry.rb:77:in `block in add_ingested_metadata'
    # ./app/models/bulkrax/csv_entry.rb:75:in `each'
    # ./app/models/bulkrax/csv_entry.rb:75:in `add_ingested_metadata'
    # ./app/models/bulkrax/csv_entry.rb:43:in `build_metadata'
```

Tracing this line it is because the `factory_class` is nil.  And if the
model column were to be a later column, any columns that occurred before
we established the factory class may have been skipped.

With this commit, we're echoing the functionality introduced in #705 and #703.

Related to:

- https://github.com/scientist-softserv/britishlibrary/issues/211
- https://github.com/samvera-labs/bulkrax/pull/705
- https://github.com/samvera-labs/bulkrax/pull/703

Closes #718

## Update spec/fixtures/csv/factory_class_test.csv

f47505d321630c45f2782fddc5392767465951a1

Co-authored-by: Benjamin Kiah Stroud <32469930+bkiahstroud@users.noreply.github.com>

### DIFF
--- a/spec/fixtures/csv/factory_class_test.csv
+++ b/spec/fixtures/csv/factory_class_test.csv
@@ -1,2 +1,2 @@
-work_type,,title,shape,source_identifier
+work_type,title,shape,source_identifier
 Avacado,A tasty treat,Kind of Lumpy,1234-5678
--- a/spec/fixtures/csv/factory_class_test.csv
+++ b/spec/fixtures/csv/factory_class_test.csv
@@ -1,0 +1,2 @@
+work_type,,title,shape,source_identifier
+Avacado,A tasty treat,Kind of Lumpy,1234-5678
--- a/spec/models/bulkrax/csv_entry_spec.rb
+++ b/spec/models/bulkrax/csv_entry_spec.rb
@@ -18,6 +18,40 @@ module Bulkrax
       end
     end
 
+    describe '#build_metadata' do
+      around do |spec|
+        # Why am I doing this instead of passing it into the parser?  Because of
+        # Bulkrax::ApplicationParser#model_field_mappings.  It uses the bulkrax configuration.
+        old = Bulkrax.field_mappings['Bulkrax::CsvParser']
+        Bulkrax.field_mappings['Bulkrax::CsvParser'] = {
+          "model" => { from: ["model", "work_type"] },
+          "title" => { from: ["title"] },
+          "shape" => { from: ["shape"] },
+          'source_identifier' => { from: ['source_identifier'], source_identifier: true }
+        }
+        class ::Avacado < Work
+          property :shape, predicate: ::RDF::Vocab::DC.format
+        end
+        spec.run
+        Object.send(:remove_const, :Avacado)
+        Bulkrax.field_mappings['Bulkrax::CsvParser'] = old
+      end
+      let(:path) { File.expand_path('../../fixtures/csv/factory_class_test.csv', __dir__) }
+      let(:importer) do
+        FactoryBot.build(
+          :bulkrax_importer_csv,
+          parser_fields: { "import_file_path" => path }
+        )
+      end
+      let(:data) { { work_type: "Avacado", title: "A Yummy Friend", shape: "Kind of Lumpy", source_identifier: "1234-5678" } }
+      subject(:entry) { described_class.new(importerexporter: importer, raw_metadata: data, identifier: data.fetch(:source_identifier)) }
+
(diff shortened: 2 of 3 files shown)


spec/fixtures/csv/factory_class_test.csv
```diff
@@ -0,0 +1,2 @@
+work_type,,title,shape,source_identifier
```
bkiahstroud, line 1: Accidental extra comma? 

```suggestion
work_type,title,shape,source_identifier
```

spec/models/bulkrax/csv_entry_spec.rb
```diff
@@ -18,6 +18,40 @@ module Bulkrax
...
+          'source_identifier' => { from: ['source_identifier'], source_identifier: true }
+        }
+        class ::Avacado < Work
+          property :shape, predicate: ::RDF::Vocab::DC.format
+        end
+        spec.run
+        Object.send(:remove_const, :Avacado)
+        Bulkrax.field_mappings['Bulkrax::CsvParser'] = old
+      end
+      let(:path) { File.expand_path('../../fixtures/csv/factory_class_test.csv', __dir__) }
```
bkiahstroud, line 39: Is there a benefit to using `File.expand_path ...` instead of utilizing [the configured fixture_path](https://github.com/samvera-labs/bulkrax/blob/c3a638139be3ec537aaee250489c7e3c8a4309f3/spec/rails_helper.rb#L72) or [a simple string](https://github.com/samvera-labs/bulkrax/blob/main/spec/parsers/bulkrax/csv_parser_spec.rb#L134)?

bkiahstroud (commented):
A couple clarification questions. Otherwise, looks good!

spec/models/bulkrax/csv_entry_spec.rb
jeremyf: No benefit, it's what I've carried forward.

ShanaLMoore (approved):

ShanaLMoore:
This looks good, however I'm kind of curious why the multiple field mapping wasn't working 🤔 Is multiple field mapping a problem for other properties as well?

jeremyf:
> This looks good, however I'm kind of curious why the multiple field mapping wasn't working 🤔 Is multiple field mapping a problem for other properties as well? @jeremyf

The underlying problem is that there is something in the processing that raised a `NameError` which meant we had a a `nil`.  Another confounding problem is that during the import, the `factory_class` can change.

The below method is that code and it does a lot of work:

```ruby
def factory_class
  fc = if self.parsed_metadata&.[]('model').present?
         self.parsed_metadata&.[]('model').is_a?(Array) ? self.parsed_metadata&.[]('model')&.first : self.parsed_metadata&.[]('model')
       elsif self.mapping&.[]('work_type').present?
         self.parsed_metadata&.[]('work_type').is_a?(Array) ? self.parsed_metadata&.[]('work_type')&.first : self.parsed_metadata&.[]('work_type')
       else
         Bulkrax.default_work_type
       end

  # return the name of the collection or work
  fc.tr!(' ', '_')
  fc.downcase! if fc.match?(/[-_]/)
  fc.camelcase.constantize
rescue NameError
  nil
rescue
  Bulkrax.default_work_type.constantize
end
```

PR #725, opened 2023-02-02
Ensuring self-contained memoization of method

jeremyf:
Prior to this commit, the guard condition `return self.collection_ids if collections_created?` served as memoization (e.g. caching the results of an expensive operation).  However, it was not a self-contained memoization (e.g. the method is fully in charge of managing the memoization).  It instead relied on the previous implementation of `#collections_created?`.

Downstream implementations may change `#collection_created?` which could create unintended consenquences of assignign collections based on the OAI spec.  Also, if we look to other implementations in Bulkrax of `#collections_created?` we see that many simply `return true`.

The impact of the previous code is that were someone to make the above change (e.g. always `return true`), we end up never assigning any `setSpec` `collection_ids` and always returning `[]` for `#find_collection_ids`.  dependent on external methods.

With this commit, we are internalizing the memoization mechanism so that this method is less dependent on external factories.

For further discussion of the interplay of the methods regarding collection creation see the [comments found here][1].

References:

- https://github.com/scientist-softserv/adventist-dl/pull/247
- https://github.com/scientist-softserv/adventist-dl/issues/240

[1]: https://github.com/scientist-softserv/adventist-dl/issues/240#issuecomment-1412456590

### DIFF
--- a/app/models/bulkrax/oai_entry.rb
+++ b/app/models/bulkrax/oai_entry.rb
@@ -91,7 +91,8 @@ module Bulkrax
     # If OAI-PMH doesn't return setSpec in the headers for GetRecord, use parser.collection_name
     #   in this case, if 'All' is selected, records will not be added to a collection.
     def find_collection_ids
-      return self.collection_ids if collections_created?
+      return self.collection_ids if defined?(@called_find_collection_ids)
+
       if sets.blank? || parser.collection_name != 'all'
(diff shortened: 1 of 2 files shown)


kirkkwang (approved):
Nice, looks like a good way of getting around it

laritakr (approved):

PR #731, opened 2023-02-09
Favoring ActiveRecord query over in-memory select

jeremyf:
Prior to this commit, we were using array selection.  This loads each object into memory and then selects based on the criteria.

With this commit, we're relying on ActiveRecord's querying to perform the selection.  This means we only load in memory the objects that match our selection criteria.


orangewolf (approved):

orangewolf:
top gun meme

bkiahstroud (approved):

ShanaLMoore (approved):

jeremyf:
The failures are legit, in that our stubs and mocks are assuming a different approach.  To remediate would require a significant rework of the factory bot declarations.  I'm closing this for the moment.

PR #737, opened 2023-02-13
replace multiple times through the pending realtionship loop

orangewolf:
with a single run that never adds all the objects to memory

### DIFF
--- a/app/jobs/bulkrax/create_relationships_job.rb
+++ b/app/jobs/bulkrax/create_relationships_job.rb
@@ -51,88 +51,48 @@ module Bulkrax
     # parent_identifier or child_identifier param. For example, if a parent_identifier is passed, we know @base_entry
     # is the child in the relationship, and vice versa if a child_identifier is passed.
     def perform(parent_identifier:, importer_run_id:) # rubocop:disable Metrics/AbcSize
-      pending_relationships = Bulkrax::PendingRelationship.find_each.select do |rel|
-        rel.importer_run_id == importer_run_id && rel.parent_id == parent_identifier
-      end.sort_by(&:order)
+      # TODO permission checks on child / parent write access
 
-      @importer_run_id = importer_run_id
-      @parent_entry, @parent_record = find_record(parent_identifier, importer_run_id)
-      @child_records = { works: [], collections: [] }
-      pending_relationships.each do |rel|
-        raise ::StandardError, %("#{rel}" needs either a child or a parent to create a relationship) if rel.child_id.nil? || rel.parent_id.nil?
-        @child_entry, child_record = find_record(rel.child_id, importer_run_id)
-        if child_record
-          child_record.is_a?(::Collection) ? @child_records[:collections] << child_record : @child_records[:works] << child_record
+      # TODO rename order field literally anything else. use it for sort
+      pending_relationships = Bulkrax::PendingRelationship.where(parent_id: parent_identifier, importer_run_id: importer_run_id).order('created_at')
+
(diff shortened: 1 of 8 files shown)


sephirothkod (approved):

app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -51,88 +49,75 @@ class CreateRelationshipsJob < ApplicationJob
...
+      child_record.save!
+    end
+
+    def add_to_work(child_record, parent_record)
+      return true if parent_record.ordered_members.include?(child_record)
+
+      parent_record.ordered_members << child_record
+      @parent_record_members_added = true
+      # TODO: Do we need to save the child record?
+      child_record.save!
```
orangewolf, line 120: i bet you dont need to save here.

orangewolf:
@jeremyf this looks good to me. I can't approve because I opened the PR.

app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -51,88 +49,75 @@ class CreateRelationshipsJob < ApplicationJob
...
-      @parent_entry, @parent_record = find_record(parent_identifier, importer_run_id)
-      @child_records = { works: [], collections: [] }
-      pending_relationships.each do |rel|
-        raise ::StandardError, %("#{rel}" needs either a child or a parent to create a relationship) if rel.child_id.nil? || rel.parent_id.nil?
-        @child_entry, child_record = find_record(rel.child_id, importer_run_id)
-        if child_record
-          child_record.is_a?(::Collection) ? @child_records[:collections] << child_record : @child_records[:works] << child_record
-        end
+      importer_run = Bulkrax::ImporterRun.find(importer_run_id)
+      ability = Ability.new(importer_run.user)
```
bkiahstroud, line 53: Bulkrax's CLI [doesn't currently require](https://github.com/samvera-labs/bulkrax/blob/main/lib/generators/bulkrax/templates/config/bulkrax_api.yml#L65-L66) `--user_id` to be set, so that could introduce unexpected errors here. 

I don't think the CLI currently gets a ton of use, so I don't think figuring out how to handle that is critical for this change to go through, but it's worth mentioning.

app/models/bulkrax/pending_relationship.rb
```diff
@@ -3,5 +3,9 @@
 module Bulkrax
   class PendingRelationship < ApplicationRecord
     belongs_to :importer_run
+
+    # Ideally we wouldn't have a column named "order", as it is a reserved SQL term.  However, if we
+    # quote the column, all is well...for the application.
+    scope :ordered, -> { order("#{quoted_table_name}.#{connection.quote_column_name('order')}") }
```
bkiahstroud, line 9: @jeremyf what do you mean by "for the application"?

jeremyf: In this case the application logic should work.  But querying via SQL command would require quoting (whereas any non-reserved column names might not).

app/jobs/bulkrax/create_relationships_job.rb
jeremyf: That's definitely good to know.  And I suspect makes sense to have a fallback configured user.

jeremyf: I can add that configuration, as it's good to capture that.

app/jobs/bulkrax/create_relationships_job.rb
jeremyf: It was previously happening in the code.  I'm inclined to keep it…for now.

PR #738, opened 2023-02-13
Renaming method to reflect it's state changing nature

jeremyf:
Prior to this commit, `status_info` was changing underlying runner information.  This is unexpected behavior given the method name (status_info implies getting information about the status information.)

With this commit, I'm deprecating the `status_info` method with a v6.0.0 deprecation horizon.


orangewolf (approved):
smart. I like it

PR #741, opened 2023-02-15
Capturing record and header metadata for OAI parser

jeremyf:
Prior to this commit, we didn't see the header information.  With this information, we now have the header information.

### DIFF
--- a/app/models/bulkrax/oai_entry.rb
+++ b/app/models/bulkrax/oai_entry.rb
@@ -37,7 +37,7 @@ module Bulkrax
     def build_metadata
       self.parsed_metadata = {}
       self.parsed_metadata[work_identifier] = [record.header.identifier]
-      self.raw_metadata = { xml: record.metadata.to_s }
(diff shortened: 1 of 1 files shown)


app/models/bulkrax/oai_entry.rb
```diff
@@ -37,7 +37,7 @@ def thumbnail_url
     def build_metadata
       self.parsed_metadata = {}
       self.parsed_metadata[work_identifier] = [record.header.identifier]
-      self.raw_metadata = { xml: record.metadata.to_s }
```
orangewolf, line 40: the change to `record` here seems like it will cause anything that looks at raw_metadata to need to change, wouldn't it?

jeremyf: I don't think anything looks at raw_metadata in the build process.  (Especially given that I just added the previous state recently.)

orangewolf (approved):

PR #749, opened 2023-02-21
Adding `Bulkrax::ParserExportRecordSet`

jeremyf:
Prior to this commit the `Bulkrax::CsvParser` had complex query logic
for finding the records to process for export.  The logic was memory
inefficient: calling `#current_record_ids` twice (a very expensive
queries that issued many calls to Fedora which could've been a SOLR
query).  In large exports, the processes would spend over 90% of the
time in the finding file sets.

With this commit the `Bulkrax::ParserExportRecordSet` finder logic
echoes the logic once found in `Bulkrax::CsvParser`.  The purpose is to
isolate the place where the SOLR queries happen and to create less
memory intensive processes.

Further the `Bulkrax::ParserExportRecordSet.for` returns an object that
"quacks" like an Array of Arrays, with each inner Array having two
elements.  This makes the testing a bit easier.

Further, this extraction helps us create the conditions for other
exporters to re-use the export from logic that was presently tucked away
in the CSV Parser.

Of particular note, there are no real SOLR query tests of these changes,
nor the code prior to this commit.

Related to:

### DIFF
--- a/app/parsers/bulkrax/bagit_parser.rb
+++ b/app/parsers/bulkrax/bagit_parser.rb
@@ -85,18 +85,6 @@ module Bulkrax
       status_info(e)
     end
 
-    def total
-      @total = importer.parser_fields['total'] || 0 if importer?
-
-      @total = if exporter?
-                 limit.nil? || limit.zero? ? current_record_ids.count : limit
-               end
-
-      return @total || 0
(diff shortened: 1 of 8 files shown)


orangewolf (approved):
This looks really great! Good work

PR #752, opened 2023-02-28
Favoring explicit "empty" lines handler

jeremyf:
Prior to this commit, we were relying on the documented `CSV.read`'s `:skip_lines` option.

However, in testing the CSV that is part of this commit, the `CSV.read` would never complete.  Diving deep into the internals, it kept finding lines that upon inspecttion were an empty string.

With this commit, we're moving the skip logic outside of the parser and instead wrapping the results.  This is perhaps less performant, but compared to never completing due to infinite string processing, it actually is more performant.

Further, I have added additional configuration options in the event that this change does not work downstream.  Those configuration options should allow for more immediate relief as the general solution works its way back into Bulkrax.

This refines changes made in 792ccf0a7cfd03eb9083e76b134081af606df329.

Related to:

- https://github.com/samvera-labs/bulkrax/pull/751
- https://github.com/scientist-softserv/adventist-dl/issues/113
- https://github.com/scientist-softserv/britishlibrary/issues/280

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -13,34 +13,48 @@ module Bulkrax
       data.headers.flatten.compact.uniq
     end
 
-    # Following on the {::CSV.new} documentation, this little module is responsible for determining
-    # if we should skip the line we read via {CSV.read}.
-    #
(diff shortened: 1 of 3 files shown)


kirkkwang (approved):
Cool, good catch!

PR #753, opened 2023-03-01
filters out keys with blank values for required elements check

sephirothkod:


jeremyf (changes requested):
Please write a test the demonstrates what this code change is doing.

I'm also wondering if it makes more sense to have `#required_elements?` do the transformation.

It seems like putting it in that method might be a more robust solution for other entry handlers.

app/models/bulkrax/csv_entry.rb
```diff
@@ -70,9 +70,10 @@ def self.data_for_entry(data, _source_id, parser)
 
     def build_metadata
       raise StandardError, 'Record not found' if record.nil?
```
jeremyf, line 72: As I'm thinking on this more, I think it would also make sense to extract this guard condition to it's own method.  The rest of this method is calls to other methods; which makes the raise errors section a bit of an outlier.  It also means that in https://github.com/scientist-softserv/palni-palci/pull/428 you needed to override the entire method.

sephirothkod: totally agree! I was actually thinking along the same lines yesterday when i was looking at this. I have refactored everything now and I think it's a lot better, but I'm open to any other changes you might suggest.

app/models/bulkrax/csv_entry.rb
```diff
@@ -90,6 +86,14 @@ def build_metadata
       self.parsed_metadata
     end
 
+    def validate_record
+      raise StandardError, 'Record not found' if record.nil?
+      unless importerexporter.parser.required_elements?(record)
+        raise StandardError,
+              "Missing required elements, missing element(s) are: #{importerexporter.parser.missing_elements(keys_without_numbers(keys_with_values)).join(', ')}"
```
jeremyf, line 93: One adjustment is that the `required_elements?` now knows what the keys are that are missing.  And there's a duplication of that knowledge here.  But I definitely like this pathway.

Issue #760, opened 2023-03-16
EPIC: Introduce Concept and Handling of Existing Derivatives


jeremyf:
## Problem Statement

As people leverage Bulkrax to migrate from one solution to Hyrax, we are noting that many of them have existing derivatives that they would like to bring along.

For example, a curated thumbnail of a video; an already generated HOCR file, etc.

At present (<2023-03-16 Thu>), Bulkrax and Hyrax has no consideration for importing/leveraging existing derivatives.

This issue outlines a design overview and a draft task breakdown to address this problem statement.

## Design Overview

Add the following CSV columns to a `FileSet` entry:

-   **derivative\_thumbnail (arity 1):** When given this will be used as the thumbnail for the FileSet.
-   **derivative\_translation (arity many):** When given this will be added as translations for the FileSet.
-   **deriviatve\_hocr (arity 1):** When given this will be used as the HOCR file.

We likely will want to consider arbitrary suffixes.  We would store these suffixes on a child database table of the `Bulkrax::Entry` (e.g. `bulkrax_entry_derivatives`).  At present, we’re working with URLs, but may also want to consider path relative files as well (e.g. from the importer’s file).

Then during the `CreateDerivativesJob`, we would leverage a `DerivativeFileLocator` service.  The interface of a locator is:

-   Given a `FileSet` and a `derivative type` (e.g. thumbnail, hocr, translation)
-   Return (or yield) that `derivative file(s)`

For each `derivative type` we would register a set of strategies for locating the file.  The `DerivativeFileLocator` would try each strategy, breaking after the first successful locating.

One strategy would be: For the given `derivative type` check the `bulkrax_entry_derivatives` for a path/URL.  When it exists, download the file at the given URL and return that as the `derivative file`.

Another strategy would be to take the original file of the `FileSet` and generate the derivative.

## Task Breakdown [Draft]

-   [ ] Create `Bulkrax::EntryDerivative` that belongs to `Bulkrax::Entry`.  The fields would be: `entry_id` (as a foreign key), `derivative_type`, and `path`.
-   [ ] Adjust parser/entry to capture the `derivative_*` fields.  To consider: parser mappings with a `:from` directive.
-   [ ] Create the `Hyrax::DerivativeFileLocator`; this will likely live in Bulkrax but is generally applicable.
    -   Its public API method is `.call(file_set:, derivative_type:)` which returns an array of file name paths.  *Would it make sense instead to yield the paths?  In this way we’d yield once for each file.*
[...]

PR #761, opened 2023-03-16
Adding `Samvera::Derivatives` for interface discussion

jeremyf:
This commit begins to define an interface for the separation of concerns regarding the long-standing implementation of `Hydra::Derivatives` and `Hyrax::FileSetDerivativesService` as those relate to the existence of pre-existing derivatives.

This does not address the implementation details related to Bulkrax handling this work.  I would instead assume that Bulkrax would implement a subclass(es) `Samvera::Derivatives::FileLocator::Strategy`, which would in turn address the specific needs and implementation details.

As I've explored the implementations, a challenge of Hydra::Derivatives is that is combines the concern of finding the derivative (by making it) and applying that derivative to the corresponding file set.

Also, there is a higher level concern that is not addressed in this implementation; namely what are the derivatives of concern for a given `FileSet`; something that is likely based on the mime_type of the original file and metadata from the parent work.

Why not create another gem?

1. The immediate need is Bulkrax related.
2. The plumbing is laid for Bulkrax's test suite.
3. We probably don't need another gem to triangulate for a general solution.

That said, I am name spacing this interface in `Samvera` so if we find it useful we can extract a `samvera-derivatives` gem.  After all, not everyone's going to want Bulkrax.

Why not put this in Hyrax?

1. Because Bulkrax works across many different Hyrax versions.  Thus by being in Bulkrax it is generally more available than if it were in Hyrax; something which we'd need to backport.

An advantage of establishing the interface is that work can be done to now "fan out".  That is someone can work on a locator strategy while another works on the applicator strategy.

Related to:

- https://github.com/samvera-labs/bulkrax/issues/760
- https://github.com/scientist-softserv/utk-hyku/issues/343

### DIFF
--- a/lib/samvera/derivatives.rb
+++ b/lib/samvera/derivatives.rb
@@ -1,0 +1,126 @@
+# frozen_string_literal: true
+
+##
+# Why Samvera and not Hyrax?  Because there are folks creating non-Hyrax applications that will
+# almost certainly want to leverage this work.  And there might be switches for `defined?(Hyrax)`.
+#
+# Further the following four gems have interest in the interfaces of this module:
+#
+# - Hyrax
+# - Hydra::Derivatives
+# - Bulkrax
+# - IiifPrint
+#
+# As such, it makes some sense to isolate the module and begin defining interfaces.
+module Samvera
+  ##
+  # This module separates the finding/creation of a derivative binary (via {FileLocator}) and
+  # applying that derivative to the FileSet (via {FileApplicator}).
+  #
+  # @todo Consider how we might bundle/adapt the existing Hyrax::FileSetDerivativesService to serve
+  #       as both a locator and an applicator.  What would need to be true?  Conceptually it's a bit
+  #       challenging because the mapping of behavior is different.  However, it would be possible
+  #       to have the FileLocator fail to return any paths and then if the paths are nil, run the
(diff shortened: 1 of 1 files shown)


lib/samvera/derivatives.rb
```diff
@@ -0,0 +1,203 @@
...
+        # @note What's going on with this logic?  To continue to leverage
+        #       Hyrax::FileSetDerivativeService, we want to let that wrapped service (as a
+        #       FromLocation) to do it's original work.  However, we might have multiple strategies
+        #       in play for application.  That case is when we want to first check for an existing
+        #       thumbnail and failing that generate the thumbnail.  The from_location could either
+        #       be the found thumbnail...or it could be the wrapped Hyrax::FileSetDerivativeService
+        #       that will create the thumbnail and write it to the location.  The two applicator
+        #       strategies in that case would be the wrapper and logic that will write the found
+        #       file to the correct derivative path.
+        def write!
```
jeremyf, line 179: This may not be the best place for this switching logic; I suspect there's another class that's somewhat itching to emerge.

orangewolf:
I’m curious why this isn’t happening in iiif-print where we already massively modify the derivative process?

jeremyf:
Moving this to IIIF Print

PR #762, opened 2023-03-17
Adding `Bulkrax::EntryDerivative` model

jeremyf:
## Adding `Bulkrax::EntryDerivative` model

8ee3fe37c1097d3a365311e5a058f911c969ca37

This is preliminary modeling for imports to handle associating existing
derivatives with a `Bulkrax::Entry`.

Hyrax has implicit derivative_types that are not declared, but in part
assumed as a general list of derivatives.  In this model, we want to
begin name those derivatives to address the following use cases:

```gherkin
Given a FileSet
When I provide a thumbnail derivative
Then I want to add that as the thumbnail for the FileSet

Given a FileSet
When I do not provide a thumbnail derivative
Then I want to generate a thumbnail
And add the generated as the thumbnail for the FileSet
```

Related to:

- https://github.com/samvera-labs/bulkrax/pull/761
- https://github.com/samvera-labs/bulkrax/issues/760
- https://github.com/scientist-softserv/utk-hyku/issues/371

## Stubbing out derivatives locator/applicator logic

923c52756b90502a40bbe3d2f8cf79aa7398b03d

With this commit, I begin to take the work already done in [this PR][1]
and start pseudo-coding my approach.

All of this continues to circle around the configuration for a
downstream implementation.  However, it's drawing closer.

[1]: https://github.com/samvera-labs/bulkrax/pull/761

### DIFF
--- a/app/models/bulkrax/entry.rb
+++ b/app/models/bulkrax/entry.rb
@@ -16,6 +16,8 @@ module Bulkrax
     alias importer importerexporter
     alias exporter importerexporter
 
+    has_many :derivatives, dependent: :destroy
+
     serialize :parsed_metadata, Bulkrax::NormalizedJson
     # Do not serialize raw_metadata as so we can support xml or other formats
     serialize :collection_ids, Array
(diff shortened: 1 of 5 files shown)


jeremyf:
Closing this for now.

PR #764, opened 2023-03-20
Improve CreateRelationshipsJob's performance by setting reindexing to LIMITED_REINDEX

### LINKED ISSUE #758: Improve CreateRelationshipsJob Performance
In the `csv_parser`, each record is imported using it's own import job. In contrast, all the work of adding records to collections happens serially in `CreateRelationshipsJob`. 

One way to parallelize this would be to refactor `CreateRelationshipsJob` into `CreateRelationshipJob`. Move the logic from `CreateRelationshipsJob`'s `perform` into `ScheduleRelationshipsJob`'s `perform`, and use `CreateRelationshipsJob`'s `process` as `CreateRelationshipJob`'s `perform`. Each `CreateRelationshipJob` would only be responsible for adding the relationship for one record, not all the records in an import.

Or, a new job called `AddToCollectionJob` could be written:

~~~
module Bulkrax
  class AddToCollectionJob < ApplicationJob
    queue_as :import

    # rubocop:disable Rails/SkipsModelValidations
    def perform(child_record, parent_record)
      child_record.member_of_collections << parent_record
      child_record.save!
    end
end
~~~

then rewrite `CreateRelationshipsJob` to schedule `AddToCollectionJob`s in `process`. 

This would speed up record imports significantly.

cudevmaxwell:
This change matches the behavior in Hyrax's `CollectionsMembershipActor`, when a record is added through the UI.

Fixes #758.

### DIFF
--- a/app/jobs/bulkrax/create_relationships_job.rb
+++ b/app/jobs/bulkrax/create_relationships_job.rb
@@ -112,6 +112,7 @@ module Bulkrax
     end
 
     def add_to_collection(child_record, parent_record)
+      parent_record.try(:reindex_extent=, Hyrax::Adapters::NestingIndexAdapter::LIMITED_REINDEX)
       child_record.member_of_collections << parent_record
       child_record.save!
     end


laritakr:
I don't understand why this is necessary, given that the defaults are already set in Hyrax. They default to LIMITED unless it's necessary to do more.

cudevmaxwell:
Hmm... It looks like the 4.0.0 release will use a more efficient Solr graph query. That change was backported to 3.x before the 3.5.0 release. Instead of accepting this PR, maybe we should try setting HYRAX_USE_SOLR_GRAPH_NESTING in the env to 'true'?

laritakr (approved):

cudevmaxwell:
Our tests with `HYRAX_USE_SOLR_GRAPH_NESTING = true` showed a big boost in performance, even faster than adding the `LIMITED_REINDEX` fix. I'm leaning towards closing this PR without a merge. If folks are running into performance issues with Hyrax 3.5.0 and Bulkrax 5.1.0, they should enable HYRAX_USE_SOLR_GRAPH_NESTING like we did, or use the backport gem for older Hyraxi as suggested by @laritakr. (Thanks for reviewing this PR, btw).

PR #767, opened 2023-03-23
update reset rake task to switch tenants

ShanaLMoore:
When testing the previous implementation against a hyku app, the rake task failed because it wasn't able to switch into tenants properly.


orangewolf (changes requested):
putting a hold on this till we can talk about it

ShanaLMoore:
Closing this PR. I misunderstood that there is a [corresponding wrapper in hyku](https://github.com/samvera/hyku/blob/main/lib/tasks/reset.rake) that I should've used when testing out this task.

PR #770, opened 2023-03-28
Adding guard for querying 0 rows for FileSets

jeremyf:
Prior to this commit, when we would query to export file sets we didn't guard for no file sets associated with the works.  In [this importer][1] we saw a SOLR exception about an invalid query.  The SOLR query had the following:

```
fl=id&rows=0&q=has_model_ssim%3AFileSet+AND+id%3A%28%29+&qt=standard
```

In the above, we are querying for FileSets (selecting on the ID field). However there were no candidate file sets (e.g. `rows=0` and `id:()`) associated with the works for the export.

With this commit we guard against running a query for 0 rows.  Which should eliminate the particular problem.

Related to:

- https://github.com/scientist-softserv/britishlibrary/issues/353
- https://github.com/scientist-softserv/britishlibrary/pull/328
- https://github.com/samvera-labs/bulkrax/pull/759

[1]: https://qa.bl-staging.notch8.cloud/exporters/5?locale=en

### DIFF
--- a/app/parsers/bulkrax/parser_export_record_set.rb
+++ b/app/parsers/bulkrax/parser_export_record_set.rb
@@ -134,8 +134,15 @@ module Bulkrax
                          end
       end
 
+      # @note In most cases, when we don't have any candidate file sets, there is no need to query SOLR.
+      #
+      # @see Bulkrax::ParserExportRecordSet::Importer#file_sets
       def file_sets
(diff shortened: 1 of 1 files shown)


ShanaLMoore:
I tested this locally by reproducing the issue the client was seeing. 

I am confirming that these changes resolved the issue. 

The first export failed with a solr execption. The second one (which included the changes of this PR) succeeds.  

image

image

ShanaLMoore (approved):

PR #774, opened 2023-03-30
Adding quotes around candidate_file_set_ids

jeremyf:
Prior to this commit, we were not wrapping the file_sets_query in quotes.

If we look at other instances of querying on ID, we're wrapping the identifiers in quotes (e.g. `id:("ONE" OR "TWO" OR "THREE")`).

With this commit, we normalize this query to be the same as other similar queries.

One particular challenge is that Bulkrax does not have a Solr nor Fedora instance for testing.

I verified the fix by running, in the Rails console, the following:

```ruby
switch!('kew')
work_id = Article.first.id
file_sets_query = %(has_model_ssim:FileSet AND id:("0bd95381-0d3e-4df6-a0fc-8876e94369bc" OR "989a736d-17b7-472f-b5fd-d544af2b74f4" OR "#{work_id}"))
file_sets_query_kwargs = { fl: "id", method: :post, rows: 3 }
ActiveFedora::SolrService.query(file_sets_query, **file_sets_query_kwargs)
```

Screen Shot of the results of the above ruby code; hint it works

See: https://kew.iro.bl.uk/exporters/146?locale=en

Related to:

### DIFF
--- a/app/parsers/bulkrax/parser_export_record_set.rb
+++ b/app/parsers/bulkrax/parser_export_record_set.rb
@@ -141,7 +141,7 @@ module Bulkrax
         @file_sets ||= if candidate_file_set_ids.empty?
                          []
                        else
-                         ActiveFedora::SolrService.query(file_sets_query, **file_set_query_kwargs)
+                         ActiveFedora::SolrService.query(file_sets_query, **file_sets_query_kwargs)
                        end
       end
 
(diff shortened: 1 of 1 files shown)


ShanaLMoore:
I also added patch-ver label. For some reason, bug-fix doesn't seem to appease the label checker. a TODO for later!

ShanaLMoore:
> I also added patch-ver label. For some reason, bug-fix doesn't seem to appease the label checker. a TODO for later!

Actually, I already found it. That label needs to be added to [this file](https://github.com/samvera-labs/bulkrax/blob/main/.github/release.yml).

Issue #775, opened 2023-03-30
Introduce pagination for File Set Export Query


jeremyf:
At present if we are querying with more than 1024 "OR" clauses in a field select, we receive a 400 response from the server.

The ids size is 1196.  When using the following code, with a limit of IDs 0 through 1023 the query works.  When we try 0 to 1024, the query fails.

```ruby
file_sets_query = "has_model_ssim:FileSet AND "
# ;; 1024 entries
# We cannot request more than 1024 IDs
# We currently set maxBooleanClauses to 1024,
ids = File.read(File.expand_path("ids.txt", __dir__)).split.map(&:strip)[0..1023]
file_sets_query += "id:(\"" + ids.join('" OR "') + "\")"
file_sets_query_kwargs = { fl: "id", method: :post, rows: ids.size }
Account.first.switch!
ActiveFedora::SolrService.query(file_sets_query, **file_sets_query_kwargs)
```

See https://solr.apache.org/guide/8_1/format-of-solr-xml.html#global-maxbooleanclauses

PR #776, opened 2023-03-30
Addressing SOLR maxBooleanClauses for WorkType export

jeremyf:
Prior to this commit, we were querying all of the candidate file set IDs for the associated work type.  In the case of the below exporter, we had 1196 IDs to query.

As we were testing if there was a query size limit, we noticed that we could query 1024 IDs but not 1025.  We add other query parameters, to bump the query size, but at 1024 IDs things worked.  At 1025 they failed.

Digging further, we saw that there is hard (yet configurable) limit for SOLR query OR clauses.  In our case, 1024.  (Hint: if you powers of 2, dig a little).

With this commit, we instead request 512 IDs at a time and join those results into a single array.

Note: there are other queries that will need similar fixing but for the time being, this (we hope) resolves a client urgency.

See: https://kew.iro.bl.uk/exporters/146?locale=en
See: https://solr.apache.org/guide/8_1/query-settings-in-solrconfig.html#query-sizing-and-warming

Related to:

- https://github.com/samvera-labs/bulkrax/issues/775

### DIFF
--- a/app/parsers/bulkrax/parser_export_record_set.rb
+++ b/app/parsers/bulkrax/parser_export_record_set.rb
@@ -134,17 +134,12 @@ module Bulkrax
                          end
       end
 
+      SOLR_QUERY_PAGE_SIZE = 512
+
       # @note In most cases, when we don't have any candidate file sets, there is no need to query SOLR.
       #
       # @see Bulkrax::ParserExportRecordSet::Importer#file_sets
-      def file_sets
-        @file_sets ||= if candidate_file_set_ids.empty?
-                         []
-                       else
-                         ActiveFedora::SolrService.query(file_sets_query, **file_sets_query_kwargs)
-                       end
-      end
-
+      #
       # Why can't we just use the candidate_file_set_ids?  Because Hyrax is pushing child works into the
       # `file_set_ids_ssim` field.
       #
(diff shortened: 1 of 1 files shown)


app/parsers/bulkrax/parser_export_record_set.rb
```diff
@@ -134,30 +134,34 @@ def collections
...
+                         candidate_file_set_ids.each_slice(SOLR_QUERY_PAGE_SIZE) do |ids|
+                           fsq = "has_model_ssim:#{Bulkrax.file_model_class} AND id:(\"" + ids.join('" OR "') + "\")"
+                           fsq += extra_filters if extra_filters.present?
+                           results += ActiveFedora::SolrService.query(
+                             fsq,
+                             { fl: "id", method: :post, rows: ids.size }
+                           )
+                         end
+                         results
+                       end
```
summer-cook, line 164: this makes sense to me based on the PR description. 

the spacing/indentation of lines 152-164 looks off for me.. not sure if its just github being weird though

summer-cook (approved):

app/parsers/bulkrax/parser_export_record_set.rb
jeremyf: The indentation is atypical due to paranthesis placement; however all of this appeases Rubocop.

Issue #777, opened 2023-04-04
Bulkrax fails to recognize an imported zip file as zip


jeremyf:
The following code, in both the [application_parser](https://github.com/samvera-labs/bulkrax/blob/d14d55fbf4a9b6c7a3275346725da9b853236287/app/parsers/bulkrax/application_parser.rb#L367-L369) and [importer_exporter_behavior](
https://github.com/samvera-labs/bulkrax/blob/d14d55fbf4a9b6c7a3275346725da9b853236287/app/models/concerns/bulkrax/importer_exporter_behavior.rb#L53-L55) fails to recognize a zip file.  This was introduced in Introduced in #744.

```ruby
def zip?
  parser_fields&.[]('import_file_path') && ::Marcel::MimeType.for(parser_fields['import_file_path']).include?('application/zip')
end
```

The prior implementation had the following logic:

```ruby
def zip?
  parser_fields&.[]('import_file_path') && MIME::Types.type_for(parser_fields['import_file_path']).include?('application/zip')
end
```

That logic worked.

With this change, the importer worked.

```ruby
def zip?
  parser_fields&.[]('import_file_path') && ::Marcel::MimeType.for(File.open(parser_fields['import_file_path'])).include?('application/zip')
end
```

PR #744, opened 2023-02-17
decouple from hyrax

labradford:

PR #778, opened 2023-04-04
HotFix: Patching Marcel MimeType sniffing

jeremyf:
Note: This does not included tests, which is something that #744  also did not.  However, we have verified the behavior.

Reading the [documentation for the Marcel gem][1], it requires a file handle (e.g. `File.new` or `File.open`) instead of a String.

When given a `String` for `::Marcel::MimeType.for` the returned value is `application/octet-stream`.

Related to:

- https://github.com/samvera-labs/bulkrax/issues/777

[1]: https://github.com/rails/marcel


laritakr (approved):

alishaevn:
👍🏾

PR #779, opened 2023-04-04
Adding spec to verify zip? detection

jeremyf:
Prior to this commit, we had just changed the logic to resolve issue #777 (without tests).  We had verified the behavior in downstream implementations.

With this commit, I have done two things:

1. Introduced tests on the `#zip?` method
2. Removed duplication of logic in favor of delegation.

Closes https://github.com/samvera-labs/bulkrax/issues/777

Related to:

- https://github.com/samvera-labs/bulkrax/pulls/778


app/parsers/bulkrax/application_parser.rb
```diff
@@ -358,16 +359,6 @@ def zip
...
-    # Is this a file?
-    def file?
-      parser_fields&.[]('import_file_path') && File.file?(parser_fields['import_file_path'])
-    end
-
-    # Is this a zip file?
-    def zip?
-      parser_fields&.[]('import_file_path') && ::Marcel::MimeType.for(File.new(parser_fields['import_file_path'])).include?('application/zip')
-    end
-
```
bkiahstroud, line 370: @jeremyf should we be `raise`ing errors here instead or removing them like how we do in [#records](https://github.com/samvera-labs/bulkrax/blob/main/app/parsers/bulkrax/application_parser.rb#L64), for example?

jeremyf: No…we need the method to be on parser, but are relying on the `delegate` declaration to add that method.

In the case of a `NotImplementedError` that is usually used for an `@abstract` class to signify that a descendant should implement that method.

PR #781, opened 2023-04-06
move config to a Configurable class

no-reply:
the older style config setup is incompatible with Rails 6, preventing support for Hyrax 4.

this is a minimal pass at extracting the config and keeping the current API.

----

there's a RuboCop rule disable in this PR:

there's some debate about the value of this style guide rule at
https://github.com/rubocop/ruby-style-guide/issues/556.

it seems in our case like using `module_function` is going to be prohibitively
hard, and i think the longer term desire is to remove both `extend Forwardable`
and `extend self`, expecting callers to just call `#config` directly. the
behavior around defaults needs to be cleaned up to make this possible, but it
seems like a bad plan to bend this implementation around (unhelpful?) style
rules for code we'd rather work on removing.

### DIFF
--- a/lib/bulkrax.rb
+++ b/lib/bulkrax.rb
@@ -6,35 +6,91 @@ require 'active_support/all'
 
 # rubocop:disable Metrics/ModuleLength
 module Bulkrax
-  class << self
-    # @todo Move from module attribute methods to a configuration class.  With module attributes,
-    #       when we make a change we are polluting the global space.  This means that our tests that
-    #       modify these config values are modifying global state.  Which is not desirous, as it can
-    #       introduce unexpected flakey tests.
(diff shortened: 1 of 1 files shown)


orangewolf (approved):

orangewolf:
@no-reply I think this is a minor enhancement level change (other than its required for Rails 6 support). It doesn't appear to change the API at all so I do not think it needs to be in a major release. Does that jive with your understanding? If not, can you mod the labels? if so, feel free to merge it

orangewolf (approved):

no-reply:
yes, agreed. I kept this separate from the rest of the upgrade because i
was (relatively) sure it's non-breaking

PR #782, opened 2023-04-06
add support for Rails 6, Hyrax 4, and Blacklight 7

no-reply:
depends on #781

this makes no attempt to ensure UI-layer support for BL7 and whatever bootstrap
versions it requires (excepting whatever is tested by RSpec). the scope here is
simply to make it *possible* for a Hyrax 4 application to install and use
Bulkrax.


orangewolf (commented):
this looks good to me so far

jeremyf (approved):

PR #783, opened 2023-04-07
Non hyrax exports

sephirothkod:
sets up exports for non hyrax systems.


no-reply:
just a heads up that after the JS changes in this, i'm no longer seeing the export form unhide the appropriate options.

I'm investigating, but wondered if anyone else could reproduce?

no-reply:
looks like this was fixed in https://github.com/samvera-labs/bulkrax/pull/791

PR #790, opened 2023-04-18
Fix multi_value_element_split_on example regex

mcritchlow:
This makes the default `multi_value_element_split_on` regex valid, should someone simply decide to uncomment it as-is.


kirkkwang:
Thanks for fixing that, Matt!

kirkkwang (approved):

PR #791, opened 2023-04-18
Set turbolinks to false for the exporter index page

kirkkwang:
This commit sets turbolinks to false for exporters because sometimes the form page does not load the correct fields when choosing something from `Export From`.

In this screenshot, under the `Export From` _should_ have another field where you search for the collection, but it doesn't... 
image

### DIFF
--- a/app/views/bulkrax/exporters/index.html.erb
+++ b/app/views/bulkrax/exporters/index.html.erb
@@ -1,7 +1,7 @@
 <% provide :page_header do %>
   <h1><span class="fa fa-cloud-download" aria-hidden="true"></span> Exporters</h1>
   <div class="pull-right">
-    <%= link_to new_exporter_path, class: 'btn btn-primary' do %>
+    <%= link_to new_exporter_path, class: 'btn btn-primary', data: { turbolinks: false } do %>
       <span class="fa fa-edit" aria-hidden="true"></span> <%= t(:'helpers.action.exporter.new') %>
     <% end %>
   </div>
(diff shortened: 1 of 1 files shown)


bkiahstroud:
@ShanaLMoore I could be wrong, but not having turbolinks might break the jQuery DataTables on the exporter index page. 

Though the way I'm reading this code, it's not disabling turbolinks **on** the exporter index page, it's disabling them **from** the index page. I.e. when navigating to the form from the index page, it sends a param to disable turbolinks on that page without actually disabling it on the index page. Am I reading that correctly?

ShanaLMoore (changes requested):
cc @kirkkwang Could you confirm and address @bkiahstroud concerns before merging this?

bkiahstroud:
> @bkiahstroud It's my understanding that this should only impact the form elements.

Great, that's completely fine with me. Mostly read "for the exporter index page" in the title and wanted to double check 😅

kirkkwang:
https://share.getcloudapp.com/NQuW5E6y

@bkiahstroud is this the functionality you're talking about?  As you can see the turbolinks are false, but i can still sort the table

bkiahstroud:
@kirkkwang yes, as long as that table is still working as expected than we should be good to go (sort, search, pagination, etc.)

kirkkwang:
Great thanks @bkiahstroud for keeping that in mind!  @ShanaLMoore good to merge?

no-reply:
i'm not sure if it's known, but it looked to me like this issue was introduced in https://github.com/samvera-labs/bulkrax/pull/783#issuecomment-1516657037

it's possible there's a fix available without turning off turbolinks. at least for now, i'm very happy with this, and am confirming that it fixes the issue we've seen on the `surfliner` side.

PR #803, opened 2023-05-05
do-not-duplicate-source-identifier-on-export

alishaevn:
- ref: https://github.com/scientist-softserv/atla_digital_library/issues/295

# story
when exporting from ATLA on the `v4-4-patch` branch, we would have an `identifier` column **and** `identifier_1` column. one would be an array, while the other would be a string

# expected behavior
- Do not duplicate source identifier on export 
- backport `#build_system_metadata fix` from `v5.0.0` (#712)

# demo
image

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -107,7 +107,9 @@ module Bulkrax
     # Metadata required by Bulkrax for round-tripping
     def build_system_metadata
       self.parsed_metadata['id'] = hyrax_record.id
-      self.parsed_metadata[source_identifier] = hyrax_record.send(work_identifier)
+      source_id = hyrax_record.send(work_identifier)
+      source_id = source_id.to_a.first if source_id.is_a?(ActiveTriples::Relation)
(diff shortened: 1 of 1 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -149,7 +151,7 @@ def build_mapping_metadata
       mapping = fetch_field_mapping
       mapping.each do |key, value|
         # these keys are handled by other methods
-        next if ['model', 'file', related_parents_parsed_mapping, related_children_parsed_mapping].include?(key)
+        next if ['model', 'file', related_parents_parsed_mapping, related_children_parsed_mapping, source_identifier].include?(key)
```
alishaevn, line 154: we're already adding the `source_identifier` to `self.parsed_metadata` in `#build_system_metadata`

alishaevn:
@jeremyf @sephirothkod @bkiahstroud I plan to add the change from `#build_mapping_metadata` into the `main` branch of bulkrax too.

any objections?

aprilrieger (approved):

bkiahstroud (approved):

Issue #804, opened 2023-05-10
Create Relationships Job needs a record lock


laritakr:
`Hyrax::Actors::OrderedMembersActor` clears ordered members, so another job could interfere with the relationships we are in process of forming. We need to acquire a lock on the parent record before adding relationships. Note that it is probably unnecessary to re-save all of the child works... updating the index is probably adequate. However if the children are saved, locks should also be acquired for each of them.

Use [Hyrax's lockable module around this code](https://github.com/samvera-labs/bulkrax/blob/main/app/jobs/bulkrax/create_relationships_job.rb#L65-L77) (look at the ordered members actor for an example) to prevent the potential for another job causing lost relationships.

Related to 
- https://github.com/scientist-softserv/iiif_print/issues/233 (but far less likely to occur in Bulkrax than in IiifPrint)
- https://github.com/samvera-labs/bulkrax/issues/810

ShanaLMoore:
As shared in slack, I went to an interesting railsconf talk about race conditions. It covered locking options such as:
[redlock-rb](https://github.com/leandromoreira/redlock-rb), [with_advisory_lock](https://github.com/ClosureTree/with_advisory_lock), an active record [options](https://thelazylog.com/understanding-locking-in-rails-activerecord/) (which from my photos looks like all he did was re find a record ??) and [and_wrap_original](https://www.rubydoc.info/github/rspec/rspec-mocks/RSpec%2FMocks%2FMessageExpectation:and_wrap_original) to mimic a race condition in the specs.

[Hyrax](https://github.com/samvera/hyrax/blob/main/app/services/hyrax/lock_manager.rb#L2) uses redlock.

laritakr:
This is complicated by the error handling - if we don't get a lock, we get an error on the parent and aren't going to report each individual pending relationship's error. A refactor is needed.

PR #811, opened 2023-05-18
🐛 Add record locking for create_relationships_job

laritakr:
With this commit we're adding two major concepts:

- Conditional locking of records (when locking is available)
- Checking that a parent_record exists before proceeding on further
  checks/tests that could raise errors

Further, I've taken the liberty to move the increment failed
relationships out of the general error handling and instead to happen
only if we had a parent record; with the assumption that when we don't
have a parent, we only want to report one error instead of reporting one
error per child.

Closes: #804

- https://github.com/samvera-labs/bulkrax/issues/804

Co-authored-by: LaRita Robinson <larita@scientist.com>

### DIFF
--- a/app/jobs/bulkrax/create_relationships_job.rb
+++ b/app/jobs/bulkrax/create_relationships_job.rb
@@ -64,40 +64,78 @@ module Bulkrax
       @parent_record_members_added = false
       @child_members_added = []
 
-      ActiveRecord::Base.uncached do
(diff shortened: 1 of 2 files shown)


laritakr (commented):
I approve it, but since I started the work, I feel like official approval isn't my decision.

kirkkwang (approved):

PR #817, opened 2023-06-13
:gift: temp branch to allow ADL to bypass fedora updates and import r…

ShanaLMoore:
…elationships only

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -41,6 +41,7 @@ module Bulkrax
       @source_identifier_value = source_identifier_value
       @klass = klass || Bulkrax.default_work_type.constantize
       @importer_run_id = importer_run_id
+      @importer = Bulkrax::ImporterRun.find(importer_run_id).importer
     end
     # rubocop:enable Metrics/ParameterLists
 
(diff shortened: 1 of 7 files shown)


app/factories/bulkrax/object_factory.rb
```diff
@@ -70,6 +71,12 @@ def run!
     end
 
     def update
+      if @importer.import_relationships_only
+        logger = Rails.logger
+        logger.info("Skipping #{klass} #{object.id} (#{Array(attributes[work_identifier]).first})")
```
jeremyf, line 76: This is great to have logging; wondering if there's more information to include.  `"Skipping updating metadata for #{klass} ID=#{object.id} (#{Array(attributes[work_identifier]).first})"`

app/views/bulkrax/importers/_form.html.erb
```diff
@@ -23,6 +23,11 @@
 
   <%= form.input :parser_klass, collection: Bulkrax.parsers.map {|p| [p[:name], p[:class_name], {'data-partial' => p[:partial]}]}, label: "Parser", input_html: { class: 'form-control' } %>
 
+  <%= form.input :import_relationships_only?,
```
jeremyf, line 26: I think you likely won't want the trailing `?` ; I don't know if it will attempt to cast that method to `:import_relationships?=`

ShanaLMoore, line 26: This should only get displayed when the user selects a CVS parser.

PR #835, opened 2023-07-12
Fix modal for hydra apps

summer-cook:
# Summary
The update importer modal was not opening in wvu hydra, likely because it used a regular link instead of a button. The button_to rails tag also didn't work as expected, so it needed to just render as a normal button. This PR fixes it so it will open in all apps

**Note**: this was tested in both wvu hydra & atla hyrax to ensure it works across both

# Related
- https://github.com/scientist-softserv/west-virginia-university/issues/65

# Video
https://share.getcloudapp.com/Z4uG1JEd

# Acceptance criteria
- As a user, i can click the update importer button from the update importer form. It opens to a modal, where i have several options to select how I would like to update my importer.

### DIFF
--- a/app/views/bulkrax/importers/edit.html.erb
+++ b/app/views/bulkrax/importers/edit.html.erb
@@ -10,7 +10,9 @@
         <%= render 'form', importer: @importer, form: form %>
         <div class="panel-footer">
           <div class='pull-right'>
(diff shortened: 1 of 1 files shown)


kirkkwang (approved):
Very cool, thanks for fixing!

PR #837, opened 2023-07-19
🎁 Add alt and title attributes

kirkkwang:
This commit will add the `alt` and `title` attributes to the list tags so when users hover over the links they will see the name of the link on the dashboard.

image
image

### DIFF
--- a/app/views/hyrax/dashboard/sidebar/_bulkrax_sidebar_additions.html.erb
+++ b/app/views/hyrax/dashboard/sidebar/_bulkrax_sidebar_additions.html.erb
@@ -1,10 +1,14 @@
 <% if current_ability.can_import_works? %>
-  <%= menu.nav_link(bulkrax.importers_path) do %>
+  <%= menu.nav_link(bulkrax.importers_path,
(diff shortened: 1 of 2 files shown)


app/views/hyrax/dashboard/sidebar/_bulkrax_sidebar_additions.html.erb
```diff
@@ -1,10 +1,14 @@
 <% if current_ability.can_import_works? %>
-  <%= menu.nav_link(bulkrax.importers_path) do %>
+  <%= menu.nav_link(bulkrax.importers_path,
+                    alt: t('bulkrax.admin.sidebar.importers'),
```
jeremyf, line 3: I'm unclear why we have both `alt` and `title` tag?  This element looks to render as a link, without an image.  Thus `title` should be adequate.  If it were an image, then we would want an `alt` tag.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/alt

jeremyf (changes requested):

PR #839, opened 2023-07-20
Prefer [] over [""] for blank values that are split

orangewolf:


summer-cook (approved):

bkiahstroud:
```ruby
irb(main):002:0> [].blank?
=> true
irb(main):003:0> [""].blank?
=> false
```

This ^ is the only potential issue with this change that I can think of off the top of my head; if a project is doing presence checks on "empty" values and that logic depends on the original behavior.

jeremyf:
This may offer a partial solution to:

- https://github.com/scientist-softserv/palni-palci/issues/624

PR #840, opened 2023-07-20
Move js require higher in the install generator

kirkkwang:
# Summary
In order to make Bulkrax's javascript work in Hydra, we needed to add bootstrap's JS to bulkrax, along with jquery since bootstrap depends on jquery. 
In hyku apps, this caused unintended behavior with the dataTables in the dashboard. 
The data tables needed to be required AFTER bulkrax in order to keep their functionality in hyku.

This commit will alter the install generator to insert the require for `bulkrax/application` above dataTables require. 

co-author: @summer-cook 

Ref:
  - https://github.com/samvera-labs/bulkrax/pull/836

# Screenshots

<details><summary>generated require</summary>
image

</details>

# Note
This may need to be changed for existing hyku applications with Bulkrax. 
adding bootstrap javascript was a potential breaking change for some applications: existing hyku applications will need to change the order of these requires.

### DIFF
--- a/lib/generators/bulkrax/install_generator.rb
+++ b/lib/generators/bulkrax/install_generator.rb
@@ -47,11 +47,21 @@ class Bulkrax::InstallGenerator < Rails::Generators::Base
   def add_javascripts
     file = 'app/assets/javascripts/application.js'
     file_text = File.read(file)
-    js = '//= require bulkrax/application'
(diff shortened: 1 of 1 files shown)


kirkkwang:
Instead of tinkering the generator, we are opting to adjust Hyda's application.js to include bootstrap instead of having it in Bulkrax

PR #842, opened 2023-07-25
🐛 Remove navtabs.js

kirkkwang:
This commit will revert a previous change introducing the navtabs.js file to make the nav tabs work for Hydra because it inadvertently broke things for Hyku.


kirkkwang:
favoring modifying instead of removing #843

PR #844, opened 2023-07-28
:bug: Add lock to child record

ShanaLMoore:
We are adding a lock to child record due to odd
race-condition-like behaviors seen when collections are forming relations. Reindexing would fail because #ordered_works is not returning anything.

- Issue: https://github.com/scientist-softserv/utk-hyku/issues/500
- Related to: https://github.com/samvera-labs/bulkrax/issues/804


orangewolf (approved):

ShanaLMoore:
Closing this PR because this turned out to not be related to the issue as expected. When tested using this branch the errors still occurred.

Issue #829, opened 2023-07-08
Valkyrization - Check for any miscellaneous direct references to ActiveFedora


bkiahstroud:
# Acceptance Criteria 

- [ ] Rework all references to `ActiveFedora` 

# Implementation Details 

TODO pending discussion of how to Valkyrize Bulkrax

PR #853, opened 2023-08-15
find existing records more often, leading to better round tripping experiance

orangewolf:
ActiveFedora use to find objects with {field => value}. Now it only reliably works with {solr_name(field) => value} so we use that instead. Also if find by id fails we should _still_ check source_identifier to see if we can find it that way.

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Bulkrax
-  class ObjectFactory
+  class ObjectFactory # rubocop:disable Metrics/ClassLength
     extend ActiveModel::Callbacks
     include Bulkrax::FileFactory
     include DynamicRecordLookup


app/factories/bulkrax/object_factory.rb
```diff
@@ -102,7 +103,8 @@ def find_or_create
     end
 
     def search_by_identifier
-      query = { work_identifier =>
+      work_index = ::ActiveFedora.index_field_mapper.solr_name(work_identifier, :facetable)
```
bkiahstroud, line 106: 1. Will `work_identifier` always be `:facetable`? Is there a chance this will different project-to-project, and thus fail sometimes? 
2. Is there a simple way to achieve this without calling `ActiveFedora` directly? With [Valkyrization](https://github.com/samvera-labs/bulkrax/projects/3) on the horizon, using `ActiveFedora` directly is going the way of the dinosaur. If the answer is "no" that's fine, I don't think it should block this PR; [this ticket](https://github.com/samvera-labs/bulkrax/issues/829) will rework this line later

orangewolf: I think the answer with both is to add a new config variable to bulkrax with a default that just is `"#{work_identifier}_ssim"`

alishaevn (approved):

app/factories/bulkrax/object_factory.rb
alishaevn: I'm reverting this change because it's breaking my ability to import FileSet's using the row.

ref: https://github.com/samvera-labs/bulkrax/issues/867

PR #857, opened 2023-08-28
add hyrax version to the test matrix

orangewolf:


orangewolf:
a good idea, but obsolete

PR #881, opened 2023-10-30
Reduce SQL calls when incrementing/decrementing run counters

bkiahstroud:
# Story

SQL logs before changes (using `#increment!`):

```sql
Bulkrax::ImporterRun Load (8.9ms)  SELECT  "bulkrax_importer_runs".* FROM "bulkrax_importer_runs" WHERE "bulkrax_importer_runs"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
Bulkrax::ImporterRun Update All (2.5ms)  UPDATE "bulkrax_importer_runs" SET "processed_records" = COALESCE("processed_records", 0) + 1 WHERE "bulkrax_importer_runs"."id" = $1  [["id", 1]]
```

SQL logs after changes (using `#increment_counter`):

```sql
Bulkrax::ImporterRun Update All (6.5ms)  UPDATE "bulkrax_importer_runs" SET "processed_records" = COALESCE("processed_records", 0) + 1 WHERE "bulkrax_importer_runs"."id" = $1  [["id", 1]]
```

The `SELECT` statement serves no purpose since the `UPDATE` statement is atomic.

I haven't done any benchmarks on these changes, but since less SQL statements seems like a pretty obvious win and since we update counters _a ton_, I'm hoping we'll see at least some speed improvements

### DIFF
--- a/app/jobs/bulkrax/create_relationships_job.rb
+++ b/app/jobs/bulkrax/create_relationships_job.rb
@@ -97,9 +97,11 @@ module Bulkrax
       end
 
       if errors.present?
-        # rubocop:disable Rails/SkipsModelValidations
-        importer_run.increment!(:failed_relationships, number_of_failures)
-        # rubocop:enable Rails/SkipsModelValidations
+        ImporterRun.connection.execute(<<-SQL)
+          UPDATE bulkrax_importer_runs
+          SET failed_relationships = COALESCE(failed_relationships, 0) + #{number_of_failures}
+          WHERE id = #{importer_run_id}
+        SQL
 
         parent_entry&.set_status_info(errors.last, importer_run)
 
@@ -107,9 +109,11 @@ module Bulkrax
         reschedule({ parent_identifier: parent_identifier, importer_run_id: importer_run_id })
         return false # stop current job from continuing to run after rescheduling
       else
-        # rubocop:disable Rails/SkipsModelValidations
-        Bulkrax::ImporterRun.find(importer_run_id).increment!(:processed_relationships, number_of_successes)
-        # rubocop:enable Rails/SkipsModelValidations
+        ImporterRun.connection.execute(<<-SQL)
+          UPDATE bulkrax_importer_runs
+          SET processed_relationships = COALESCE(processed_relationships, 0) + #{number_of_successes}
+          WHERE id = #{importer_run_id}
+        SQL
       end
     end
     # rubocop:enable Metrics/MethodLength
--- a/app/jobs/bulkrax/create_relationships_job.rb
+++ b/app/jobs/bulkrax/create_relationships_job.rb
@@ -97,7 +97,9 @@ module Bulkrax
       end
 
       if errors.present?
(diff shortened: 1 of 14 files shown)


app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -97,19 +97,23 @@ def perform(parent_identifier:, importer_run_id:) # rubocop:disable Metrics/AbcS
...
         parent_entry&.set_status_info(errors.last, importer_run)
 
         # TODO: This can create an infinite job cycle, consider a time to live tracker.
         reschedule({ parent_identifier: parent_identifier, importer_run_id: importer_run_id })
         return false # stop current job from continuing to run after rescheduling
       else
-        # rubocop:disable Rails/SkipsModelValidations
-        Bulkrax::ImporterRun.find(importer_run_id).increment!(:processed_relationships, number_of_successes)
-        # rubocop:enable Rails/SkipsModelValidations
+        ImporterRun.connection.execute(<<-SQL)
```
jeremyf, line 112: Should this call `ImporterRun.increment_counter(:processed_relationships, importer_run_id, amount: number_of_successes)` ?

I don't like the raw SQL so far from the model.

jeremyf: Oops…the increment_counter is a Rails method.  Hmm.

jeremyf: Screenshot 2023-10-31 at 12 42 42 PM

Wondering about using the `update_counters` method: https://api.rubyonrails.org/classes/ActiveRecord/CounterCache/ClassMethods.html#method-i-update_counters

bkiahstroud: `#update_counters` seems good to me. Same SQL, just gives you control over the amount, which is nice. 

I'm curious, why do you not like raw SQL?

orangewolf: does this raw sql work in mysql? all versions of postgres? msql? oracle db? sqlite? because the old version does and that's the reason we avoid raw sql whenever we can.

app/jobs/bulkrax/import_file_set_job.rb
```diff
@@ -21,14 +21,14 @@ def perform(entry_id, importer_run_id)
       entry.build
       if entry.succeeded?
         # rubocop:disable Rails/SkipsModelValidations
-        ImporterRun.find(importer_run_id).increment!(:processed_records)
-        ImporterRun.find(importer_run_id).increment!(:processed_file_sets)
+        ImporterRun.increment_counter(:processed_records, importer_run_id)
```
orangewolf, line 24: these are great

app/jobs/bulkrax/create_relationships_job.rb
bkiahstroud: Replaced the raw SQL with `#update_counters` and CI is still green so I think we're good!

jeremyf (approved):

PR #883, opened 2023-11-07
fix syntax error in ERB

dunn:
not sure how this one got through CI


kirkkwang (approved):
Thanks for finding this!

Issue #866, opened 2023-10-03
`ObjectFactory#search_by_identifier` bug


alishaevn:
# story
when using the default `source_identifier` value on `v5.4.0`, it does not return the correct `work_index` value in `ObjectFactory#search_by_identifier`. it returns `source_sim` while according to solr, the proper value should be `source_tesim`.
therefore a work is not found. this means that on the bulkrax entries show page, the "Item Link" says "Item has not yet been imported successfully". however, viewing the works page shows that the work **was** created.

ref: https://github.com/samvera/active_fedora/blob/v13.3.0/lib/active_fedora/indexing/field_mapper.rb#L29-L39

``` ruby
def search_by_identifier
  work_index = ::ActiveFedora.index_field_mapper.solr_name(work_identifier, :facetable)
  query = { work_index =>
            source_identifier_value }
  # Query can return partial matches (something6 matches both something6 and something68)
  # so we need to weed out any that are not the correct full match. But other items might be
  # in the multivalued field, so we have to go through them one at a time.
  match = klass.where(query).detect { |m| m.send(work_identifier).include?(source_identifier_value) }
  return match if match
end
```
# screenshots
<details>
<summary>view here</summary>

<img width="1343" alt="Screenshot 2023-10-03 at 5 53 48 PM" src="https://github.com/samvera-labs/bulkrax/assets/29032869/27e85b50-ed7e-4965-a0ab-c11d260b7444">
<img width="1373" alt="Screenshot 2023-10-03 at 5 56 15 PM" src="https://github.com/samvera-labs/bulkrax/assets/29032869/1219dd8d-614e-4cde-8744-ad63347b1b4d">
</details>

# acceptance criteria
- [ ]

PR #884, opened 2023-11-09
make search string used to look up objects configurable

### LINKED ISSUE #867: fix the `::ActiveFedora.index_field_mapper.solr_name` search
# story
in `v5.4.0` we are unable to import a `FileSet` in a csv while using the `FileSet` row. at least while using the default `work_identifier` value of `:source`. it keeps returning a `MissingParentError` error.

The root of the problem comes from `ObjectFactory#search_by_identifier`. commit [6473259](https://github.com/samvera-labs/bulkrax/pull/853/commits/64732597a02b47d5058797c7d4eebe5aeff40dfd) changed the way we find the work. that change has also caused another issue (ref: #866). we reach this method from `ImportFileSetJob#find_parent_record`.

I will be reverting this change in order to have working functionality for the samvera training workshop tomorrow. creating this ticket however so that we can figure out how to make the intended change work, once we have more time to look into it.

# demo
in solr the version is actually `source_tesim`, not `source_ssim` (the intention). however, the call to AF gives us `source_sim`, which is still wrong.

<details>
<summary>screenshots</summary>

Screenshot 2023-10-04 at 9 17 37 AM
image
</details>

orangewolf:
# Release Notes
This PR requires anyone with custom code that creates an ObjectFactory or that overrides the ObjectFactory#initializer or ObjectFactory#search_by_identifier methods to update their code to take in to account the new `work_identifier_search_field`

Resolves #867 and #866

### DIFF
--- a/lib/tasks/bulkrax_tasks.rake
+++ b/lib/tasks/bulkrax_tasks.rake
@@ -124,7 +124,7 @@ namespace :bulkrax do
   end
 
   def make_new_exports
-    Bulkrax::Exporter.all.each { |e| Bulkrax::ExporterJob.perform_later(e.id) }
+    Bulkrax::Exporter.find_each { |e| Bulkrax::ExporterJob.perform_later(e.id) }
   rescue => e
     puts "(#{e.message})"
   end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,9 @@
 # Bixby coding conventions
+require: rubocop-factory_bot
+
 # Added AllCops config for further modification
 AllCops:
-  TargetRubyVersion: 2.6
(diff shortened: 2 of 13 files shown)


lib/tasks/bulkrax_tasks.rake
```diff
@@ -124,7 +124,7 @@ namespace :bulkrax do
   end
 
   def make_new_exports
-    Bulkrax::Exporter.all.each { |e| Bulkrax::ExporterJob.perform_later(e.id) }
+    Bulkrax::Exporter.find_each { |e| Bulkrax::ExporterJob.perform_later(e.id) }
```
jeremyf, line 127: ❤️

jeremyf (approved):

PR #891, opened 2023-12-11
🐛 Fix #work_identifier_search_field logic

jeremyf:
The chaining logic for was grabbing the first letter of a single search_field value; instead of grabbing the first value in the 'search_field' object.

To fix the problem, we cast the value of `search_field` into an array and then grab the first array element.

Related to:

- https://github.com/samvera/bulkrax/issues/890


ShanaLMoore (approved):
Tested from: https://github.com/scientist-softserv/utk-hyku/pull/577/files

PR #895, opened 2024-01-17
🎁 Add `Bulkrax.persistence_adapter`

jeremyf:
# 🎁 Add `Bulkrax.persistence_adapter`

At present we have not incorporated using the
`Bulkrax.persistence_adapter` into the application.  It is instead put
forward for discussion and review.
    
Why introduce the adapter strategy?  Because there are several instances
of Bulkrax that are installed along various points of Hyrax
versions as well as non-Hyrax versions; which means that there are some
cases where we'll be using ActiveFedora and some where we'll be moving
towards Valkyrie.
    
The goal of the adapter is two fold:
    
- Clarify and define the data interface between Bulkrax and the Main App
- Allow for ease of maintaining Bulkrax releative to varying versions of
  Hyrax and Hyku; thus allowing folks to upgrade to new features of
  Bulkrax without mandating an upgrade of their entire persistence
  strategy.
    
Related to:
    
- https://github.com/scientist-softserv/hykuup_knapsack/issues/35

### DIFF
--- a/lib/bulkrax.rb
+++ b/lib/bulkrax.rb
@@ -35,6 +35,36 @@ module Bulkrax
                   :reserved_properties,
                   :server_name
 
+    attr_writer :persistence_adapter
+
+    ##
+    # Configure the persistence adapter used for persisting imported data.
+    #
+    # @return [Bulkrax::PersistenceLayer::AbstractAdapter]
+    # @see Bulkrax::PersistenceLayer
+    def persistence_adapter
+      @persistence_adapter || derived_persistence_adapter
+    end
+
+    def derived_persistence_adapter
+      if defined?(Hyrax)
+        # There's probably some configuration of Hyrax we could use to better refine this; but it's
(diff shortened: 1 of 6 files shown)


jeremyf:
@orangewolf does this approach make sense to proceed further?  It's also something to consider for the `iiif_print` gem.

orangewolf:
well... I got here eventually. I don't understand what this does that is different than the ObjectFactory. I'm not groking the need for both. Its not hurting anything at the moment, but we should talk soon before it takes off to far =-)

jeremyf:
Thanks for the prompt, I think these methods could be added as class methods to the object factory and we'd be in good shape.

jeremyf:
Here's an example of where we need a class method on the object factory:

https://github.com/samvera/bulkrax/blob/4fe18403b6885801f927fb18eefed90259610c98/app/parsers/bulkrax/csv_parser.rb#L212-L216

PR #898, opened 2024-01-24
I897 Bulkrax readiness for Hyku 6 and Hyrax 4 & 5

ShanaLMoore:
These steps should get us to a "user functional" state of Bulkrax and IiifPrint, which we can then begin further testing of Hyku.

- [x]  Move transactions out of initializer; create tests (akin to IiifPrint’s tests) to demonstrate the step composition.
  a. Convert the `create_with_bulk_behavior` and `update_with_bulk_behavior` to a constant; that way we can reference it in IiifPrint and document the “magic” string.
- [x]  Update IiifPrint to sniff out if Bulkrax has a transaction for `update_with_bulk_behavior` and `create_with_bulk_behavior` (check if the constants are defined); then inject accordingly.
  - Update:  create_with_bulk_behavior is not needed for iiif_print
- [x] Address TODOs (`git diff main..hyrax-4-valkyrie-support | rg TODO`)
  - [x]  TODO: uncomment when we upgrade Hyrax 4.x
     - How can we handle both Hyrax 4.x and Hyrax 3.x versions?  Note IiifPrint has gem version sniffing examples `if Gem::Version.new(Hyrax::VERSION) >= Gem::Version.new('4.0.0')`
  - [x] TODO: What other fields are necessary: %i[id edit_users edit_groups read_groups work_members_attributes]?
      - We probably want to add the above array to the code.
  - [x] TODO: there has got to be a better way. Only array types have 'of'
      - Punt, but if it comes to you...
  - [x] TODO: move outside of initializer?
      - Do it, so we can write tests and have a place for constants that IiifPrint can work through.
  - [x] TODO: uninitialized constant BulkraxTransactionContainer::InlineUploadHandler
      - We'll need to be mindful of 
- [x] Specs should pass
  - [x] Reconcile HasMatchers with HasMappingExt. (get rid of HasMappingExt and make `multiple?` work for both Valkyrie and ActiveFedora). 
    - [x] **diffs**: #field_supported?, #multiple?, #multiple_field?, #ar_multiple?, #valkyrie_multiple?, #schema_form_definitions
  - [x] Make sure specs works for both Valkyrie resources (Bulkrax::ValkyrieObjectFactory) and Active Fedora

ref: 
- https://github.com/samvera/bulkrax/issues/897
- https://github.com/scientist-softserv/iiif_print/pull/324


app/factories/bulkrax/valkyrie_object_factory.rb
```diff
@@ -165,12 +165,12 @@ def destroy_existing_files
     private
 
     def transaction
```
jeremyf, line 167:
```suggestion
    # TODO Rename to create_transaction
    def transaction
```

app/transactions/bulkrax/container.rb
jeremyf, line 1: With the `bulkrax/transactions.rb`, let's move this file into `bulkrax/transactions/container.rb`

spec/bulkrax/transactions/container_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Hyrax::Transactions::Container do
```
jeremyf, line 5:
```suggestion
# Yes, we're testing Hyrax::Transactions::Container and not Bulkrax::Transactions::Container, because we want to see the
# impact of the change on Hyrax's implementation.
RSpec.describe Hyrax::Transactions::Container do
```

app/models/concerns/bulkrax/has_matchers.rb
```diff
@@ -125,9 +129,14 @@ def field_supported?(field)
 
       return false if excluded?(field)
       return true if supported_bulkrax_fields.include?(field)
+
       property_defined = factory_class.singleton_methods.include?(:properties) && factory_class.properties[field].present?
 
-      factory_class.method_defined?(field) && (Bulkrax::ValkyrieObjectFactory.schema_properties(factory_class).include?(field) || property_defined)
+      if factory_class == Bulkrax::ValkyrieObjectFactory
```
laritakr, line 135: factory_class is not the right thing to test here.

laritakr (changes requested):
see comment... i'm working on the valkyrie spec and found an issue.

PR #899, opened 2024-01-25
Ams gbh valkyrie support

jeremyf:
Note, this is in direct conflict with the `hyrax-4-valkyrie-support` branch and its PR.  We're trying to avoid a breaking change.


ShanaLMoore (changes requested):
Add specs for valkyrie. Get the specs to pass.

PR #900, opened 2024-01-25
♻️ Extract Bulkrax::FactoryClassFinder

jeremyf:
This refactor introduces consolidating logic for determining an entry's factory_class.

The goal is to begin to allow for us to have a CSV record that says "model = Work" and to use a "WorkResource".

Note, there are downstream implementations that overwrite `factory_class` and we'll need to consider how we approach that.


laritakr (approved):

spec/spec_helper.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'dry/monads'
```
orangewolf, line 3: @jeremyf this dry-monads is now a dev dependency of bulkrax with this change. but I dont see where you are actually using it? can this come out or do we need to add it to the gemspec?

jeremyf: Without the `require 'dry/monads'` the specs failed to load.  This was the quickest adjustment.

Looking at it now, my assumption is that `dry/monads` is a development dependency in that we are creating transactions and merging those transactions.

So the shorter solution would be to move this to a development dependency.

PR #901, opened 2024-01-26
🎁 Config option for coercing a factory class name

jeremyf:
Prior to this commit, once we had a factory class name we always called `#constantize` on the string.

With this commit, we retain that default behavior by:

1. Introducing a function that does this thing
2. Exposing a configuration end-point for that function

Further, we add a looming Valkyrie convention; namely that for each legacy named model (e.g. `Work`) there will likely be a corresponding model with the Resrouce suffix (e.g. `WorkResource`).

Related to:

- https://github.com/scientist-softserv/hykuup_knapsack/issues/125

### DIFF
--- a/app/services/bulkrax/factory_class_finder.rb
+++ b/app/services/bulkrax/factory_class_finder.rb
@@ -3,27 +3,61 @@
 module Bulkrax
   class FactoryClassFinder
     ##
-    # @param entry [Bulkrax::Entry]
-    # @return [Class]
-    def self.find(entry:)
-      new(entry: entry).find
+    # The v6.0.0 default coercer.  Responsible for converting a factory class name to a constant.
+    module DefaultCoercer
+      ##
+      # @param name [String]
+      # @return [Class] when the name is a coercible constant.
+      # @raise [NameError] when the name is not coercible to a constant.
+      def self.call(name)
+        name.constantize
+      end
     end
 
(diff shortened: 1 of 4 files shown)


app/services/bulkrax/factory_class_finder.rb
```diff
@@ -2,28 +2,62 @@
...
+      # @param suffix [String] the suffix we use for a naming convention.
+      #
+      # @return [Class] when the name is a coercible constant.
+      # @raise [NameError] when the name is not coercible to a constant.
+      def self.call(name, suffix: SUFFIX)
+        if name.end_with?(suffix)
+          name.constantize
+        else
+          begin
+            "#{name}#{suffix}".constantize
```
laritakr, line 34: So this makes "Resource" a 'reserved' term referring only to a model which has a non-resource class as well?

I think this is the behavior in our hyrax adapter, but I don't think we perpetuated it outside of the adapter, did we? Should this behavior be in an adapter rather than this module? (just a suggestion/question, as I haven't looked into the code in Hyrax.... I just saw the behavior)

laritakr: It looks below like the coercer is configurable, so I guess someone could write their own... so maybe never mind, as I hadn't looked all the way down. lol

laritakr (approved):

PR #907, opened 2024-01-31
use find_by_source_identifier instead of find_by_bulkrax_identifier

ShanaLMoore:
find_by_bulkrax_identifier.rb should be more dynamic. 

related HYKU changes: 
- https://github.com/samvera/hyku/blob/i35-valkyrize-hyku/config/initializers/wings.rb#L49-L58


app/parsers/bulkrax/application_parser.rb
```diff
@@ -68,15 +68,15 @@ def records(_opts = {})
     # @return [Symbol] the name of the identifying property in the source system from which we're
     # importing (e.g. is *not* this application that mounts *this* Bulkrax engine).
     #
-    # @see #work_identifier
+    # @see #source_identifier
```
kirkkwang, line 71: hmmm why the change here?

app/factories/bulkrax/object_factory.rb
```diff
@@ -28,10 +28,10 @@ class ObjectFactory # rubocop:disable Metrics/ClassLength
     class_attribute :transformation_removes_blank_hash_values, default: false
 
     define_model_callbacks :save, :create
-    attr_reader :attributes, :object, :source_identifier_value, :klass, :replace_files, :update_files, :work_identifier, :work_identifier_search_field, :related_parents_parsed_mapping, :importer_run_id
+    attr_reader :attributes, :object, :source_identifier_value, :klass, :replace_files, :update_files, :work_identifier, :work_identifier_search_field, :related_parents_parsed_mapping, :importer_run_id, :entry
 
     # rubocop:disable Metrics/ParameterLists
-    def initialize(attributes:, source_identifier_value:, work_identifier:, work_identifier_search_field:, related_parents_parsed_mapping: nil, replace_files: false, user: nil, klass: nil, importer_run_id: nil, update_files: false)
+    def initialize(attributes:, source_identifier_value:, work_identifier:, work_identifier_search_field:, related_parents_parsed_mapping: nil, replace_files: false, user: nil, klass: nil, importer_run_id: nil, update_files: false, entry:)
```
ShanaLMoore, line 34: I'm worries this is a breaking change for upgraded non valkyrie users. 

Should I make it optional?

ShanaLMoore: disregard, I don't need it. Object factory is already passing work_identifier: parser.work_identifier.

laritakr (approved):

PR #908, opened 2024-02-02
🧹 Make CreateRelationshipJob work for Valkyrie

kirkkwang:
This will add a relationships path for Valkyrie objects.  It also will add a transactions call so set child flag will fire off in IIIF Print.

ref:
  - https://github.com/scientist-softserv/hykuup_knapsack/issues/141

Child links to parent work now, after successful job completion: 
image

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -90,17 +90,18 @@ module Bulkrax
     def find
       found = find_by_id if attributes[:id].present?
       return found if found.present?
-    rescue Valkyrie::Persistence::ObjectNotFoundError
+      return search_by_identifier if attributes[work_identifier].present?
       false
-    else
-      search_by_identifier if attributes[work_identifier].present?
     end
 
     def find_by_id
+      # TODO: Push logic into Bulkrax.persistence_adapter; maybe
       # Rails / Ruby upgrade, we moved from :exists? to :exist?  However we want to continue (for a
       # bit) to support older versions.
       method_name = klass.respond_to?(:exist?) ? :exist? : :exists?
       klass.find(attributes[:id]) if klass.send(method_name, attributes[:id])
+    rescue Valkyrie::Persistence::ObjectNotFoundError
+      false
     end
 
     def find_or_create
@@ -110,11 +111,13 @@ module Bulkrax
     end
 
     def search_by_identifier
+      # TODO: Push logic into Bulkrax.persistence_adapter; maybe
       query = { work_identifier_search_field =>
                 source_identifier_value }
       # Query can return partial matches (something6 matches both something6 and something68)
       # so we need to weed out any that are not the correct full match. But other items might be
       # in the multivalued field, so we have to go through them one at a time.
+      #
       match = klass.where(query).detect { |m| m.send(work_identifier).include?(source_identifier_value) }
       return match if match
     end
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -92,7 +92,7 @@ module Bulkrax
       return found if found.present?
     rescue Valkyrie::Persistence::ObjectNotFoundError
       false
-    ensure
+    else
       search_by_identifier if attributes[work_identifier].present?
     end
 
--- a/app/jobs/bulkrax/create_relationships_job.rb
+++ b/app/jobs/bulkrax/create_relationships_job.rb
@@ -81,7 +81,7 @@ module Bulkrax
           if @parent_record_members_added
             parent_record.save!
             # Ensure that the new relationship gets indexed onto the children
-            if parent_record.kind_of?(Valkyrie::Resource)
+            if parent_record.is_a?(Valkyrie::Resource)
               @child_members_added.each do |child|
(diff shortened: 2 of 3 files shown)


lib/bulkrax/persistence_layer/active_fedora_adapter.rb
```diff
@@ -6,7 +6,7 @@ class ActiveFedoraAdapter < AbstractAdapter
       def self.find(id)
         ActiveFedora::Base.find(id)
       rescue ActiveFedora::ObjectNotFoundError => e
-        raise PersistenceLayer::RecordNotFound, e.message
+        raise PersistenceLayer::ObjectNotFoundError, e.message
```
jeremyf, line 9: Good catch!

jeremyf (approved):

app/factories/bulkrax/object_factory.rb
```diff
@@ -90,17 +90,18 @@ def update
     def find
       found = find_by_id if attributes[:id].present?
       return found if found.present?
-    rescue Valkyrie::Persistence::ObjectNotFoundError
+      return search_by_identifier if attributes[work_identifier].present?
```
ShanaLMoore, line 93: This change fixed the reason why a link was not being displayed on a successful import. Before it would leave a message that the work didn't successfully import when that wasn't true. 

image

app/jobs/bulkrax/create_relationships_job.rb
```diff
@@ -165,13 +172,34 @@ def add_to_collection(child_record, parent_record)
...
+
+      # TODO: Hyrax is in the process of extracting an "Action" object that we could call.  It does
+      # provide validation that we may want to consider.
+      #
+      # NOTE: We may need to look at the step args we're passing, see
+      #      `Hyrax::WorksControllerBehavior#update_valkyrie_work`
+      #      Hyrax's `./app/controllers/concerns/hyrax/works_controller_behavior.rb`
+      #
+      change_set = Hyrax::ChangeSet.for(parent_record)
+      Hyrax::Transactions::Container['change_set.update_work'].call(change_set)
```
no-reply, line 194: i'm curious about this choice. won't this ensure that the entire update work process (the same one that happens via the controller action) takes place just to change the work membership?

Hyrax::Publisher provides an `object.membership.updated` event, and i wonder whether it wouldn't be better to save the work with the updated membership and publish that? https://github.com/samvera/hyrax/blob/fdcabe651850b1885d7bd3e1b36777bf5f336ee8/lib/hyrax/publisher.rb#L183-L192

kirkkwang: @no-reply, for our uses, we're trying to hit this [transaction](https://github.com/scientist-softserv/iiif_print/blob/9dec8418c879cf5f9346e4105ae221bdd9946902/app/transactions/hyrax/transactions/steps/set_child_flag.rb), so you're suggesting we put a publisher here and then that means this can become a listener?

kirkkwang: [here](https://github.com/scientist-softserv/iiif_print/blob/b4694c1b33cba86161768ea7f1bd9fecbf1b083c/app/listeners/iiif_print/listener.rb) was the original idea of the listener

no-reply: @kirkkwang yes. i think what i'm suggesting (and i'm just popping in as an outsider here), is that you want this behavior to trigger whenever an object's membership changes, not only when a given transaction runs.

making it a `listener` means you'll trigger these changes when outside callers  (in Hyrax or in the downstream application) change membership, as long as they dutifully publish the expected message.

i think a helpful line to draw around transactions and listeners is:
  - use a transaction step when you want to EXPLICITLY implement some behavior as part of a given operation, but you might want to (explicitly) implement different behavior in a different context.
    - e.g. i might want to set the `current_user` as the "depositor" every time a `create` action happens in the hyrax works controller, but handle this matter differently if i create an object through an API endpoint.
    - it would be normal for the Hyrax engine, Bulkrax's own code, or the downstream application to call different transactions that may or may not include steps the others might use.
   - use a listener when you want to ensure some behavior *always* runs in response to some event, regardless of that event's context.
     - e.g. whenever metadata updates for an object, i want to reindex that object.
 
another way of thinking about it is:
  - if in a naive MVC architecture i would have just written the code in the controller, or as a model method, like `do_this; do_that; do_other`, then you want that to be a transaction step.
  -  if in a naive MVC architecture i would have written a callback, you want a listener. (it might be helpful to think of a listener as a multiplexed callback, decoupled from the model layer).

kirkkwang: Thanks @no-reply that's super helpful!

kirkkwang: FYI I made a ticket here https://github.com/scientist-softserv/iiif_print/issues/330

kirkkwang: @no-reply should [this](https://github.com/samvera/hyrax/blob/double_combo/lib/hyrax/transactions/steps/add_to_parent.rb#L19-L32) also publish `object.membership.updated`?  I see that it there's a membership update but it publishes `object.metadata.updated`.  Can/should it do both?

PR #912, opened 2024-02-08
🧹 Make DeleteJob work wth new class method .find

kirkkwang:
The DeleteJob previously was not working with the old factory#find method because when it is doing a delete action, the parsed_metadata does not get generated like during a regular import.  Because of this, the #search_by_identifier method fails to find anything because we don't have a `work_identifier` field which would have came from the parsed_metadata.  So instead, we are using the new class method .find which will take an id (which we find on the raw_metadata) to find the object.  We make sure to reindex and publish the action to any relevant listeners.

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -189,6 +189,10 @@ module Bulkrax
       Rails.logger.info("#{msg} (#{Array(attributes[work_identifier]).first})")
     end
 
+    def delete(obj, _user)
+      obj&.delete
+    end
+
     private
 
     # @param [Hash] attrs the attributes to put in the environment
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -122,7 +122,8 @@ module Bulkrax
     def find
       found = find_by_id if attributes[:id].present?
       return found if found.present?
-      return search_by_identifier if attributes[work_identifier].present?
+      return search_by_identifier if source_identifier_value.present?
+
       false
     end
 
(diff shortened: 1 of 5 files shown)


app/jobs/bulkrax/delete_job.rb
```diff
@@ -5,8 +5,16 @@ class DeleteJob < ApplicationJob
     queue_as :import
 
     def perform(entry, importer_run)
-      obj = entry.factory.find
-      obj&.delete
+      obj = entry.factory.class.find(entry.raw_metadata["id"])
```
jeremyf, line 8: What about writing: `entry.factory.delete`

And then each factory would implement that according to the strategy.

ShanaLMoore:
todo: update hyku bulkrax && include bulkrax application helper in application help

app/jobs/bulkrax/delete_job.rb
```diff
@@ -5,8 +5,10 @@ class DeleteJob < ApplicationJob
...
```
jeremyf, line 8: I'm confused, why can we do `entry.factory.find`?

It seams like we could be able to call `entry.factory.delete(user: importer_run.importer.user)` and it would do the right thing.

jeremyf (changes requested):
See inline comment

app/factories/bulkrax/object_factory.rb
```diff
@@ -122,7 +122,8 @@ def update
     def find
       found = find_by_id if attributes[:id].present?
       return found if found.present?
-      return search_by_identifier if attributes[work_identifier].present?
+      return search_by_identifier if source_identifier_value.present?
```
jeremyf, line 125: Love it!

jeremyf (approved):

PR #914, opened 2024-02-08
Update Importer Index and Show Entries With Search, Filtering, Sort and More

orangewolf:
Datatables is already used in Hyrax and elsewhere in Bulkrax. This connects it via Ajax and consolidates the 3 object type tables down in to one easy to use grid.

This change is only breaking if you have overridden the importer show page.

[Demo Video](https://share.zight.com/z8uB6kZO)


app/views/bulkrax/shared/_entries_tab.html.erb
```diff
@@ -1,44 +1,16 @@
-<div role="tabpanel" class="tab-pane bulkrax-nav-tab-table-left-align <%= pagination_anchor == 'work-entries' ? 'active' : '' %>" id="<%= pagination_anchor %>">
-  <table class='table table-striped'>
+<div role="tabpanel" class="tab-pane bulkrax-nav-tab-table-left-align" >
```
jeremyf, line 1: There's no declared `tablist` nor `tab` roles, so I think we should not have a `tabpanel`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tabpanel_role

jeremyf (changes requested):
The `./app/views/bulkrax/exporters/show.html.erb` also renders the `shared/_entries_tab` partial and the change to the `shared/_entries_tab` appears to break how the exporter's rendering of that tab.

orangewolf:
> The `./app/views/bulkrax/exporters/show.html.erb` also renders the `shared/_entries_tab` partial and the change to the `shared/_entries_tab` appears to break how the exporter's rendering of that tab.

this is being fixed in a seperate pr that's coming in hot

PR #915, opened 2024-02-08
Strip out all special characters from csv headers

orangewolf:
These cause nothing but pain.

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -16,11 +16,12 @@ module Bulkrax
     class_attribute(:csv_read_data_options, default: {})
 
     # there's a risk that this reads the whole file into memory and could cause a memory leak
+    # we strip any special characters out of the headers. looking at you Excel
     def self.read_data(path)
       raise StandardError, 'CSV path empty' if path.blank?
       options = {
         headers: true,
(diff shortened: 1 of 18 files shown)


app/models/bulkrax/csv_entry.rb
```diff
@@ -16,11 +16,12 @@ def self.fields_from_data(data)
     class_attribute(:csv_read_data_options, default: {})
 
     # there's a risk that this reads the whole file into memory and could cause a memory leak
+    # we strip any special characters out of the headers. looking at you Excel
     def self.read_data(path)
       raise StandardError, 'CSV path empty' if path.blank?
       options = {
         headers: true,
-        header_converters: ->(h) { h.to_s.strip.to_sym },
+        header_converters: ->(h) { h.to_s.gsub(/[^\w\d -]+/, '').strip.to_sym },
```
jeremyf, line 24: Do we want underscore as well?

orangewolf: @jeremyf \w includes underscore.

orangewolf:
closing this one for #918 instead as it is a cleaner PR

app/models/bulkrax/csv_entry.rb
jeremyf: Today I (re-)learned!

PR #929, opened 2024-02-23
♻️ Favor member_ids_ssim over file_set_ids_ssim

jeremyf:
For several years Hyrax has index `file_set_ids_ssim` as a verbatim copy of `member_ids_ssim`.  With Hyrax 5, we're removing the `file_set_ids_ssim` from indexing; And given that it's been a verbatim copy since 2017 or so, it's relatively safe to assume that we can favor, without application impact, the `member_ids_ssim` over the `file_set_ids_ssim` value.

It would be nice to have `file_set_ids_ssim` but not as a verbatim copy. Someday, we might have nice things.

In the case of Bulkrax, given that we don't require Hyrax, we need to have a configurable option as well as one that is a consistent fallback to what came before.

Related to:
- https://github.com/samvera/hyrax/pull/6513
- https://github.com/samvera/hyrax/commit/7108409c619cd2ba4ae8c836b9f3b429a7e9837b

### DIFF
--- a/app/parsers/bulkrax/parser_export_record_set.rb
+++ b/app/parsers/bulkrax/parser_export_record_set.rb
@@ -120,7 +120,7 @@ module Bulkrax
       #
       # No sense attempting to query for more than the limit.
       def query_kwargs
-        { fl: "id,member_ids_ssim", method: :post, rows: row_limit }
+        { fl: "id,#{Bulkrax.solr_key_for_member_file_ids}", method: :post, rows: row_limit }
       end
 
       # If we have a limit, we need not query beyond that limit
--- a/app/parsers/bulkrax/parser_export_record_set.rb
+++ b/app/parsers/bulkrax/parser_export_record_set.rb
@@ -113,14 +113,14 @@ module Bulkrax
       #
       # @see #file_sets
       def candidate_file_set_ids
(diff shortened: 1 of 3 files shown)


app/parsers/bulkrax/parser_export_record_set.rb
```diff
@@ -113,14 +113,14 @@ def each
...
-        @candidate_file_set_ids ||= works.flat_map { |work| work.fetch("#{Bulkrax.file_model_class.to_s.underscore}_ids_ssim", []) }
+        @candidate_file_set_ids ||= works.flat_map { |work| work.fetch(Bulkrax.solr_key_for_member_file_ids, []) }
       end
 
       # @note Specifically not memoizing this so we can merge values without changing the object.
       #
       # No sense attempting to query for more than the limit.
       def query_kwargs
-        { fl: "id,#{Bulkrax.file_model_class.to_s.underscore}_ids_ssim", method: :post, rows: row_limit }
+        { fl: "id,member_ids_ssim", method: :post, rows: row_limit }
```
jeremyf, line 123:
```suggestion
        { fl: "id,#{Bulkrax.solr_key_for_member_file_ids}", method: :post, rows: row_limit }
```

kirkkwang (approved):

PR #932, opened 2024-03-04
fix oai after application_parser rework

orangewolf:


app/parsers/bulkrax/application_parser.rb
```diff
@@ -242,16 +242,28 @@ def calculate_type_delay(type)
...
+    def record_raw_metadata(record)
+      record.to_h
+    end
+
+    def record_deleted?(record)
+      record[:delete].present?
+    end
+
+    def record_remove_and_rerun?(record)
+      record[:remove_and_rerun].present?
```
bkiahstroud, line 254: Non-blocking: Are these scenarios where casting to a boolean would be better than `.present?` Since this is the `ApplicationParser`, I'm thinking about a CSV import where the value in the `:delete` column is `"false"`, for example

bkiahstroud (approved):

PR #945, opened 2024-03-26
only show zero file_size warning once per download

cziaarm:
If you keep those log messages (which are handy) then this change will mean we don't get the logs flooded when downloading multi GB files.


kirkkwang (approved):
very interesting syntax!

for our future selves, ref:
https://www.ruby-forum.com/t/ruby-way-to-take-some-action-only-once-in-a-loop/151433/8

PR #954, opened 2024-05-10
🐛 Fix FileSet csv rows for AF imports

kirkkwang:
# Summary

This commit will fix the functionality for adding a FileSet row in the CSV to add additional metadata to FileSets.  The problem was located in `Bulkrax::ObjectFactory#create_file_set`.  When it tries to find a record, it actually comes back as a `FalseClass` because of the new logic introduced recently which breaks further down.  If we add raise an error in our `Bulkrax::ImportFileSetJob#find_parent_record` which is at a higher point than before, then we can requeue the job and check again. If the parent record is created by then, we will execute the rest of the job.

Issue: 
- https://github.com/scientist-softserv/hykuup_knapsack/issues/216

# Screenshots

image

image

# Testing Instructions

This fix was made in the PALs repo, here is the CSV used:
[pals-with-file-row.zip](https://github.com/samvera/bulkrax/files/15276997/pals-with-file-row.zip)

### DIFF
--- a/app/jobs/bulkrax/import_file_set_job.rb
+++ b/app/jobs/bulkrax/import_file_set_job.rb
@@ -16,7 +16,12 @@ module Bulkrax
       # e.g. "parents" or "parents_1"
       parent_identifier = (entry.raw_metadata[entry.related_parents_raw_mapping] || entry.raw_metadata["#{entry.related_parents_raw_mapping}_1"])&.strip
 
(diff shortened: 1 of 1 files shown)


ShanaLMoore:
Relevant ticket: https://github.com/scientist-softserv/hykuup_knapsack/issues/216

ShanaLMoore (approved):

PR #958, opened 2024-05-24
add support for downloading the original file

orangewolf:
It can be useful to be able to download the file that was originally uploaded for an import. It won't work in all cases (such as OAI) but it will work in most cases.

Screenshot 2024-05-23 at 22 24 30
Screenshot 2024-05-23 at 22 24 23

### DIFF
--- a/app/models/bulkrax/importer.rb
+++ b/app/models/bulkrax/importer.rb
@@ -149,6 +149,14 @@ module Bulkrax
       @seen ||= {}
     end
 
+    def original_file?
+      File.exists?(self.parser_fields['import_file_path'])
+    end
+
+    def original_file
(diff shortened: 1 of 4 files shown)


app/models/bulkrax/importer.rb
```diff
@@ -149,6 +149,14 @@ def seen
       @seen ||= {}
     end
 
+    def original_file?
+      File.exists?(self.parser_fields['import_file_path'])
```
bkiahstroud, line 153: This should appear rubocop 🤖👮‍♂️ 

```suggestion
      File.exist?(self.parser_fields['import_file_path'])
```

bkiahstroud (changes requested):

bkiahstroud (approved):

PR #960, opened 2024-06-12
fix a memory leak when copying large files by using the hyrax uploader

orangewolf:
instead of a custom one

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -250,7 +250,7 @@ module Bulkrax
 
     def file_set_operation_for(user:)
       Hyrax::Operation.create!(user: user,
-        operation_type: "Attach Remote File")
+                               operation_type: "Attach Remote File")
     end
   end
   # rubocop:enable Metrics/ClassLength
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -245,7 +245,7 @@ module Bulkrax
       actor.file_set.import_url = remote_file['url']
       auth_header = remote_file.fetch('auth_header', {})
 
-      ImportUrlJob.perform_now(actor.file_set, operation_for(user: @user), auth_header)
+      ImportUrlJob.perform_now(actor.file_set, file_set_operation_for(user: @user), auth_header)
     end
 
     def file_set_operation_for(user:)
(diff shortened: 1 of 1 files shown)


app/factories/bulkrax/object_factory.rb
```diff
@@ -243,18 +243,14 @@ def update_file_set(attrs)
     def handle_remote_file(remote_file:, actor:, update: false)
```
ShanaLMoore, line 243: with these changes, update is an unused variable. Do we still need it?

app/factories/bulkrax/object_factory.rb
```diff
@@ -243,18 +243,14 @@ def update_file_set(attrs)
...
-      end
+      ImportUrlJob.perform_now(actor.file_set, operation_for(user: @user), auth_header)
+    end
 
-      tmp_file.rewind
-      update == true ? actor.update_content(tmp_file) : actor.create_content(tmp_file, from_url: true)
-      tmp_file.close
+    def file_set_operation_for(user:)
+      Hyrax::Operation.create!(user: user,
+        operation_type: "Attach Remote File")
```
ShanaLMoore, line 253: lint correction
```suggestion
                               operation_type: "Attach Remote File")
```

app/factories/bulkrax/object_factory.rb
```diff
@@ -243,18 +243,14 @@ def update_file_set(attrs)
...
-        tmp_file.write(url_file.read)
-      end
+      ImportUrlJob.perform_now(actor.file_set, operation_for(user: @user), auth_header)
+    end
 
-      tmp_file.rewind
-      update == true ? actor.update_content(tmp_file) : actor.create_content(tmp_file, from_url: true)
-      tmp_file.close
+    def file_set_operation_for(user:)
+      Hyrax::Operation.create!(user: user,
```
bkiahstroud, line 252: Last year, we did a lot of work to decouple Bulkrax from Hyrax. Is is fine to call `Hyrax` directly here because it's in the `ObjectFactory`?

orangewolf: yes, the object factory has to be aware of Hyrax, its the one place that code is allowable.

app/factories/bulkrax/object_factory.rb
orangewolf: yes, probably. I walked the code path and I _think_ calling create_content all the time instead of sometimes calling update_content is the right call. You are attaching an unknowable file (remote source could change it) if you are downloading it from a url, so it makes sense to me that it has to be treated as new. I only have a confidence level of 72.5% on this though

orangewolf: updated

bkiahstroud (approved):

PR #961, opened 2024-06-18
i84 - update visibility label and help text

ShanaLMoore:
This was contributed back per PALNI PALCI. Since it's possible to override whatever setting the user selects at the form level, help text is provided. If the visibility is declared at the CSV level, the visibility selected at the form level will be ignored.

Issue:
- https://github.com/scientist-softserv/palni_palci_knapsack/issues/84

## BEFORE

image

## AFTER

image

### DIFF
--- a/app/views/bulkrax/importers/_csv_fields.html.erb
+++ b/app/views/bulkrax/importers/_csv_fields.html.erb
@@ -1,13 +1,15 @@
 <div class='csv_fields'>
 
   <%= fi.input :visibility,
+    label: 'Default Visibility',
     collection: [
       ['Public', 'open'],
       ['Private', 'restricted'],
       ['Institution', 'authenticated']
     ],
     selected: importer.parser_fields['visibility'] || 'open',
-    input_html: { class: 'form-control' }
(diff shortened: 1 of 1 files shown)


laritakr (approved):

app/views/bulkrax/importers/_csv_fields.html.erb
```diff
@@ -1,13 +1,15 @@
...
+    label: 'Default Visibility',
     collection: [
       ['Public', 'open'],
       ['Private', 'restricted'],
       ['Institution', 'authenticated']
     ],
     selected: importer.parser_fields['visibility'] || 'open',
-    input_html: { class: 'form-control' }
+    input_html: { class: 'form-control' },
+    hint: 'If your CSV includes the visibility field, it will override the default setting.'
```
ShanaLMoore, line 12: I implemented this following pre-existing bulkrax patterns, which doesn't seem to use locales or support translations. 

ref: https://github.com/samvera/bulkrax/blob/main/app/views/bulkrax/importers/_csv_fields.html.erb#L23

PR #964, opened 2024-06-21
Delete and redo file sets

orangewolf:


ShanaLMoore (dismissed):

app/jobs/bulkrax/delete_file_set_job.rb
```diff
@@ -2,4 +2,22 @@
 
 module Bulkrax
   class DeleteFileSetJob < DeleteJob; end
+
+  def perform(entry, importer_run)
```
ShanaLMoore, line 6: Do you need to pass the importer_run? It doesn't seem like it's getting used here.

app/jobs/bulkrax/delete_file_set_job.rb
```diff
@@ -2,4 +2,22 @@
 
 module Bulkrax
   class DeleteFileSetJob < DeleteJob; end
```
ShanaLMoore, line 4: Also, ignore me if I'm reading this wrong, but is this an extra end?

app/jobs/bulkrax/delete_file_set_job.rb
orangewolf: its used in the parent job (so the call to super gets it)

app/jobs/bulkrax/delete_file_set_job.rb
orangewolf: grr - I really hate the `; end` syntax. it takes an important thing and hides it for no real gain but to save a linefeed. fixed

app/jobs/bulkrax/delete_file_set_job.rb
orangewolf: this was a really good catch Shana. its legal to define the method on the module so its syntaxitcally correct but would not have ever been called the way I had it

app/jobs/bulkrax/delete_file_set_job.rb
orangewolf: this was a really good catch Shana. its legal to define the method on the module so its syntaxitcally correct but would not have ever been called the way I had it

ShanaLMoore (approved):

PR #965, opened 2024-07-11
🐛 bom bug fix

ShanaLMoore:
Issue:
- https://github.com/scientist-softserv/adventist_knapsack/issues/680

The client would import CSVs that had identifier columns, but the importer would still fail with "Missing at least one required element, missing element(s) are: identifier issue when the CSV contains the identifier column."

This commit utitlizes the BOM method, previously created to normalize data, and adds bom support for reading CSVs.


ShanaLMoore:
I don't think this PR is needed after all. Bulkrax main seemed to already have a solution for this: 

image

PR #968, opened 2024-08-16
Allow for only thumbnail url in csv

laritakr:
Handles situation where the csv has a thumbnail_url but no related_url to use as remote files.

Previously threw an error: `NoMethodError - undefined method + for nil:NilClass`

Screenshot 2024-08-16 at 7 48 15 PM

### DIFF
--- a/app/factories/bulkrax/valkyrie_object_factory.rb
+++ b/app/factories/bulkrax/valkyrie_object_factory.rb
@@ -266,6 +266,7 @@ module Bulkrax
 
     def combined_files_with(remote_files:)
       thumbnail_url = self.attributes['thumbnail_url']
+      return [thumbnail_url] if remote_files.blank?
       @combined_files ||= (thumbnail_url.present? ? remote_files + [thumbnail_url] : remote_files)
     end
 


ShanaLMoore (dismissed):

app/factories/bulkrax/valkyrie_object_factory.rb
```diff
@@ -266,6 +266,7 @@ def create_work(attrs)
 
     def combined_files_with(remote_files:)
       thumbnail_url = self.attributes['thumbnail_url']
+      return [thumbnail_url] if remote_files.blank?
```
ShanaLMoore, line 269: to clarify, if thumbnail_url is also blank it'll return an empty array?

ShanaLMoore (approved):

app/factories/bulkrax/valkyrie_object_factory.rb
laritakr: An empty array should avoid the error (the error was trying to merge the thumbnail url into a nil remote_files value.) 

But would be a good thing to double-check.

PR #976, opened 2024-09-06
Importing local files broke due to remote files work

laritakr:
# Summary

Prior work to handle remote files did not work for local files.

Prior work: 
- https://github.com/samvera/bulkrax/pull/975
- https://github.com/samvera/bulkrax/pull/971
- https://github.com/samvera/bulkrax/pull/968
- https://github.com/samvera/bulkrax/pull/967

Refs:
- https://github.com/scientist-softserv/palni-palci/issues/1010

### DIFF
--- a/app/factories/bulkrax/valkyrie_object_factory.rb
+++ b/app/factories/bulkrax/valkyrie_object_factory.rb
@@ -413,7 +413,7 @@ module Bulkrax
       end
     end
 
-    def uploaded_s3_files(remote_files: {})
+    def uploaded_s3_files(remote_files: [])
       return [] if remote_files.blank?
 
       s3_bucket_name = ENV.fetch("STAGING_AREA_S3_BUCKET", "comet-staging-area-#{Rails.env}")
@@ -425,7 +425,7 @@ module Bulkrax
       end.compact
     end
 
-    def uploaded_remote_files(remote_files: {})
+    def uploaded_remote_files(remote_files: [])
       remote_files.map do |r|
         file_path = download_file(r["url"])
         next unless file_path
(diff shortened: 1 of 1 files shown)


app/factories/bulkrax/valkyrie_object_factory.rb
```diff
@@ -264,25 +263,57 @@ def create_work(attrs)
...
+    def prep_fileset_content(attrs)
+      # combine remote_files + thumbnail_url [Array < { url:, file_name:, * }]
       thumbnail_url = HashWithIndifferentAccess.new(self.attributes)['thumbnail_url']
+      all_remote_files = merge_thumbnails(remote_files: attrs["remote_files"], thumbnail_url: thumbnail_url)
+      # combine local & remote files [Array < Hash &/or String]
+      all_local_files = self.attributes['file']
+      all_files = all_local_files + all_remote_files
+
+      # collect all uploaded files [Array < Hyrax::UploadedFile]
+      uploaded_local = uploaded_local_files(uploaded_files: attrs[:uploaded_files])
```
ShanaLMoore, line 278: apologies if I'm missing it, but do we need a fallback here for when these fields don't exist? like: 

```
all_remote_files = merge_thumbnails(remote_files: attrs["remote_files"] || [], thumbnail_url: thumbnail_url)
uploaded_local = uploaded_local_files(uploaded_files: attrs[:uploaded_files] || [])
uploaded_remote = uploaded_remote_files(remote_files: all_remote_files)
```

laritakr: You're right, we aren't calling that method without the parameter, but if we did, it would break. I changed the default to an empty array since remote_files should always be an array of hashes.

ShanaLMoore (approved):

PR #977, opened 2024-09-16
🐛 Fix importers sorting for last run and next run

kirkkwang:
# Summary
## 🐛 Fix importers sorting for last run and next run

7d985dda63bb027fb9f1cb7416709a54a420968a

This commit will fix the sorting error that happens when the user is on
the importers index and attempts to sort by the last run or next run. We
add some migrations to add fields that the datatables can use so it can
sort properly. Previously, this was not working because the
last_imported_at and next_import_at fields were actually methods on the
importer_run object and not the importer object.  We are adding a few
callbacks to the importer and importer_run models to ensure that the
fields are properly set when they are called from either the web or the
worker.

Ref:
- https://github.com/samvera/bulkrax/issues/956

https://github.com/user-attachments/assets/65c5b70e-5c9a-40ac-897d-c417fe24b636

## 🤖 Update upload-artifact and download-artifact

aca33e568c06236c5652a0e500da33c5d46cba10

CI was getting errors because the versions we were previously using were
deprecated.  This commit updates the actions to the latest versions.

## 🐛 Remove Downloadable Files sorting

d4d4749b543ca0f6631edfc61287491012b36098

The downloadable files sorting was broken plus, it's not clear now a
downloadable file should be sorted.

## ⚙️ Add guard for new migrations

7011d726bd9f2f37dc19a5f8bddb4089c09c3827

This commit will add a guard to the new migrations to ensure that they
do not run if the columns already exist in the database.

# Note

To get the changes all the Importers need to be updated, something like:
```rb
Importers.all.each(&:save)
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,10 +43,11 @@ jobs:
         run: bundle exec rake spec
 
       - name: Upload coverage results
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4.4.0
         with:
           name: coverage-report-${{ matrix.ruby }}
-          path: coverage
+          path: coverage/**
+          include-hidden-files: true
 
   coverage:
     runs-on: ubuntu-latest
@@ -56,7 +57,7 @@ jobs:
 
     steps:
       - name: Download coverage report
-        uses: actions/download-artifact@v2
+        uses: actions/download-artifact@v4.1.8
         with:
           name: coverage-report-2.7
           path: coverage
--- a/app/assets/javascripts/bulkrax/datatables.js
+++ b/app/assets/javascripts/bulkrax/datatables.js
@@ -67,7 +67,7 @@ Blacklight.onLoad(function() {
         { "data": "name" },
         { "data": "status_message" },
         { "data": "created_at" },
-        { "data": "download" },
+        { "data": "download", "orderable": false },
         { "data": "actions", "orderable": false }
       ],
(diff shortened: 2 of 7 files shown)


ShanaLMoore:
Exporters were also reported with the same issue. Is there an exporter table we need to add a migration to as well?

kirkkwang:
@ShanaLMoore thanks for checking on that.  The only issue I saw was the downloadable files sorting, it was broken, but I'm just taking it out since no one was using it.  It's not clear to me how to sort that column anyways.  Number of files?  Date? etc.  If there's demand for this feature then I think we can figure something out later.

ShanaLMoore:
Sounds good. My other question is for existing applications. In addition to the migration, how are we handling existing importers to populate those fields? And would we need to for the sort to work properly? Perhaps a rake task is needed?

kirkkwang:
@ShanaLMoore We just need to update all the Importers and the changes will get applied.

ShanaLMoore:
right, so maybe we should add a rake task to the migration to accomplish this? vs relying on the dev to remember to do.

kirkkwang:
@ShanaLMoore ah okay, that's a good idea.  So how would that look like?  The dev has to run migrations, and then migration automatically runs the rake task?

ShanaLMoore:
yes, i'll try to find an example. @laritakr has set this up before somewhere.

ShanaLMoore (approved):

PR #980, opened 2024-09-26
Avoid false positive from `.try(:internal_resource)`

bkiahstroud:
Using `#try` on `:internal_resource` can yield unexpected results. If the method is undefined, it can return a truthy value instead of the typical `nil`.

E.g.

```ruby
Hyrax::FileSet.try(:internal_resource) || 'hi'
=> #<Dry::Types::Result::Failure input=:internal_resource error=#<Dry::Struct::Error: [Hyrax::FileSet.new] can't convert Symbol into Hash>>
```

E.g. 

```ruby
Bulkrax.collection_model_internal_resource
=> #<Dry::Types::Result::Failure input=:internal_resource error=#<Dry::Struct::Error: [CollectionResource.new] can't convert Symbol into Hash>>
```

This was discovered while trying to export Valkyrie Resources from HykuUP (see screenshot) 

<details><summary>Screenshot</summary>

image

</details>

### DIFF
--- a/lib/bulkrax.rb
+++ b/lib/bulkrax.rb
@@ -98,7 +98,22 @@ module Bulkrax
     attr_writer :collection_model_class
 
     def collection_model_internal_resource
-      collection_model_class.try(:internal_resource) || collection_model_class.to_s
+      # WARN: Using #try on :internal_resource can yield unexpected results.
+      # If the method is undefined, it can return a truthy value instead of
+      # the typical nil.
+      #
+      # E.g.
+      # ```ruby
+      # Hyrax::FileSet.try(:internal_resource) || 'hi'
+      # => #<Dry::Types::Result::Failure input=:internal_resource error=...
+      # ```
+      ir = begin
+             collection_model_class.internal_resource
+           rescue NoMethodError
+             nil
+           end
+
+      ir.presence || collection_model_class.to_s
     end
 
     def file_model_class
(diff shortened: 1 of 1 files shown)


lib/bulkrax.rb
```diff
@@ -98,7 +98,22 @@ def collection_model_class
...
+      # Hyrax::FileSet.try(:internal_resource) || 'hi'
+      # => #<Dry::Types::Result::Failure input=:internal_resource error=...
+      # ```
+      ir = begin
+             collection_model_class.internal_resource
+           rescue NoMethodError
+             nil
+           end
+
+      ir.presence || collection_model_class.to_s
```
orangewolf, line 116:
```suggestion
      # WARN: Using #try on :internal_resource can yield unexpected results.
      # If the method is undefined, it can return a truthy value instead of
      # the typical nil.
      #
      # E.g.
      # ```ruby
      # Hyrax::FileSet.try(:internal_resource) || 'hi'
      # => #<Dry::Types::Result::Failure input=:internal_resource error=...
      # ```
      ir = collection_model_class.respond_to?(:internal_resource) ?
              collection_model_class.internal_resource : collection_model_class.to_s
```

orangewolf (dismissed):
we can make this more compact I think and more efficient since raising and catching is expensive.

lib/bulkrax.rb
orangewolf:
```ruby
irb(main):001:0> Hyrax::FileSet.try(:internal_resource) || 'hi'
=> #<Dry::Types::Result::Failure input=:internal_resource error=#<Dry::Struct::Error: [Hyrax::FileSet.new] can't convert Symbol into Hash>>
irb(main):002:0> Hyrax::FileSet.internal_resoruce
(irb):2:in `<main>': undefined method `internal_resoruce' for Hyrax::FileSet:Class (NoMethodError)
irb(main):003:0> Hyrax::FileSet.respond_to?(:internal_resoruce)
=> false
```

bkiahstroud: Thanks, that's much cleaner 🙂

laritakr (approved):

orangewolf (approved):

PR #984, opened 2024-11-12
Access the default split pattern from the parser instead of the config

bkiahstroud:
Transfer responsibility of accessing the split pattern from the config to the `ApplicationParser`. This enables the default split pattern to be overwritten

Hyku will soon get an `Account` setting that allows for per-tenant overrides for the default split pattern. This change enables that feature

The default value is still set in the config. Existing installations should see no difference in behavior; this change is backwards-compatible


bkiahstroud:
The [sister Hyku PR](https://github.com/samvera/hyku/pull/2380) for this change was closed / replaced by:
- https://github.com/samvera/hyku/pull/2384

PR #991, opened 2024-12-06
Fixes issues with filesets exporting in Valkyrie

sephirothkod:
Fixes issues with FileSets exporting in Valkyrie versions of Hyrax/Hyku

refs #990 

- demodulized the solr fileset call and wrapped it in quotes so that it doesn't throw solr errors.
- added fileset name call for valkyrie. verifies mime type like in the original non-valkyrie version. (this was previously noted as possibly not working with valkyrie. the errors didn't show because the initial solr error prevented filesets from being found in solr in the first place)

### DIFF
--- a/app/parsers/bulkrax/parser_export_record_set.rb
+++ b/app/parsers/bulkrax/parser_export_record_set.rb
@@ -173,7 +173,7 @@ module Bulkrax
       # @see https://github.com/samvera/hyrax/blob/64c0bbf0dc0d3e1b49f040b50ea70d177cc9d8f6/app/indexers/hyrax/work_indexer.rb#L15-L18
       def file_sets
         @file_sets ||= ParserExportRecordSet.in_batches(candidate_file_set_ids) do |batch_of_ids|
-          fsq = "has_model_ssim:#{Bulkrax.file_model_internal_resource} AND id:(\"" + batch_of_ids.join('" OR "') + "\")"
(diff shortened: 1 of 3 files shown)


app/parsers/bulkrax/parser_export_record_set.rb
```diff
@@ -173,7 +173,7 @@ def collections
       # @see https://github.com/samvera/hyrax/blob/64c0bbf0dc0d3e1b49f040b50ea70d177cc9d8f6/app/indexers/hyrax/work_indexer.rb#L15-L18
       def file_sets
         @file_sets ||= ParserExportRecordSet.in_batches(candidate_file_set_ids) do |batch_of_ids|
-          fsq = "has_model_ssim:#{Bulkrax.file_model_internal_resource} AND id:(\"" + batch_of_ids.join('" OR "') + "\")"
+          fsq = "has_model_ssim:\"#{Bulkrax.file_model_internal_resource.demodulize}\" AND id:(\"" + batch_of_ids.join('" OR "') + "\")"
```
bkiahstroud, line 176: What does `#demodulize` do?

orangewolf (approved):

app/parsers/bulkrax/parser_export_record_set.rb
sephirothkod: It removes the module from a class name. So Hyrax::FileSet would become just FileSet. Which is how Solr stores filesets. If there isn't a module to remove it will do nothing.

PR #994, opened 2024-12-10
Add guard to uploaded_files migration

kirkkwang:
This commit will check if uploaded_files is a table before adding the file_name column to it.  In a Hydra application, it is not so this was crashing.

### DIFF
--- a/db/migrate/20240806161142_add_file_name_to_uploaded_files.rb
+++ b/db/migrate/20240806161142_add_file_name_to_uploaded_files.rb
@@ -1,7 +1,5 @@
 class AddFileNameToUploadedFiles < ActiveRecord::Migration[5.2]
   def change
-    if table_exists?(:uploaded_files)
-      add_column :uploaded_files, :filename, :string unless column_exists?(:uploaded_files, :filename)
-    end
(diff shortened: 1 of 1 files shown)


db/migrate/20240806161142_add_file_name_to_uploaded_files.rb
```diff
@@ -1,5 +1,7 @@
 class AddFileNameToUploadedFiles < ActiveRecord::Migration[5.2]
   def change
-    add_column :uploaded_files, :filename, :string unless column_exists?(:uploaded_files, :filename)
+    if table_exists?(:uploaded_files)
+      add_column :uploaded_files, :filename, :string unless column_exists?(:uploaded_files, :filename)
+    end
```
orangewolf, line 5:
```suggestion
    add_column :uploaded_files, :filename, :string if table_exists?(:uploaded_files) && !column_exists?(:uploaded_files, :filename)
```

orangewolf: either this or do both in the `if` block. don't do both `if` and `unless`

orangewolf (changes requested):
see code suggestion

kirkkwang:
Sounds good

orangewolf (approved):

PR #1004, opened 2025-01-15
guard file uploader js to only execute when needed

orangewolf:


ShanaLMoore (approved):
confirmed that this works against hyku main

PR #1018, opened 2025-02-27
pass mkdir when first referencing unzip path in `#untar`

bkiahstroud:
This fixes a bug where `File.directory?` was being called on `nil` because the unless statement was running before the `Dir.mkdir` statement, which was supposed to create the new dir


kirkkwang (approved):

bkiahstroud:
rubocop is failing for unrelated reasons that will be resolved by #1020, merging

PR #1021, opened 2025-02-27
re-fix `#untar` to correctly create dir

bkiahstroud:
I got a false positive when testing this locally for #1018, the `mkdir` flag is needed for both `importer_unzip_path` invocations


kirkkwang (approved):

kirkkwang:
Okay that makes sense

PR #1022, opened 2025-03-06
Replace spaces with underscores in filenames

bkiahstroud:
When imported metadata is parsed, filenames containing spaces have those spaces replaced with underscores. Before this commit, this would cause the entire importer to fail since the importer thought it didn't have the file at all ("file name.jpg" != "file_name.jpg"). If we're coercing the metadata, it only makes sense to coerce the filenames as well

### DIFF
--- a/app/jobs/bulkrax/importer_job.rb
+++ b/app/jobs/bulkrax/importer_job.rb
@@ -30,6 +30,7 @@ module Bulkrax
       return unless parser.file? && parser.zip?
 
       parser.unzip(parser.parser_fields['import_file_path'])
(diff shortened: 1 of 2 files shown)


bkiahstroud:
One could argue this is a new feature, not a bug fix. What do others think?

laritakr:
> One could argue this is a new feature, not a bug fix. What do others think?

Did we used to handle filenames with spaces? If so it's definitely a bug fix. 

But I tend to see it as a bug fix anyway because not having this causes the importer to fail without letting the user see why.

laritakr (approved):
A much needed fix!

PR #1025, opened 2025-04-04
Bulkrax editing without model

laritakr:
# Summary

https://github.com/samvera/bulkrax/issues/1024

This commit allows a CSV importer edit to find the object to edit by the identifier rather than requiring the class. 

# Behavior before change

If a CSV did not include the model, the edit would create new valkyrie works using the defined default model. This resulted in two works in the resource table, with only the newer one indexed. The new work would only contain the metadata present in the edit CSV. The prior work was left untouched.

# Behavior after change

The source identifier alone is adequate to find the object to edit, and the importer view shows the correct model name when it links to the edited object.

(Note that it still shows the default type above the raw and parsed metadata)

<details>
<summary>Screenshot</summary>

Screenshot 2025-04-04 at 3 46 07 PM

</details>

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -91,7 +91,7 @@ module Bulkrax
     # @note HEY WE'RE USING THIS FOR A WINGS CUSTOM QUERY.  BE CAREFUL WITH
     #       REMOVING IT.
     #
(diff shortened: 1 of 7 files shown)


orangewolf (approved):
go LaRita! I knew you could do it =-D

laritakr:
> go LaRita! I knew you could do it =-D

There's still apparently some sort of bug due to slugs so I had to move it back to draft for now.

PR #1036, opened 2025-05-28
🧹 Set @object for valkyrie object factory #create and #update

ShanaLMoore:
Issues:
- https://github.com/samvera/bulkrax/issues/1035
- original implementation solved this: https://github.com/notch8/wvu_knapsack/issues/42
- reverts and updates changes: https://github.com/samvera/bulkrax/commit/d0c93d8ee9c8dbb9338f4ab9a289fee9294aba1e

Valkyrie object factory now has its own #create and #update methods to reassign @object. Prior this change was made in the object factory interface and caused issues for non valkyrie users.

TODO: 

@dolsysmith and @kerchner to test in GW. 
@laritakr @ShanaLMoore @kirkkwang @sjproctor to test in hyku main

Please report back in the comments section

### DIFF
--- a/app/parsers/bulkrax/oai_dc_parser.rb
+++ b/app/parsers/bulkrax/oai_dc_parser.rb
@@ -61,8 +61,6 @@ module Bulkrax
       ['contributor', 'coverage', 'creator', 'date', 'description', 'format', 'identifier', 'language', 'publisher', 'relation', 'rights', 'source', 'subject', 'title', 'type']
     end
 
-    delegate :list_sets, to: :client
-
     def create_objects(types = [])
       types.each do |object_type|
         send("create_#{object_type.pluralize}")


dolsysmith:
This fix does resolve the bug we discussed today: the work is now imported without any errors. However, the work is still not added to the collection, and the `Bulkrax::PendingRelationship` table is getting the following status reported: 
```
The property `bulkrax_identifier` is not defined on class 'Collection::GeneratedResourceSchema
```
I did check [our customization](https://github.com/gwu-libraries/scholarspace-hyrax/blob/master/app/models/collection.rb) of the collection model, and the `bulkrax_identifier` property is defined.

kirkkwang:
@dolsysmith shot in the dark suggestion here, can you define the `bulkrax_identifier` property above `include Hyrax::BasicMetadata`?  So basically move it to right under `include ::Hyrax::CollectionBehavior`

kerchner:
@dolsysmith I see that @kirkkwang 's suggestion is probably on target - in the work types, the custom properties are below `include ::Hyrax::WorkBehavior` and above `include ::Hyrax::BasicMetadata` but in `collection.rb` it looks like I put it _below_ `include Hyrax::BasicMetadata`.  Hopefully this will result in different behavior.

dolsysmith:
@kirkkwang Thank you for this suggestion. That change to our code seems to have resolved the error thrown by the `CreateRelationshipsJob`, and the work is now being added to the collection by Bulkrax.

kirkkwang:
Glad to hear it!

ShanaLMoore:
I'm so glad @kirkkwang remembered this Hyrax gotcha! 😓 🙃 That module must be [included](https://github.com/samvera/hyrax/blob/004b6fa021848f11787f52909c196e7e97ce02d6/lib/generators/hyrax/work/templates/model.rb.erb#L11-L13) as the last line of any model.

ShanaLMoore:
Thanks for bringing it to our attention @dolsysmith!  It can be a challenge to make sure we don't break other systems along the way. 

After we confirm this PR is still good for Hyku main, we'll cut another release.

laritakr:
I hadn't seen this gotcha before! Great catch, Kirk! 

Is there somewhere we could document this in a readme or the wiki?

laritakr (approved):

Issue #1046, opened 2025-07-30
Bulkrax Deletions sometimes raises error when finding ActiveFedora::Base objects using Solr keys


maxkadel:
[Originating issue comment](https://github.com/notch8/palni_palci_knapsack/issues/371#issuecomment-3136824197)

The relevant part of the stacktrace is:
```
["/usr/local/bundle/bundler/gems/rails-d16199e50708/activemodel/lib/active_model/attribute_methods.rb:469:in `method_missing'",
  "/usr/local/bundle/bundler/gems/bulkrax-204acd1249de/app/factories/bulkrax/object_factory.rb:119:in `block in search_by_property'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation.rb:152:in `block in each'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation/finder_methods.rb:119:in `block (2 levels) in find_each'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation/finder_methods.rb:118:in `each'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation/finder_methods.rb:118:in `block in find_each'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation/finder_methods.rb:154:in `block in search_in_batches'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation/finder_methods.rb:150:in `loop'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation/finder_methods.rb:150:in `search_in_batches'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation/finder_methods.rb:117:in `find_each'",
  "/usr/local/bundle/gems/active-fedora-14.0.1/lib/active_fedora/relation.rb:152:in `each'",
  "/usr/local/bundle/bundler/gems/bulkrax-204acd1249de/app/factories/bulkrax/object_factory.rb:114:in `detect'",
  "/usr/local/bundle/bundler/gems/bulkrax-204acd1249de/app/factories/bulkrax/object_factory.rb:114:in `search_by_property'",
  "/usr/local/bundle/bundler/gems/bulkrax-204acd1249de/app/services/wings/custom_queries/find_by_source_identifier.rb:26:in `find_by_property_value'",
[...]

PR #1048, opened 2025-08-01
Use name_field for custom wings queries 

maxkadel:
When objects were in a state where they were no longer in the Solr index, but were still in Fedora, they were raising an error on delete when trying to find the Fedora record, because the ObjectFactory was using the :solr_field_tesim flavor as a method against the object, which is usually an undefined method.

Also fixes a typo

Connected to #1046


laritakr (approved):
Let's test this on actual fedora data before merging, but it looks like it should be the solution.

laritakr:
Testing successfully moved past this point, although actual deletion still requires Hyrax work.

PR #1049, opened 2025-08-07
Handle deletion & other issues with Valkyrie objects

laritakr:
## Summary
- fixes method `find_by_id`: This method was failing because it tried to use the find class method on an instance. This caused a silent deletion failure, where the importer reported success but the object was not deleted. It also caused the show page to be broken.
- adds another way to find the class from an entry The FactoryClassFinder did not find the class for deletions and other cases where the metadata was not parsed. This resulted in a fallback to the default work class.
- handles situations causing entry show page errors

### DIFF
--- a/app/factories/bulkrax/object_factory.rb
+++ b/app/factories/bulkrax/object_factory.rb
@@ -166,7 +166,7 @@ module Bulkrax
 
     def delete(_user)
       obj = find
-      return false unless obj
+      raise ObjectFactoryInterface::ObjectNotFoundError, "Object not found to delete" unless obj
 
       obj.delete(eradicate: true)
     end
--- a/app/factories/bulkrax/valkyrie_object_factory.rb
+++ b/app/factories/bulkrax/valkyrie_object_factory.rb
@@ -155,7 +155,7 @@ module Bulkrax
       Hyrax.query_service.find_by(id: id)
       # Because Hyrax is not a hard dependency, we need to transform the Hyrax exception into a
       # common exception so that callers can handle a generalize exception.
-    rescue Hyrax::ObjectNotFoundError => e
+    rescue Hyrax::ObjectNotFoundError, Valkyrie::Persistence::ObjectNotFoundError => e
       raise ObjectFactoryInterface::ObjectNotFoundError, e.message
     end
 
@@ -252,7 +252,7 @@ module Bulkrax
 
     def delete(user)
       obj = find
-      return false unless obj
+      raise ObjectFactoryInterface::ObjectNotFoundError, "Object not found to delete" unless obj
 
       Hyrax.persister.delete(resource: obj)
       Hyrax.index_adapter.delete(resource: obj)
@@ -381,7 +381,7 @@ module Bulkrax
     end
 
     def find_by_id
-      find(id: attributes[:id]) if attributes.key? :id
+      self.class.find(attributes[:id]) if attributes.key? :id
     end
 
     ##
--- a/app/parsers/bulkrax/csv_parser.rb
+++ b/app/parsers/bulkrax/csv_parser.rb
@@ -38,7 +38,7 @@ module Bulkrax
             #       We aren't right now because so many Bulkrax users are in between Fedora and Valkyrie
             if model.casecmp('collection').zero? || model.casecmp('collectionresource').zero?
               @collections << r
-            elsif model.casecmp('fileset').zero?
+            elsif model.casecmp('fileset').zero? || model.casecmp('hyrax::fileset').zero?
               @file_sets << r
             else
               @works << r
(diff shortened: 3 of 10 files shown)


app/parsers/bulkrax/csv_parser.rb
```diff
@@ -38,7 +38,7 @@ def build_records
             #       We aren't right now because so many Bulkrax users are in between Fedora and Valkyrie
             if model.casecmp('collection').zero? || model.casecmp('collectionresource').zero?
               @collections << r
-            elsif model.casecmp('fileset').zero?
+            elsif model.casecmp('fileset').zero? || model.casecmp('hyrax::fileset').zero?
```
laritakr, line 41: Ensures that Hyrax::FileSet records aren't grouped as works.

app/services/bulkrax/factory_class_finder.rb
```diff
@@ -66,27 +66,68 @@ def find
...
+
+    def model_from_parsed_metadata
+      Array.wrap(entry.parsed_metadata['model']).first if entry.parsed_metadata&.[]('model').present?
+    end
+
+    def work_type_from_parsed_metadata
+      Array.wrap(entry.parsed_metadata['work_type']).first if entry.importerexporter&.mapping&.[]('work_type').present?
+    end
+
+    def model_from_raw_metadata
```
laritakr, line 112: A new strategy to pull model out of raw metadata.

app/services/bulkrax/factory_class_finder.rb
```diff
@@ -66,27 +66,68 @@ def find
...
+      Array.wrap(entry.raw_metadata&.[]('model'))&.first if entry.raw_metadata&.[]('model').present?
+    end
+
+    def model_from_mapped_field
```
laritakr, line 116: another new strategy, similar to the above, but also checking each of the terms mapped to store the work's model.

app/factories/bulkrax/valkyrie_object_factory.rb
```diff
@@ -155,7 +155,7 @@ def self.find(id)
       Hyrax.query_service.find_by(id: id)
       # Because Hyrax is not a hard dependency, we need to transform the Hyrax exception into a
```
laritakr, line 156: this method previously didn't work with valkyrization. Now that it is working, we needed to capture the valkyrie errors as well.

app/factories/bulkrax/object_factory.rb
```diff
@@ -166,7 +166,7 @@ def find_by_id
 
     def delete(_user)
       obj = find
-      return false unless obj
+      raise ObjectFactoryInterface::ObjectNotFoundError, "Object not found to delete" unless obj
```
laritakr, line 169: It is misleading for a deletion to not happen but be reported as deleted. This ensures that the one running the importer knows about the failure.

app/factories/bulkrax/valkyrie_object_factory.rb
```diff
@@ -252,7 +252,7 @@ def self.ordered_file_sets_for(object)
 
     def delete(user)
       obj = find
-      return false unless obj
+      raise ObjectFactoryInterface::ObjectNotFoundError, "Object not found to delete" unless obj
```
laritakr, line 255: It is misleading for a deletion to not happen but be reported as deleted. This ensures that the one running the importer knows about the failure.

app/views/bulkrax/entries/show.html.erb
```diff
@@ -34,18 +34,22 @@
...
 	    <%# TODO Consider how to account for Bulkrax.collection_model_class %>
-        <% factory_record = @entry.factory.find %>
-        <% if factory_record.present? %>
-          <% factory_record_class = factory_record.class %>
-          <% factory_record_class_human = factory_record_class.model_name.human %>
-          <strong><%= factory_record_class_human %> Link:</strong>
-          <% if defined?(Hyrax) && factory_record_class_human == 'Collection' %>
-            <%= link_to factory_record_class_human, hyrax.polymorphic_path(factory_record) %>
+        <% begin %>
+          <% factory_record = @entry.factory.find rescue nil %>
```
laritakr, line 38: Now that find works, we need a way to handle cases where the record isn't found (when we successfully delete a record, it is no longer there and we don't want the find error to bubble up here)

app/services/bulkrax/factory_class_finder.rb
```diff
@@ -66,27 +66,68 @@ def find
...
+      # Normalize the string format
+      normalize_class_name(fc)
     rescue
       entry.default_work_type
     end
+
+    ##
+    # Try each strategy in sequence to find a factory class name
+    # @return [String, nil] the factory class name or nil if none found
+    def find_factory_class_name
```
maxkadel, line 87: My colleague @sandbergja came up with a pattern I liked for something similar - basically make each strategy its own class, then have the classes have a consistent api of `.eligible?`, then you could select all eligible factories. 

See [orangelight requests router](https://github.com/pulibrary/orangelight/blob/main/app/models/requests/router.rb) - just something to think about for the future!

PR #1052, opened 2025-08-08
Raise an error for a method that doesn't do anything yet

maxkadel:
- I recognize that this might be a breaking change, but if we hit it on real data, that could cause surprising problems


laritakr (approved):
I think this is an important change to make! Good catch, and hopefully we can get it implemented soon. :)

PR #1059, opened 2025-08-25
Raise specific errors

maxkadel:
- Raise specific errors (when they are knowable), so that we can use more targeted error handling in future.
- Add more error handling testing
- Also remove some code that has been dead for 3 years.

Existing code that rescues StandardError will still work, since all of these custom errors inherit from StandardError.

### DIFF
--- a/app/models/bulkrax/csv_entry.rb
+++ b/app/models/bulkrax/csv_entry.rb
@@ -368,7 +368,40 @@ module Bulkrax
       Bulkrax::CsvMatcher
     end
 
+    def collection_identifiers
+      return @collection_identifiers if @collection_identifiers.present?
+
+      parent_field_mapping = self.class.parent_field(parser)
+      return [] unless parent_field_mapping.present? && record[parent_field_mapping].present?
+
+      identifiers = []
+      split_references = record[parent_field_mapping].split(Bulkrax.multi_value_element_split_on)
+      split_references.each do |c_reference|
+        matching_collection_entries = importerexporter.entries.select do |e|
+          (e.raw_metadata&.[](source_identifier) == c_reference) &&
+            e.is_a?(CsvCollectionEntry)
+        end
+        raise ::StandardError, 'Only expected to find one matching entry' if matching_collection_entries.count > 1
+        identifiers << matching_collection_entries.first&.identifier
+      end
+      @collection_identifiers = identifiers.compact.presence || []
+    end
+
+    def collections_created?
+      # TODO: look into if this method is still needed after new relationships code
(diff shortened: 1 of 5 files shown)


K8Sewell (approved):

ShanaLMoore:
@maxkadel Has this code been tested in an application like Hyku?

maxkadel:
> @maxkadel Has this code been tested in an application like Hyku? In particular has it been tested with collections?

Not yet - I haven't been able to get my local Hyku to come up with my local Bulkrax yet.

ShanaLMoore:
cc @maxkadel gotcha. Let me know when you want to pair; team v can help with that. We definitely need to test bulkrax from the context of an app and not strictly rely on specs. 😅 🙏🏿

maxkadel:
@ShanaLMoore - if you'd feel more comfortable if I split the removing the dead collection code into its own PR I'm also happy to do that.

ShanaLMoore:
Sure, maybe I’m being overly cautious, but as a general rule of thumb we should verify Bulkrax changes in the context of a Hyku app, not just rely on specs (until the state of them improve).

On the dead code, I’d feel better if we could confirm that none of our older clients are overriding or depending on it before it gets removed. I’m especially thinking about PALS, Adventist, and the British Library.

As a general practice, I’d also lean toward marking code like this as deprecated first (with a warning), and then removing it in a later release. That way anyone with overrides has a chance to adapt. Splitting it into a separate PR sounds good too; that would make it easier to review and test in isolation.

app/models/bulkrax/csv_entry.rb
```diff
@@ -351,40 +368,7 @@ def self.matcher_class
       Bulkrax::CsvMatcher
     end
 
-    def collection_identifiers
```
orangewolf, line 354: @ShanaLMoore is this the method you were talking about? I don't think this is unused. I think it is specifically used in the find method. It gets called the first time and then cached

laritakr: I definitely see it being used in csv_entry.rb. 

```
    def find_collection_ids
      return self.collection_ids if collections_created?
      if collection_identifiers.present?
        collection_identifiers.each do |collection_id|
          c = find_collection(collection_id)
          skip = c.blank? || self.collection_ids.include?(c.id)
          self.collection_ids << c.id unless skip
        end
      end

      self.collection_ids
    end
```

ShanaLMoore (dismissed):
dismiss comment/review

laritakr (approved):
Looks good.

PR #1064, opened 2025-09-15
Delete the file sets when we delete a work

laritakr:
## Summary

We are not using transactions when deleting the
work, so the filesets were getting left behind.

### DIFF
--- a/app/factories/bulkrax/valkyrie_object_factory.rb
+++ b/app/factories/bulkrax/valkyrie_object_factory.rb
@@ -297,6 +297,9 @@ module Bulkrax
       Hyrax.persister.delete(resource: obj)
       Hyrax.index_adapter.delete(resource: obj)
       Hyrax.publisher.publish('object.deleted', object: obj, user: user)
+      # delete the file sets when we delete a work
+      return unless obj.respond_to?(:file_sets) && obj.file_sets.present?
+      destroy_existing_files if obj.respond_to?(:file_sets) && obj.file_sets.present?
     end
 
     def run!


app/factories/bulkrax/valkyrie_object_factory.rb
```diff
@@ -558,10 +564,10 @@ def destroy_existing_files
           .value!
       end
 
-      @object.member_ids = @object.member_ids.reject { |m| existing_files.detect { |f| f.id == m } }
-      @object.rendering_ids = []
-      @object.representative_id = nil
-      @object.thumbnail_id = nil
+      object.member_ids = object.member_ids.reject { |m| existing_files.detect { |f| f.id == m } }
```
maxkadel, line 567: Just curious - why did we remove the memoization here?

app/models/concerns/bulkrax/export_behavior.rb
```diff
@@ -27,10 +27,10 @@ def hyrax_record
     # Prepend the file_set id to ensure a unique filename and also one that is not longer than 255 characters
     def filename(file_set)
       # return if there are no files on the fileset
-      return if Bulkrax.object_factory.original_file(fileset: file_set).blank?
+      file = Bulkrax.object_factory.original_file(fileset: file_set)
+      return if file.blank?
```
maxkadel, line 31: Again, if we're expecting a string, let's return an empty string if we don't have the info we need, to keep a more consistent API.

```suggestion
      return '' if file.blank?
```

I know that this is a shift from the Hyrax pattern, but I think it might help us slowly dig ourselves out of the error reporting / saving / rescuing problems we've been having.

app/factories/bulkrax/valkyrie_object_factory.rb
```diff
@@ -236,7 +236,8 @@ def self.thumbnail_for(resource:)
     # @input [Fileset or FileMetadata]
     # @return [FileMetadata] the original file
     def self.original_file(fileset:)
-      fileset.try(:original_file) || fileset
+      return fileset if fileset.is_a?(Hyrax::FileMetadata)
```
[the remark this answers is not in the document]
laritakr: Yeah I don't like it, but there is a lot of intermixing between Hyrax::Filesets, Hyrax::FileMetadata, and the fedora Filesets so I like to be safe since it's difficult to test every scenario that might arise.

I wasn't sure which I might be getting in any particular situation. We often try whether an object responds to a particular method, but there was an issue if a fileset has no files in this situation. This was the best solution I could find.

PR #1072, opened 2025-10-06
Update a FileSet's metadata via Bulkrax

maxkadel:


maxkadel:
> Do you have an example for what the csv would look like for a metadata update?

Sorry - I've added example .zip files - the first one for the initial import, the second one for the update. The title should change between the first and the second one.

PR #1086, opened 2025-12-10
:bug: Improve error reporting for missing files in CSV imports

ShanaLMoore:
Issue:
- https://github.com/notch8/palni_palci_knapsack/issues/367
- https://github.com/notch8/palni_palci_knapsack/issues/527

Fail the importer when any referenced file is missing, and collect all missing filenames in a single validation error instead of stopping at the first one.

This prevents silent “successful” imports and helps users fix all issues in one attempt.

# BEFORE

Failed silently. Marked import as complete but no entries were processed: 

image

image

# AFTER

image

image


ShanaLMoore:
closing because I later realized the client had an override that impacted the behavior. This change is not needed, as confirmed by testing in Hyku main.

PR #1099, opened 2026-02-12
Support CSV import zips created from parent folder

ShanaLMoore:
# Summary

When users zip the export parent directory instead of its contents, Bulkrax could not find the files in the "files/" directory. 

zipped from parent directory: [parent-basic.zip](https://github.com/user-attachments/files/25275485/parent-basic.zip)
zipped per "bulkrax standards": [basic.zip](https://github.com/user-attachments/files/25275491/basic.zip)

# BEFORE

It can't find the files because of the nested structure.

image

image

# AFTER

image


laritakr:
Addressed by https://github.com/samvera/bulkrax/pull/1146

PR #1108, opened 2026-02-16
I1078 turbolinks fallbacks

hippietilley:
This PR adds fallbacks for Turbo and jQuery for Hyrax based apps that don't use Turbolinks.


kirkkwang (approved):
Thanks for contributing!

PR #1126, opened 2026-02-24
V2 chunk uploads

ShanaLMoore:
# Summary

V2 Bulkrax importer now uses the same chunked-upload flow as v1: files are uploaded to Hyrax’s `/uploads/` endpoint in 10 MB chunks as soon as they’re selected. A progress bar shows upload status, and “Validate Files” is disabled until all uploads finish. Both “Validate Files” and “Start Import” send only uploaded file IDs (`uploaded_files[]`); the server resolves them to existing `Hyrax::UploadedFile` records, so the file is uploaded once and reused. The tenant account’s file size limit (e.g. `Hyrax.config.uploader[:maxFileSize]`) is enforced client-side before starting upload, matching v1 behavior.

## Ticket Number

- https://github.com/samvera/bulkrax/issues/1123#issuecomment-3947239959

## Screenshots / Video

- Before: Selecting a large file showed an immediate green checkmark; “Validate Files” then timed out.

<details> 

image

</details>

- After: Selecting a file shows “Uploading… X%” and a full-width progress bar; when upload completes, the checkmark appears and “Validate Files” is enabled. 

<details> 

image

</details> 

## Expected Behavior

1. User selects CSV/ZIP → chunked upload to `/uploads/` starts immediately with a visible progress bar.
2. “Validate Files” stays disabled until all selected files have finished uploading.
3. “Validate Files” sends only `uploaded_files[]` IDs; no file bytes. Validation uses the same files already on the server.
4. “Start Import” submits the form with the same `uploaded_files[]` IDs; create uses those same uploads (no second upload).
5. Files over the tenant’s max file size are rejected before upload with a clear error (e.g. “File size exceeds the maximum allowed (X MB per file)”).

## Sample Files

- Use a large ZIP (e.g. ~300 MB) to confirm progress bar and no timeout on Validate.
- Use a file larger than the tenant limit to confirm rejection message.

## Notes

- Chunk size: 10 MB (aligned with Hyrax v1 uploader).
- Backend supports both `params[:uploaded_files]` (IDs) and legacy `params[:importer][:parser_fields][:files]` (multipart) for validation and create; v2 UI sends IDs only.
- File path mode and demo scenarios unchanged.
- Validation errors CSV download and file-remove also use uploaded file IDs where applicable.

### DIFF
--- a/app/controllers/concerns/bulkrax/importer_v2.rb
+++ b/app/controllers/concerns/bulkrax/importer_v2.rb
@@ -37,14 +37,7 @@ module Bulkrax
     def create_v2
       files = nil
       files = resolve_create_files
-
-      if params[:uploaded_files].present? && files.empty?
-        respond_to do |format|
-          format.html { render :new_v2, status: :unprocessable_entity }
-          format.json { render json: { errors: ['No valid uploaded files found'] }, status: :unprocessable_entity }
-        end
-        return
-      end
+      return render_invalid_uploaded_files_response if params[:uploaded_files].present? && files.empty?
 
       @importer = Importer.new(importer_params_v2)
       @importer.parser_klass = 'Bulkrax::CsvParser'
@@ -124,6 +117,13 @@ module Bulkrax
         base.where(user_id: current_user.id)
       else
         base.none
+      end
+    end
+
+    def render_invalid_uploaded_files_response
+      respond_to do |format|
+        format.html { render :new_v2, status: :unprocessable_entity }
+        format.json { render json: { errors: ['No valid uploaded files found'] }, status: :unprocessable_entity }
       end
     end
 


app/controllers/concerns/bulkrax/importer_v2.rb
```diff
@@ -30,10 +30,14 @@ def validate_v2
 
       admin_set_id = params[:importer]&.[](:admin_set_id)
       render json: StepperResponseFormatter.format(run_validation(csv_file, zip_file, admin_set_id: admin_set_id)), status: :ok
```
laritakr: which is better... passing a potentially large zip file around in memory, or passing an id and having it reload to validate?

ShanaLMoore: Right now we still resolve IDs to file objects in the controller and pass csv_file/zip_file into run_validation. Changing to pass IDs would mean having the validation layer accept IDs and resolve them there; we can do that as a follow-up refactor if we want that boundary to move.

PR #1129, opened 2026-02-25
I18n

sjproctor:
# Internationalization Support

Ref:
- https://github.com/notch8/hyku-community-issues/issues/76
- https://github.com/notch8/hyku-community-issues/issues/34

### Updates
- Adds i18n-tasks gem
- Moves all text from v2 into locale file
- Add translation files for 6 new locales: German (de), Spanish (es), French (fr), Italian (it), Portuguese-Brazilian (pt-BR), and Chinese (zh)
- Update i18n-tasks.yml config to use the bulkrax.<locale>.yml naming convention and set the write path so new keys are written to the correct files


kirkkwang (approved):
Nice!

PR #1165, opened 2026-04-03
Source identifier validations

laritakr:
# Validates source identifiers

- validate & report row validation warnings and errors separately
- correct method of finding in repository... finding in the entries table is not adequate to ensure the record exists, and finding in the repo by source identifier wasn't working correctly.

# Screenshot
Screenshot 2026-04-03 at 2 11 28 PM

Screenshot 2026-04-03 at 2 18 12 PM

# Exported as file

[row-errors-and-warnings_errors (1).csv](https://github.com/user-attachments/files/26468841/row-errors-and-warnings_errors.1.csv)

### DIFF
--- a/app/parsers/concerns/bulkrax/csv_parser/csv_validation_helpers.rb
+++ b/app/parsers/concerns/bulkrax/csv_parser/csv_validation_helpers.rb
@@ -135,26 +135,30 @@ module Bulkrax
       end
 
       # Builds the find_record lambda used by row validators and hierarchy extraction.
-      def build_find_record(mapping_manager, mappings)
-        work_identifier        = mapping_manager.resolve_column_name(flag: 'source_identifier', default: 'source').first&.to_s || 'source'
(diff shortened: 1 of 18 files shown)


app/parsers/concerns/bulkrax/csv_parser/csv_validation_helpers.rb
```diff
@@ -135,26 +137,26 @@ def assemble_result(headers:, missing_required:, header_issues:, row_errors:, cs
       end
 
       # Builds the find_record lambda used by row validators and hierarchy extraction.
-      def build_find_record(mapping_manager, mappings)
-        work_identifier        = mapping_manager.resolve_column_name(flag: 'source_identifier', default: 'source').first&.to_s || 'source'
-        work_identifier_search = Array.wrap(mappings.dig(work_identifier, 'search_field')).first&.to_s ||
+      def build_find_record
```
laritakr, line 140: In order to find the correct identifier from the mappings, we have to load the mappings here. The mapping manager excluded the source_identifier and the find was falling back to the wrong search term.

kirkkwang (approved):

PR #1186, opened 2026-04-27
Add Bulkrax::Ability concern and enforce importer/exporter ownership

orangewolf:
Introduces a shared Bulkrax::Ability concern so all Bulkrax authorization methods live in one place, and adds per-record ownership enforcement so users can only access importers/exporters they own (or admin all records via new can_admin_importers?/can_admin_exporters? methods).

- Add app/models/concerns/bulkrax/ability.rb with default implementations of can_import_works?, can_export_works?, can_admin_importers?, and can_admin_exporters? (all safe/conservative defaults)
- Add accessible_importers, accessible_exporters, and item_accessible? helpers to ApplicationController; admin users see all records, others see only their own
- Scope importer_table, set_importer, continue, upload_corrected_entries, upload_corrected_entries_file, and export_errors to accessible_importers; API requests bypass the ownership scope to preserve existing API behavior
- Scope exporter_table, set_exporter, and download to accessible_exporters
- Check item_accessible? in entries_controller update/destroy and when loading parent importer/exporter in show actions
- Guard edit/delete action links in datatables_behavior so they only appear for owners and admins; scope recordsTotal/recordsFiltered counts to accessible records
- Update install_generator to include Bulkrax::Ability instead of injecting raw method definitions; add commented override examples
- Update test app Ability to include Bulkrax::Ability
- Add spec/models/concerns/bulkrax/ability_spec.rb for new concern
- Add ownership enforcement specs to importers and exporters controller specs (non-owner denied, owner allowed, admin allowed)
- Fix export_errors and datatables specs to account for ownership scoping

### DIFF
--- a/app/controllers/bulkrax/importers_controller.rb
+++ b/app/controllers/bulkrax/importers_controller.rb
@@ -30,7 +30,7 @@ module Bulkrax
 
     def importer_table
       order = table_order.presence || Arel.sql('last_imported_at DESC NULLS LAST')
-      @importers = Importer.order(order).page(table_page).per(table_per_page)
+      @importers = accessible_importers.order(order).page(table_page).per(table_per_page)
       @importers = @importers.where(importer_table_search) if importer_table_search.present?
       respond_to do |format|
         format.json { render json: format_importers(@importers) }
@@ -177,7 +177,7 @@ module Bulkrax
 
     # PUT /importers/1
     def continue
-      @importer = Importer.find(params[:importer_id])
+      @importer = accessible_importers.find(params[:importer_id])
       params[:importer] = { name: @importer.name }
       @importer.validate_only = false
       update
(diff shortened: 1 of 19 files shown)


app/controllers/bulkrax/importers_controller.rb
```diff
@@ -13,14 +13,33 @@ class ImportersController < ::Bulkrax::ApplicationController
...
+    before_action :load_and_authorize_importer_by_importer_id,
+                  only: [:continue, :upload_corrected_entries, :upload_corrected_entries_file, :export_errors, :original_file]
     with_themed_layout 'dashboard' if defined?(::Hyrax)
 
     # GET /importers
     def index
       # NOTE: We're paginating this in the browser.
       if api_request?
+        # TODO: Scope API index by token owner once token-to-user wiring is in
+        # place.  Tracked in [ISSUE].  Until then, API clients see all importers
```
laritakr, line 41: Do we have an actual issue to link here for tracking?

laritakr (approved):
Looks great. I saw a few places referring to the API token work... would be ideal to update to reference an actual ticket.

